PR #89, opened 2016-01-11
PCHR-434: Added ID's to panels, so they can be linked to

Agares4:
Added ID's to "tasks" and "documents", so they can be linked to from outside.

This PR is required by:  https://github.com/compucorp/civihr-tasks-assignments/pull/9


nishant-bhorodia:
Approved

PR #91, opened 2016-01-14
Pchr 262 - absence reports, refactor, settings + fixes

gregmeszaros:
- More abstracted whitelisting for the template file - based on the report types it will load the templates/civihr-employee-portal-custom-reports-filter-result.tpl file. Previously the report types were hardcoded
- Added View FTE / Level - from the demo presentation
- Added new relationship handler -> absence view can now have a relationship with contact entity. Previously the absence view we were unable to link it to contact entity in the views
- Added raw result type for duration handler - this raw result type is used to calculate the duration in the absence days taken report
- Added days taken view and report
- Fixed date range slider to show month names and not days
- Added some more default X axis filter types so they can be set in the settings page and enabled

### DIFF
--- a/civihr_employee_portal/views/includes/civihr_employee_portal_handler_absence_duration.inc
+++ b/civihr_employee_portal/views/includes/civihr_employee_portal_handler_absence_duration.inc
@@ -35,7 +35,12 @@ class civihr_employee_portal_handler_absence_duration extends views_handler_fiel
         
         else {
             $value = $this->get_value($values);
-            
+
+            // If the raw format type is selected return the count but not add any specific formatting
+            if (isset($this->options['duration_type']) && $this->options['duration_type'] == 'no_format') {
+                return $value / (6 * 80);
+            }
+
             return $value <= 480 ? $value / (6 * 80) . ' day' : $value / (6 * 80) . ' days';
         }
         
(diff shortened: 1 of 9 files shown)


civihr_employee_portal/views/includes/civihr_employee_portal_handler_absence_duration.inc
```diff
@@ -35,7 +35,12 @@ class civihr_employee_portal_handler_absence_duration extends views_handler_fiel
         
         else {
             $value = $this->get_value($values);
-            
+
+            // If the raw format type is selected return the count but not add any specific formatting
+            if (isset($this->options['duration_type']) && $this->options['duration_type'] == 'no_format') {
+                return $value / (6 * 80);
```
AkA84: Just as a curiosity, where these specific values come from?

AkA84:
Approved

gregmeszaros:
Hi,

So historically the duration records in civihr were stored with some strange logic.

1 day = 480
0.5 = 240

so if you for example are on leave for 3 days what will be stored is 1440.
When you want to know the exact duration of your leave you need to divide the duration value by 480.

screen shot 2016-01-14 at 14 04 29

PR #93, opened 2016-01-14
PCHR-792 - Restricting the 'Appraisal' SSP top menu item to 'admin' and 'manager' roles only.

coldrunKacper:
Restricting the 'Appraisal' SSP top menu item only to Admin and Manager roles + changing the menu item title from 'Appraisal Manager' to 'Appraisal'.


AkA84:
Approved

PR #94, opened 2016-01-15
PCHR-788: Added summary table in reports

Agares4:
Added "summary" table under reports:
zrzut ekranu z 2016-01-15 o 12 29 44

### DIFF
--- a/civihr_employee_portal/js/reports.js
+++ b/civihr_employee_portal/js/reports.js
@@ -957,6 +957,9 @@
             //
             // The code inside if takes care of the first possibility
             // The else takes care of the second
+            //
+            // The "gender" and "department" keys are hardcoded in Drupal Views
+            // Probably they should be renamed to something like "xaxis"/"yaxis"
             if (summary[0] !== undefined && summary[0].data.count !== undefined) {
                 for (i in summary) {
                     normalisedSummary[summary[i].data.department] = +summary[i].data.count;
--- a/civihr_employee_portal/js/reports.js
+++ b/civihr_employee_portal/js/reports.js
@@ -919,73 +919,80 @@
                         }
 
                         this.$DOM.sections.dataWrapper.fadeIn();
-                        var $DOM = this.$DOM;
-
-                        $.ajax({
-                            'type': 'GET',
-                            'url': this.getJsonUrl() + ( typeof path !== 'undefined' ? path : '' ),
-                            'success': function(data) {
-                                var mappedResults = {},
-                                    key,
-                                    countRow = $('<tr />'),
-                                    headerRow = $('<tr />'),
-                                    percentageRow = $('<tr />'),
-                                    total = 0;
-
-                                if(data.results[0] !== undefined && data.results[0].data.count !== undefined) {
-                                    for(var i in data.results) {
-                                        mappedResults[data.results[i].data.department] = +data.results[i].data.count;
-                                    }
-                                } else {
-                                    mappedResults = data.results.map(function (row) {
-                                        return {y: row.data.gender, x: 1}
-                                    }).reduce(function (accumulator, row) {
-                                        if (accumulator[row.y] === undefined) {
-                                            accumulator[row.y] = 0;
-                                        }
-
-                                        accumulator[row.y] += row.x;
-
-                                        return accumulator;
-                                    }, {});
-                                }
-
-                                for(key in mappedResults) {
-                                    total += mappedResults[key];
-                                }
-
-                                for(key in mappedResults) {
-                                    if(!mappedResults.hasOwnProperty(key)) {
-                                        continue;
-                                    }
-
-                                    var percentage = Math.round((mappedResults[key]/total)*10000)/100;
-
-                                    headerRow.append($('<th />').text(key));
-                                    countRow.append($('<td />').text(mappedResults[key]));
-                                    percentageRow.append($('<td />').text(percentage + '%'));
-                                }
-
-                                var table = $('<table />').addClass('table table-striped table-hover')
-                                    .append($('<thead />').append(headerRow))
-                                    .append($('<tbody />').append(countRow).append(percentageRow));
-
-                                var header = $('<header />').addClass('panel-heading')
-                                    .append($('<h2>Summary</h2>').addClass('panel-title'));
-
-                                var section = $('<section />').addClass('panel panel-primary')
-                                    .append(header)
-                                    .append(table);
-
-                                $DOM.sections.summaryWrapper.show().html(section);
-                            }
-                        });
                     }.bind(this));
             }.bind(this),
             error: function (data) {
                 this.$DOM.sections.dataWrapper.html('An error occured!');
             }.bind(this)
         });
+
+        $.ajax({
+            type: 'GET',
+            url: this.getJsonUrl() + ( typeof path !== 'undefined' ? path : '' ),
+            success: function(data) {
+                var mappedResults = normaliseSummaryData(data.results);
+
+                replaceSummaryTable(mappedResults, this.$DOM);
+            }.bind(this)
+        });
+
+        function normaliseSummaryData(summary) {
+            var normalisedSummary = {},
+                i;
+
+            // The data can have two formats, depending on particular report:
+            // data = { results: [{ departament: "xxx", count: 1 }, ...]) }
+            // or
+            // data = { results: [{ gender: "aaa" }, { gender: "aaa" }, ...] }
+            //
+            // The code inside if takes care of the first possibility
+            // The else takes care of the second
+            if (summary[0] !== undefined && summary[0].data.count !== undefined) {
+                for(i in summary) {
+                    normalisedSummary[summary[i].data.department] = +summary[i].data.count;
+                }
+            } else {
+                normalisedSummary = summary.map(function (row) {
+                    return { y: row.data.gender, x: 1 }
+                }).reduce(function (accumulator, row) {
+                    if (accumulator[row.y] === undefined) {
+                        accumulator[row.y] = 0;
+                    }
+
+                    accumulator[row.y] += row.x;
+
+                    return accumulator;
+                }, {});
+            }
+
+            return normalisedSummary;
+        }
+
+        function replaceSummaryTable(summary, $DOM) {
+            var $countRow = $('<tr />'),
+                $headerRow = $('<tr />'),
+                $percentageRow = $('<tr />'),
+                summaryKeys = Object.keys(summary),
+                total = summaryKeys.reduce(function(accumulator, key) { return accumulator + summary[key]; }, 0);
+
+            summaryKeys
+                .forEach(function(key) {
+                    var percentage;
+
+                    percentage = Math.round((summary[key]/total)*10000)/100;
+
+                    $headerRow.append($('<th />').text(key));
+                    $countRow.append($('<td />').text(summary[key]));
+                    $percentageRow.append($('<td />').text(percentage + '%'));
+                });
+
+            $DOM.sections.summaryWrapper
+                .show()
+                .find('table')
+                .empty()
+                .append($('<thead />').append($headerRow))
+                .append($('<tbody />').append($countRow).append($percentageRow));
+        }
 
         // Returns the URL for the ajax call
         function buildURL() {
@@ -1344,6 +1351,7 @@
         this.hideButtonTpls();
         this.hideSubFilters();
         this.setDataWrapperVisibility();
+        this.$DOM.sections.summaryWrapper.hide();
     };
 
     /**
--- a/civihr_employee_portal/js/reports.js
+++ b/civihr_employee_portal/js/reports.js
@@ -919,6 +919,67 @@
                         }
 
                         this.$DOM.sections.dataWrapper.fadeIn();
+                        var $DOM = this.$DOM;
+
+                        $.ajax({
+                            'type': 'GET',
+                            'url': this.getJsonUrl() + ( typeof path !== 'undefined' ? path : '' ),
+                            'success': function(data) {
+                                var mappedResults = {},
+                                    key,
+                                    countRow = $('<tr />'),
+                                    headerRow = $('<tr />'),
+                                    percentageRow = $('<tr />'),
+                                    total = 0;
+
+                                if(data.results[0] !== undefined && data.results[0].data.count !== undefined) {
+                                    for(var i in data.results) {
+                                        mappedResults[data.results[i].data.department] = +data.results[i].data.count;
+                                    }
+                                } else {
+                                    mappedResults = data.results.map(function (row) {
+                                        return {y: row.data.gender, x: 1}
+                                    }).reduce(function (accumulator, row) {
+                                        if (accumulator[row.y] === undefined) {
+                                            accumulator[row.y] = 0;
+                                        }
+
+                                        accumulator[row.y] += row.x;
+
+                                        return accumulator;
+                                    }, {});
+                                }
+
+                                for(key in mappedResults) {
+                                    total += mappedResults[key];
+                                }
+
+                                for(key in mappedResults) {
+                                    if(!mappedResults.hasOwnProperty(key)) {
+                                        continue;
+                                    }
+
+                                    var percentage = Math.round((mappedResults[key]/total)*10000)/100;
+
+                                    headerRow.append($('<th />').text(key));
+                                    countRow.append($('<td />').text(mappedResults[key]));
+                                    percentageRow.append($('<td />').text(percentage + '%'));
+                                }
+
+                                var table = $('<table />').addClass('table table-striped table-hover')
+                                    .append($('<thead />').append(headerRow))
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/js/reports.js
```diff
@@ -919,6 +919,67 @@
                         }
 
                         this.$DOM.sections.dataWrapper.fadeIn();
+                        var $DOM = this.$DOM;
+
+                        $.ajax({
```
AkA84: Is it necessary to have this `$.ajax` call in the `ready` handler of the data table DOM object? That is, are you sure you need to create the summary table only when the data table is built and displayed? Can't they be done in parallel, since  it seems you are not overwriting the `data` object returned by the initial ajax call?

civihr_employee_portal/templates/civihr-employee-portal-custom-reports.tpl.php
```diff
@@ -105,6 +105,9 @@ class="btn btn-lg btn-secondary-outline text-uppercase">
             </div>
         </footer>
     </section>
+    <div data-graph-section="summary">
```
kkalamarski: I don't know exactly how everything is build in reports, but is it necessary to insert all the HTML via Javascript?

civihr_employee_portal/js/reports.js
AkA84: I believe the jQuery version we are using is already using promises for ajax calls, so you can write something like this:

```js
$.ajax({
    type: 'GET',
    url: buildURL.call(this),
})
.then(function (data) {
    this.$DOM.sections.dataWrapper.html(data).ready(/* ... */);
})
.then(function () {
    return $ajax({
        'type': 'GET',
        'url': this.getJsonUrl() + ( typeof path !== 'undefined' ? path : '' )
    });
})
.then(function (data) {
    var mappedResults = {}

    /* .... */
});

File: civihr_employee_portal/js/reports.js [hunk 3]
```diff
@@ -919,6 +919,67 @@
...
+                                    }
+
+                                    var percentage = Math.round((mappedResults[key]/total)*10000)/100;
+
+                                    headerRow.append($('<th />').text(key));
+                                    countRow.append($('<td />').text(mappedResults[key]));
+                                    percentageRow.append($('<td />').text(percentage + '%'));
+                                }
+
+                                var table = $('<table />').addClass('table table-striped table-hover')
```
kkalamarski: If it's necessary to build View here, it could be cleaner if it was in separate method

civihr_employee_portal/js/reports.js
```diff
@@ -919,6 +919,67 @@
                         }
 
                         this.$DOM.sections.dataWrapper.fadeIn();
+                        var $DOM = this.$DOM;
```
AkA84: This is not necessary, just `bind` the `success` fn to this

```js
$ajax({
    type: 'GET',
    url: this.getJsonUrl() + ( typeof path !== 'undefined' ? path : '' )
    success: function (data) {

    }.bind(this);
});
```

civihr_employee_portal/templates/civihr-employee-portal-custom-reports.tpl.php
Agares4: Yeah, actually I think I could put here the header, and an empty table and populate it from JS

civihr_employee_portal/js/reports.js
Agares4: Yeah, you're right, I'll try to do those things in parallel

civihr_employee_portal/js/reports.js
```diff
@@ -919,6 +919,67 @@
...
+
+                        $.ajax({
+                            'type': 'GET',
```
AkA84: Don't use strings for object literals key names unless it's necessary, rewrite as
```js
$.ajax({
    type: 'GET',
    url: this.getJsonUrl() + ( typeof path !== 'undefined' ? path : '' ),
    success: function () {
    }
});
```

civihr_employee_portal/js/reports.js
AkA84: Totally agree, this function is way too long, refactor so you can write something like
```js
$DOM.sections.summaryWrapper.show().html(buildSummaryTable(arguments...)));
```

civihr_employee_portal/js/reports.js
```diff
@@ -919,6 +919,67 @@
...
+                            'url': this.getJsonUrl() + ( typeof path !== 'undefined' ? path : '' ),
+                            'success': function(data) {
+                                var mappedResults = {},
+                                    key,
+                                    countRow = $('<tr />'),
+                                    headerRow = $('<tr />'),
+                                    percentageRow = $('<tr />'),
+                                    total = 0;
+
+                                if(data.results[0] !== undefined && data.results[0].data.count !== undefined) {
```
AkA84:
```js
if (expression) {} // cool
if(expression) {} // not cool
```

civihr_employee_portal/js/reports.js
```diff
@@ -919,6 +919,67 @@
...
+                                    for(var i in data.results) {
+                                        mappedResults[data.results[i].data.department] = +data.results[i].data.count;
+                                    }
+                                } else {
+                                    mappedResults = data.results.map(function (row) {
+                                        return {y: row.data.gender, x: 1}
```
AkA84:
```js
{ foo: foo, bar: bar } //cool
{foo: foo, bar: bar} // not cool
```

civihr_employee_portal/templates/civihr-employee-portal-custom-reports.tpl.php
AkA84: You have to choices here, I think:

1. You stick with creating the HTML via javascript, but then please see if you can reuse some of the code already written for it, given that we're already creating a table in `reports.js`
2. You use an hidden template that you clone and populate in JS (preferred). You can see how it's done in the `cloneButtonTpl` function

civihr_employee_portal/js/reports.js
AkA84: Please add a comment explaining what's going on in this `if/else` block

civihr_employee_portal/js/reports.js
AkA84: You definitely need two functions here, one for parsing/normalizing the data, one for building the table from that data

civihr_employee_portal/js/reports.js
```diff
@@ -919,6 +919,67 @@
...
+                                        }
+
+                                        accumulator[row.y] += row.x;
+
+                                        return accumulator;
+                                    }, {});
+                                }
+
+                                for(key in mappedResults) {
+                                    total += mappedResults[key];
```
AkA84: If I understand this correctly, you can scrap the `for` loop and just write it as
```js
total = Object.keys(mappedResults).length
```

civihr_employee_portal/js/reports.js
```diff
@@ -919,6 +919,67 @@
...
+                                }
+
+                                for(key in mappedResults) {
```
AkA84:
```js
Object.keys(mappedResult).forEach(function (key) {
    // ...
});
```
From [Object.keys()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) page:
> The Object.keys() method returns an array of a given object's own enumerable properties, in the same order as that provided by a for...in loop (the difference being that a for-in loop enumerates properties in the prototype chain as well).

civihr_employee_portal/js/reports.js
```diff
@@ -919,6 +919,67 @@
...
                         this.$DOM.sections.dataWrapper.fadeIn();
+                        var $DOM = this.$DOM;
+
+                        $.ajax({
+                            'type': 'GET',
+                            'url': this.getJsonUrl() + ( typeof path !== 'undefined' ? path : '' ),
+                            'success': function(data) {
+                                var mappedResults = {},
+                                    key,
+                                    countRow = $('<tr />'),
```
AkA84: For any jQuery DOM object, please use a variable name that starts with a `$`, so they're easier to recognize
```js
    $countRow = $('<tr />'),
    $headerRow = $('<tr />'),
    $percentageRow = $('<tr />'),
```

civihr_employee_portal/js/reports.js
Agares4: mappedResults[key] is not always 1:

     mappedResults[data.results[i].data.department] = +data.results[i].data.count;

civihr_employee_portal/js/reports.js
AkA84: Apologies, what I wrote above was still for a sequential approach, obviously the parallel approach would be something like:
```js
// data table
$.ajax({
}).then(function (data) {
}

// summary table
$.ajax({
}).then(function (data) {
}

```

civihr_employee_portal/js/reports.js
kkalamarski: Not sure if I understand this, but then you can use Array.prototype.reduce() to sum it up.

civihr_employee_portal/templates/civihr-employee-portal-custom-reports.tpl.php
Agares4: Hmm... is it okay to have a table and headers in HTML and then just .append() tbody and thead to it?

civihr_employee_portal/js/reports.js
AkA84: Right, i see. But then why not calculating the `total` directly into the `reduce` function right above? You would avoid having to loop through `mappedResult`. Also if below you care about counting only the own properties, it seems you are not doing that for this particular `for in` loop

civihr_employee_portal/js/reports.js
Agares4: Yup, that's what I did in my updated version. I'll update the PR soon.

civihr_employee_portal/js/reports.js
kkalamarski: @AkA84 yeah, i somehow missed that it's the same array. Agree @up

civihr_employee_portal/templates/civihr-employee-portal-custom-reports.tpl.php
AkA84: I haven't looked into the specific of this table, but do you think it would be possible to have something in the html file like (just a very rough draft):
```html
<table data-summary-table-tpl>
    <thead>
        <tr>
            <th>Column 1</th>
            <th>Column 2</th>
            <th>Column 3</th>
            <th>Column 4</th>
        </tr>
    </thead>
    <tbody>
        <!-- tr to clone and fill with data -->
        <tr>
            <td>{column1Value}</td>
            <td>{column2Value}</td>
            <td>{column3Value}</td>
            <td>{column4Value}</td>
        </tr>
    </tbody>
</table>
```
which then you could get with javascript, clone, do the placeholder replacement with actual value and then put in the template?

Agares4: The number of columns is dynamic, so I think we have to create them in JS. I could also add handlebars or something here, but that would be an overkill.

kkalamarski:
As for me - Approved

civihr_employee_portal/js/reports.js
```diff
@@ -926,6 +926,74 @@
...
+
+            return normalisedSummary;
+        }
+
+        function replaceSummaryTable(summary, $DOM) {
+            var $countRow = $('<tr />'),
+                $headerRow = $('<tr />'),
+                $percentageRow = $('<tr />'),
+                summaryKeys = Object.keys(summary),
+                total = summaryKeys.reduce(function(accumulator, key) { return accumulator + summary[key]; }, 0);
```
kkalamarski: But I'd change this - keeping such things inline makes code harder to read. 


```javascript
var $countRow = $('<tr />'),
    $headerRow = $('<tr />'),
    $percentageRow = $('<tr />'),
    summaryKeys = Object.keys(summary),
    total;

total = summaryKeys.reduce(function(accumulator, key) {
 return accumulator + summary[key]; 
}, 0);
```

civihr_employee_portal/js/reports.js
```diff
@@ -926,6 +926,74 @@
...
+            type: 'GET',
+            url: this.getJsonUrl() + ( typeof path !== 'undefined' ? path : '' ),
+            success: function(data) {
+                var mappedResults = normaliseSummaryData(data.results);
+
+                replaceSummaryTable(mappedResults, this.$DOM);
+            }.bind(this)
+        });
+
+        function normaliseSummaryData(summary) {
```
AkA84:
```js
/**
 * Description of what this function does
 *
 * @param {type} paramName - description of the param
 * @return {type} description of what is returned
 */
function myFunction (param1, param2) {

}
```
^ for every function you wrote

civihr_employee_portal/js/reports.js
```diff
@@ -926,6 +926,74 @@
...
+
+            // The data can have two formats, depending on particular report:
+            // data = { results: [{ departament: "xxx", count: 1 }, ...]) }
+            // or
+            // data = { results: [{ gender: "aaa" }, { gender: "aaa" }, ...] }
+            //
+            // The code inside if takes care of the first possibility
+            // The else takes care of the second
+            if (summary[0] !== undefined && summary[0].data.count !== undefined) {
+                for(i in summary) {
```
AkA84:
```js
for (i in summary) {
```
sorry, code syntax needs to be consistent

civihr_employee_portal/js/reports.js
AkA84: Agree. To be honest I still haven't decided about multi-vars declaration style

This:
```js
var a = 'a';
var b = 1;
var c = {};
```
or this
```js
var a = 'a', b = 1, c = {};
```
or this
```js
var a = 'a',
    b = 1,
    c = {};
```
or even this (mimics hoisting)
```js
var a, b, c;

a = 'a';
b = 1;
c = {};
```
I usually tend to use the first format, but i'm a bit all over the place with this

civihr_employee_portal/js/reports.js
```diff
@@ -926,6 +926,74 @@
...
+                i;
+
+            // The data can have two formats, depending on particular report:
+            // data = { results: [{ departament: "xxx", count: 1 }, ...]) }
+            // or
+            // data = { results: [{ gender: "aaa" }, { gender: "aaa" }, ...] }
+            //
+            // The code inside if takes care of the first possibility
+            // The else takes care of the second
+            if (summary[0] !== undefined && summary[0].data.count !== undefined) {
```
AkA84: What if summary[0] is `undefined` because there is no actual data returned? Does your code handles this scenario?

civihr_employee_portal/js/reports.js
Agares4: sure, didn't notice this one :)

civihr_employee_portal/js/reports.js
```diff
@@ -926,6 +926,74 @@
...
+                for(i in summary) {
+                    normalisedSummary[summary[i].data.department] = +summary[i].data.count;
+                }
+            } else {
+                normalisedSummary = summary.map(function (row) {
```
AkA84: Super-terse alternative (to be checked, not 100% sure :smile:)
```js
normalisedSummary = summary.reduce(function (accumulator, row) {
    var key = row.data.gender;

    (typeof accumulator[key] === undefined) && (accumulator[key] = 0)

    return ++accumulator[key] && accumulator;
}, {});
```
but the point is that it doesn't seem you really need that `map` intermediate step, and also you are always incrementing by 1, so what's point of having `x:1`?

civihr_employee_portal/js/reports.js
```diff
@@ -926,6 +926,74 @@
...
+                    return { y: row.data.gender, x: 1 }
```
AkA84: What's this hardcoded property? Are we sure this whole thing will be about gender?

civihr_employee_portal/templates/civihr-employee-portal-custom-reports.tpl.php
AkA84: Nah, if it's to be done, for now it's via vanilla javascript with the hope that in the future we can plugin angular

civihr_employee_portal/js/reports.js
Agares4: The hardcoded properties I used are the actual property names that are used. They probably should have names like "xaxis" or "yaxis", but that would require changing all the report-related drupal views and fixing a lot of javascript, so I'm leaving it as is, at least for now.

AkA84: Ok, leave a comment there explaining this

civihr_employee_portal/js/reports.js
Agares4: Added "if" for that scenario here: https://github.com/compucorp/civihr-employee-portal/pull/94/files#diff-ea2a489336a69f400c3718abc2d8d03aR935

AkA84:
Approved

PR #95, opened 2016-01-15
PCHR-752: Remove prefix from Staff Directory Search

nishant-bhorodia:
Removed Prefixes (Mr. Mrs, Dr. etc.) from staff directory search autocomplete


Agares4:
Approved

PR #97, opened 2016-01-19
UY-24: Donot send cancellation email to Old Manager

nishant-bhorodia:
Issue:
The cancellation emails were getting sent to old managers as well.

Solution:
Modify function `function _getManagerContacts($contact_id)` to return only active managers.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -5314,7 +5314,10 @@ function _getManagerContacts($contact_id) {
     // If Leave approver is find, assign him as the manager (add contact ID to $assigned_manager_contact_ids array)
     foreach ($contactRelationships as $key => $relation) {
         if ($relation['relation'] == variable_get('relationship_name_to_use', 'has Leave Approved by')) {
-            if($relation['is_active'] == 1){
+            if($relation['is_active'] == 1
+               &&
+               ($relation['end_date'] >= date('Y-m-d') || $relation['end_date'] == ''))
+            {
                 $assigned_manager_contact_ids[] = $relation['contact_id_b'];
                 $manager_found++;
             }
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -4557,7 +4557,10 @@ function _getManagerContacts($contact_id) {
     // If Leave approver is find, assign him as the manager (add contact ID to $assigned_manager_contact_ids array)
     foreach ($contactRelationships as $key => $relation) {
         if ($relation['relation'] == variable_get('relationship_name_to_use', 'has Leave Approved by')) {
-            if($relation['is_active'] == 1){
+            if($relation['is_active'] == 1
```
Agares4: Can we extract the condition to a function? So the if here would be like: if(activeAndCurrent($relation)) { ... }
I think it'd be more readable this way.

Agares4: And are you sure these date comparisons are always going to work fine? We'd be more safe having something like strtotime($relation['end_date']) >= time()

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -4582,3 +4583,18 @@ function _getManagerContacts($contact_id) {
...
 }
+
+/**
+ * Function to determine if given manager-contact relationship is inactive or active
+ *
+ * @param array $relation
+ * @return boolean
+ */
+function relationActiveAndCurrent($relation){
+  if($relation['is_active'] == 1
```
Agares4: You can drop the if here and just `return $relation['is_active'] == 1 && ($relation['end_date'] >= date('Y-m-d') || $relation['end_date'] == '')`

Agares4:
Approved

PR #98, opened 2016-01-20
UY-22: Expose Emergency Contact ID field in drupal Views

nishant-bhorodia:
Expose `id` field from civicrm table `civicrm_value_emergency_contacts_21` to drupal views.


Agares4:
Approved

PR #100, opened 2016-01-20
PCHR-759: Show leave request for past managers

nishant-bhorodia:
PR addresses the following issue:
When the manager of a contact is changed, the new manager is not able to view the absences that were assigned to the past manager and hence not able to accept/reject those absences either.


Agares4:
Approved

PR #103, opened 2016-01-26
PCHR-817: cancelled leaves on calendar

nishant-bhorodia:
Removed cancelled leaves from "Manager Absence Approval" block and "My Leave" block on SSP


coldrunKacper:
Approved

PR #106, opened 2016-02-05
PCHR-818: Fetch absences only for current year

nishant-bhorodia:
Fetch absences belonging to current year only for Manager Calendar Block


Agares4:
Approved

PR #107, opened 2016-02-08
PCHR-830: Donot show leave requests for future manager

nishant-bhorodia:
Modified `function relationActiveAndCurrent` to return `true` if start date for relationship is in the past or is empty.


coldrunKacper:
Approved.

PR #110, opened 2016-03-24
PCHR-798: Improve my leave block default absence period to the current one

omar-compucorp:
# Problem 

My leave block absence period selector was not pointing to the current period.

# Solution

I fixed it to set the default option to the current period based on the current date , and in case the current date do not fit into any absence period it will default to the latest period in term of End Date.

selection_002

selection_001


AkA84:
I think you could squash those two commits together

civihr_employee_portal/views/includes/civihr_employee_portal_filter_absence_period.inc
```diff
@@ -55,5 +55,26 @@ class civihr_employee_portal_filter_absence_period extends views_handler_filter_
       $this->options['expose']['label'] = '';
     }
   }
+
+  /**
+   * Get the default period id from a list of periods based on current time
+   */
```
AkA84: Please also document the parameters and return values

coldrunKacper:
Yup, please document method's parameters and return values. Everything else looks fine.

coldrunKacper:
Approved.

PR #111, opened 2016-04-01
PCHR-851: Removes rejected leaves from My Leave and Manager Absence Approval calendars

davialexandre:
Both calendars shouldn't display rejected leaves. I've changed the view used to generate the My Leave calendar and the the query used to generate the Manager Absence Approval calendar to filter out those leaves.

Bellow are some screenshots showing the results:

Given 3 leave requests (one rejected, one cancelled and one waiting approval)
captura de tela 2016-04-01 as 15 50 23

Before the change, both calendars displayed the rejected requests:
captura de tela 2016-04-01 as 15 50 15
captura de tela 2016-04-01 as 15 53 42

After the change, the rejected requests are no longer displayed on both calendars:
captura de tela 2016-04-01 as 15 52 15
captura de tela 2016-04-01 as 15 49 45

### DIFF
--- a/civihr_employee_portal/src/Blocks/ManagerCalendar.php
+++ b/civihr_employee_portal/src/Blocks/ManagerCalendar.php
@@ -22,8 +22,23 @@ class ManagerCalendar {
         $managerData = get_civihr_uf_match_data($uid);
         $managerId = $managerData['contact_id'];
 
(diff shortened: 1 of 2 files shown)


coldrunKacper:
Approved.

PR #113, opened 2016-04-19
PCHR-938: Add Leave & Absences CiviHR permissions

davialexandre:
The following permissions were added:
- Access CiviHR Leave and Absences
- Manage CiviHR Leave and Absences

Here you can see the permissions available on Drupal:
captura_de_tela_2016-04-19_as_08_40_33


omarabuhussein:
Approved.

PR #114, opened 2016-04-20
PCHR-571: Move the toil button to the left

danilo-compucorp:
pchr-571

### DIFF
--- a/civihr_employee_portal/src/Blocks/AbsenceLinks.php
+++ b/civihr_employee_portal/src/Blocks/AbsenceLinks.php
@@ -16,13 +16,14 @@ class AbsenceLinks {
         $links = '';
         $links .= '<div class="chr_panel__footer">';
         $links .= '<div class="chr_actions-wrapper">';
+        $links .= civihr_employee_portal_make_link(t('Request TOIL'), 'credit', '', 'pull-left');
         $links .= '<div class="chr_actions-wrapper__inline-duo">';
-        $links .= civihr_employee_portal_make_link(t('Request TOIL'), 'credit');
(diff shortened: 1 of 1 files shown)


AkA84:
Please provide also  before/after screenshots for tablet and mobile viewports

civihr_employee_portal/src/Blocks/AbsenceLinks.php
```diff
@@ -16,13 +16,14 @@ public function generateBlock() {
         $links = '';
         $links .= '<div class="chr_panel__footer">';
         $links .= '<div class="chr_actions-wrapper">';
+        $links .= civihr_employee_portal_make_link(t('Request TOIL'), 'credit', '', 'pull-left');
```
AkA84: You moved this outside of `.chr_actions-wrapper__inline-duo`, so probably the end result in smaller viewports has changed, please check that

danilo-compucorp:
pchr-571-tablet
pchr-571-mobile

AkA84:
Ok so the end result in mobile is different. Is it possible to avoid it? (so keepining "Request Toil" and "Request Leave" on the same line with "Use TOIL" in a new line)

danilo-compucorp:
Hmm not sure if I agree / understand. 
If the "Request TOIL" is supposed to be "apart" from the other 2 buttons, why do we need to keep it side-by-side with "Request Leave" and "Use TOIL" in a new line ?

AkA84:
Totally fair point, I agree. Please point it out though in your ticket when you send it to testing

PR approved

PR #115, opened 2016-04-26
PCHR-997: new permission - manage roles and Teams

omarabuhussein:
"**administer roles and teams**" are added to be enabled by default for "**civihr_admin**" and "**administer**" Drupal roles.

selection_006


nishant-bhorodia:
Approved!

PR #116, opened 2016-04-27
PCHR-998: View My details menu item to users with newly created roles

omarabuhussein:
If someone created a new role and added a user to that role , then that user will not be able to see "My details" menu item since the roles allowed to view that menu item is hard-coded to limited set of roles which are :
**('civihr_staff', 'civihr_manager', 'civihr_admin', 'administrator' ) .**

The code changed to fetch all existing roles except the default drupal roles **( 'anonymous user', 'authenticated user' )** and updating "access arguments" for my details menu item to use any of these fetched roles.

**-- Before**
selection_024

**-- After**

selection_025


coldrunKacper:
Approved.

PR #117, opened 2016-05-05
PCHR-1007 - Individual Report Data section

brunoqueiros:
Created a template file for the table's view and shows a header with the total results. It was made some changes on the nav tabs to adjust to civihr's design and added some classes on the pagination element.

The styles are in the [PR-59](https://github.com/compucorp/civihr-employee-portal-theme/pull/59)

### DIFF
--- a/civihr_employee_portal/templates/civihr-employee-portal-civihr-report-custom.tpl.php
+++ b/civihr_employee_portal/templates/civihr-employee-portal-civihr-report-custom.tpl.php
@@ -6,15 +6,15 @@
 </div>
 <?php endif; ?>
 
-<ul class="nav nav-tabs nav-justified report-tabs panel chr_tabs__header">
+<ul class="nav nav-tabs nav-justified nav-tabs-header report-tabs">
 <?php if (!empty($tableUrl)): ?>
-    <li role="presentation" class="chr_tabs__header-item active"><a class="btn btn-default" data-tab="data">Data</a></li>
+    <li role="presentation" class="active"><a class="btn btn-default" data-tab="data">Data</a></li>
 <?php endif; ?>
 <?php if (!empty($jsonUrl)): ?>
-    <li role="presentation" class="chr_tabs__header-item"><a class="btn btn-default" data-tab="pivot-table">Pivot Table</a></li>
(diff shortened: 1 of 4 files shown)


civihr_employee_portal/templates/civihr-employee-portal-civihr-report-custom.tpl.php
```diff
@@ -6,25 +6,27 @@
 </div>
 <?php endif; ?>
 
-<ul class="nav nav-tabs nav-justified report-tabs">
+<ul class="nav nav-tabs nav-justified report-tabs panel chr_tabs__header">
```
AkA84: Just realized now that these are already Bootstrap's tabs. Why the need for a custom component (`.chr_tabs__header`)? Are the default tabs already in use anywhere in the ssp? If no, couldn't you customize those by overriding the default style? If yes, couldn't you add a modifier (say, `.nav-tabs-header`, and without adding any new classes for children element)?

AkA84: Also, why mixing a `panel` with `nav-tabs`? Is it standard practice in Bootstrap?

civihr_employee_portal/templates/civihr-employee-portal-civihr-report-custom.tpl.php
```diff
@@ -6,25 +6,27 @@
...
 
-<div class="report-content">
+<div class="report-content panel">
 <?php if (!empty($tableUrl)): ?>
     <div class="report-block data">
         <h4>Data</h4>
         <div id="reportTable"><?php print $table; ?></div>
 <?php if (!empty($exportUrl)): ?>
-        <a href="<?php print $exportUrl; ?>" id="export-report" class="btn btn-primary btn-default">Export</a>
+        <div class="export-button-wrap">
```
AkA84: This blue-button-on-bottom-right-corner thing is already covered by the {{.chr_panel}} component and in particular by the {{.chr_panel__footer}} element. You can see it in action in most of the panels in the SSP dashboard. Please take a look if you can reuse it and scrap this custom element altogether

AkA84:
Approved

PR #118, opened 2016-05-05
PCHR-1017 - Fixing Data Table pagination ajax issue by refreshing JS …

coldrunKacper:
Refreshing JavaScript View object in Drupal.views.instances array after data tabe is reloaded so pagination links are bound to proper ajax handlers.

### DIFF
--- a/civihr_employee_portal/js/reports.js
+++ b/civihr_employee_portal/js/reports.js
@@ -22,7 +22,6 @@
     };
 
     /**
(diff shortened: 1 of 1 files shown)


brunoqueiros:
It's ok, you just need comment these methods getReportTableDomID, getReportTableDomID.

brunoqueiros:
Approved!

PR #120, opened 2016-05-09
PCHR-1028 - Adding custom templates into SSP module.

coldrunKacper:
Adding Page, Exposed Filters and Data Table templates into SSP module so they can be used on custom Reports pages.


omarabuhussein:
( :+1: )

PR #123, opened 2016-05-19
PCHR-1071 - catching civihr_employee_portal_init() function exception and displaying as Drupal error message.

coldrunKacper:
In civihr_employee_portal_init() function we catch any Exception and display its message as Drupal error message.


nishant-bhorodia:
Approved.

PR #124, opened 2016-05-19
Pchr 1005 filters section

brunoqueiros:
**Before**
before

**After**
after

## Datepicker
To be able to use the datepicker component, it was necessary to transform the reports page into an angular app. It was added `bootstrap-civicrm-style.css` and `reqangular.min.js`, all the needed CSS to datepicker is inside `bootstrap-civicrm-style.css`.

To set up the HTML markup it was necessary to make changes on this PHP method `civihr_employee_portal_form_views_exposed_form_alter`.

## Fluid page
It was changed the page layout container to be fluid.


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -85,11 +85,12 @@ function civihr_employee_portal_init() {
 
     // Add our additional css libraries
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/css/custom.css");
+    drupal_add_css("sites/all/modules/civicrm/tools/extensions/civihr/org.civicrm.bootstrapcivicrm/css/bootstrap-civicrm-style.css");
```
AkA84: This could be rewritten as
```
drupal_add_css(drupal_get_path('module', 'civicrm') . "/tools/extensions/civihr/org.civicrm.bootstrapcivicrm/css/bootstrap-civicrm-style.css");
```
I am afraid that this is too specific, I think it can happen that the extensions will not be placed under `tools/extensions` (i believe one of our pilot projects has indeed this kind of setup), and as such this line will end up including a non-existing file

Can you check with the other guys (or even the core guys), if it's possible at all to have access to the extensions path in drupal? Or if there any other solution that could work in any scenario

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5486,6 +5488,50 @@ function civihr_employee_portal_form_views_exposed_form_alter(&$form, &$form_sta
             }
         }
     }
+
+    if (substr($form['#id'], 0, 33) === 'views-exposed-form-civihr-report-') {
```
AkA84: Why the `substr` with that specific index?

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5486,6 +5488,50 @@ function civihr_employee_portal_form_views_exposed_form_alter(&$form, &$form_sta
...
+    $form['submit']['#value'] = 'Filter';
+}
+
+/**
+ * Generate the necessary structure for the filter
+ *
+ * @param  [String] $title
+ * @return [array]
+ */
+function generateFilterInputStructure($title) {
```
AkA84: I don't think the name and the description of the function accurately reflects what the function actually does. Doesn't this function add a calendar button that is bound to the angular datepicker? If yes, then assign a name and function that describes that. Otherwise this is too generic

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5486,6 +5488,50 @@ function civihr_employee_portal_form_views_exposed_form_alter(&$form, &$form_sta
             }
         }
     }
+
+    if (substr($form['#id'], 0, 33) === 'views-exposed-form-civihr-report-') {
+        foreach ($form as $key => $value) {
+          if (isset($form['#info']["filter-$key"])) {
+            $form[$key]['value'] = generateFilterInputStructure($form['#info']["filter-$key"]['label']);
```
AkA84: I understand that you specifically point the `min` and `max` fields with the add-datepicker-button function, but what about `$form['#info']["filter-$key"]`? Is it safe to assume that any `filter-` field needs to have a datepicker?

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5486,6 +5488,50 @@ function civihr_employee_portal_form_views_exposed_form_alter(&$form, &$form_sta
...
+        }
+    }
+
+    $form['submit']['#value'] = 'Filter';
+}
+
+/**
+ * Generate the necessary structure for the filter
+ *
+ * @param  [String] $title
```
AkA84: No need for square bracket
```
* @param string $title <description>
* @return array
*/
```
Other than giving a more specific description, also please provide a description to $title

civihr_employee_portal/js/reports.js
```diff
@@ -331,13 +330,57 @@
...
+            'common/filters/angular-date/format-date'
+        ], function() {
+            angular.module('hrr', [
+                'ngAnimate',
+                'ui.bootstrap',
+                'common.angularDate'
+            ]).controller('FiltersController', function($scope, $timeout) {
+                $scope.format = 'dd/MM/yyyy';
+
+                $('[ng-model]').each(function () {
```
AkA84: Angular controllers should *never* have any kind of DOM access or manipulation

AkA84: If what you are trying to do is to set up the scope property that controls the `is-open` property of the datepicker, you can set everything up in the view, please look [here](https://github.com/civicrm/civihr/blob/develop/uk.co.compucorp.civicrm.appraisals/views/dashboard.html#L214-L217) for an example:
```html
<input
    type="text"
    class="form-control"
    id="manager-due-to"
    datepicker-popup
    ng-model="dashboard.filters.cycle_manager_appraisal_due.to"
    ng-init="dashboard.calendars.managerTo = false"
    ng-click="dashboard.calendars.managerTo = !dashboard.calendars.managerTo"
    is-open="dashboard.calendars.managerTo">
```
so basically you would do this in your `generateFilterInputStructure` function

civihr_employee_portal/js/reports.js
```diff
@@ -331,13 +330,57 @@
...
+                    $scope[$(this).attr('is-open')] = false;
+                });
+
+                $scope.open = function(id) {
+                    if ($('#bootstrap-theme').length === 0) {
+                        $('.datepicker-popup').wrap('<span id="bootstrap-theme" />');
```
AkA84: Why adding the `bootstrap-theme` namespace via js? Can't you add it to the html itself? (And again, you shouldn't do any DOM manipulation in the controller)

civihr_employee_portal/js/reports.js
```diff
@@ -331,13 +330,57 @@
...
+        var self = this;
+
+        require([
+            'common/angular',
+            'common/services/angular-date/date-format',
+            'common/directives/angular-date/date-input',
+            'common/moment',
+            'common/filters/angular-date/format-date'
+        ], function() {
+            angular.module('hrr', [
```
AkA84: Let's give a more meaningful name to the modules

civihr_employee_portal/templates/civihr-employee-portal-civihr-report-custom.tpl.php
```diff
@@ -1,98 +1,111 @@
-<?php if (!empty($filters)): ?>
-<a id="expose-filters-btn" class="btn btn-primary btn-default">Expose filters &raquo;</a>
-<a id="collapse-filters-btn" class="btn btn-primary btn-default hidden">Collapse filters &laquo;</a>
-<div id="report-filters" class="hidden">
-    <?php print render($filters); ?>
+<div id="hrr">
+  <?php if (!empty($filters)): ?>
+  <div id="js-filters-wrapper" class="panel panel-pane pane-block chr_panel chr_panel--no-padding panel--sliding-body">
+      <div ng-controller="FiltersController" class="pane-content">
```
AkA84: Please use the `controller as` syntax
```html
<div ng-controller="FiltersController as filters">
```

civihr_employee_portal/js/reports.js
```diff
@@ -306,18 +309,14 @@
...
-            CRM.$('#collapse-filters-btn').removeClass('hidden');
-            CRM.$('#report-filters').removeClass('hidden');
-        });
-        CRM.$('#collapse-filters-btn').bind('click', function(e) {
-            e.preventDefault();
-            CRM.$(this).addClass('hidden');
-            CRM.$('#expose-filters-btn').removeClass('hidden');
-            CRM.$('#report-filters').addClass('hidden');
+
+            $('#js-filters-wrapper').toggleClass('panel--sliding-body');
```
AkA84: Is all this to create the collapse/uncollapse effect? If yes, you think that by using the `collapse` directive from the angular's ui-bootstrap module would help simplify things? For an example look [here](https://github.com/civicrm/civihr/blob/develop/uk.co.compucorp.civicrm.appraisals/views/dashboard.html#L74-L84)
```html
<span style="cursor: pointer;" ng-click="dashboard.filtersCollapsed = !dashboard.filtersCollapsed">
    <i ng-class="{ 'fa-angle-right': dashboard.filtersCollapsed, 'fa-angle-down': !dashboard.filtersCollapsed }"></i>
    Filter Appraisals
</span>
<div collapse="dashboard.filtersCollapsed">
```

civihr_employee_portal/js/reports.js
```diff
@@ -331,13 +330,57 @@
...
+                $('[ng-model]').each(function () {
+                    $scope[$(this).attr('is-open')] = false;
+                });
+
+                $scope.open = function(id) {
+                    if ($('#bootstrap-theme').length === 0) {
+                        $('.datepicker-popup').wrap('<span id="bootstrap-theme" />');
+                    }
+
+                    $timeout(function() {
```
AkA84: Why the `$timeout`?

civihr_employee_portal/js/reports.js
```diff
@@ -331,13 +330,57 @@
...
+                        $scope[id] = true;
+                    });
+                };
+
```
AkA84: Unnecessary empty line

civihr_employee_portal/civihr_employee_portal.module
brunoqueiros: Because there isn't a specific `$form['#id']` value.
At http://localhost:8900/reports/people the value is `views-exposed-form-civihr-report-people-default`.
At http://localhost:8900/reports/leave_and_absence/ the value is `views-exposed-form-civihr-report-leave-and-absence-default`.

danilo-compucorp: Without knowing all the details on this context, I'd use something more generic like: `if( strpos($form['#id'], 'views-exposed-form-civihr-report-') !== false)` .

brunoqueiros: https://github.com/compucorp/civihr-employee-portal/pull/124/commits/71bcc6e4177b031c499c58f22488625db7b23341

danilo-compucorp:
Looks OK. PR approved.

civihr_employee_portal/js/reports.js
```diff
@@ -331,13 +323,51 @@
...
+        var self = this;
+
+        require([
+            'common/angular',
+            'common/services/angular-date/date-format',
+            'common/directives/angular-date/date-input',
+            'common/moment',
+            'common/filters/angular-date/format-date'
+        ], function() {
+            angular.module('civihrReports', [
```
danilo-compucorp: 'use strict';

civihr_employee_portal/js/reports.js
```diff
@@ -331,13 +323,51 @@
     }
 
     /**
+     * Init angular in reports custom page,
+     * and add the calendar icon
+     *
+     */
+    HRReport.prototype.initAngular = function() {
+        var self = this;
```
danilo-compucorp: Avoid using the `self` keyword. Modern browsers provide a global `self` [variable](https://developer.mozilla.org/en-US/docs/Web/API/Window/self) , and that can cause confusion. I've seen people call it `that`

brunoqueiros: That variable has been not used, I'll remove it.

danilo-compucorp:
Looks OK. PR approved

PR #125, opened 2016-05-24
PCHR-1004: Reports list page

brunoqueiros:
It was styled the reports list page to be like the PSD and added a collapse effect on 'Reports Settings' block.

**Before**
image

**After**
anim

## civihr-employee-portal-civihr-report-landing-page.tpl.php
It was overridden the markup, and created a method `setTitleMarkup` to set the title markup.

## civihr_employee_portal.module
It was fixed the variable `$base_url` position, it was been using before be defined. And was added the `report.js` and `reqangular.min.js` into /reports

## reports.js
Created a controller for the settings block, with `isCollapsed` property.


danilo-compucorp:
Looks ok. Approved

PR #126, opened 2016-05-25
PCHR-1122: Reports settings - Age group

brunoqueiros:
It was styled the _reports settings - age group_ page using the default styles `.panel` and `table`, also was added a panel header to show the results total.

**Before**
image

**After**
image

### DIFF
--- a/civihr_employee_portal/templates/civihr-employee-portal-civihr-report-settings-age-group.tpl.php
+++ b/civihr_employee_portal/templates/civihr-employee-portal-civihr-report-settings-age-group.tpl.php
@@ -1,75 +1,78 @@
-<div class="chr_table row">
-    <div class="chr_table__table-wrapper col-md-12">
-        <div class="chr_table-w-filters__table">
-            <table >
-                <?php if (!empty($title) || !empty($caption)) : ?>
-                    <caption><?php print $caption . $title; ?></caption>
-                <?php endif; ?>
-                <?php if (!empty($header)) : ?>
-                    <thead>
-                        <tr>
-                        <?php
-                          foreach ($header as $field => $label):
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/templates/civihr-employee-portal-civihr-report-settings-age-group.tpl.php
```diff
@@ -1,75 +1,78 @@
...
-                <?php endforeach; ?>
-                </tbody>
-            </table>
-        </div>
+<div class="panel-pane pane-block chr_panel chr_panel--no-padding">
+  <div class="pane-content">
+    <div class="chr_search-result__header">
+      <div class="chr_search-result__total">
+        Results
+        <span class="chr_search-result__total__count"><?php echo count($rows); ?></span>
```
danilo-compucorp: Is it possible to fix this class name, following the BEM naming convention?

brunoqueiros: At the moment no, I just reused the existing classes. To change to BEM I'll need to test all others pages that are using it to check if will not broken.

danilo-compucorp: Ok I understand the issue, and won't block the PR because of this. 
But I believe we need to plan some progressive refactorings to fix this kind of thing, this clearly tries to follow BEM, but it's [incorrect](http://getbem.com/faq/#css-nested-elements).

danilo-compucorp:
Approved

jamienovick:
good job bruno without the designs, thanks

PR #127, opened 2016-05-25
PCHR-1106 - Customize 'value between dates' View Handler to handle NULL values and custom input date format.

coldrunKacper:
Modifying views_between_dates_filter_handler.inc custom Drupal View Handler allowing to specify a date value between two date fields (columns) and adjusting it to allow NULL dates (optionally) and custom format of input date.

### DIFF
--- a/civihr_employee_portal/views/includes/views_between_dates_filter_handler.inc
+++ b/civihr_employee_portal/views/includes/views_between_dates_filter_handler.inc
@@ -100,6 +100,9 @@ class views_between_dates_filter_handler extends date_views_filter_handler_simpl
 
   /**
    * Get date value formatted according to date format set up by input field.
+   * If provided date contains slashes separated values (as in Reports exposed 
+   * data filter) then we assume 'd/m/Y' format and then convert it into
+   * standard MySQL date (Y-m-d). Otherwise original date is returned.
    * 
    * @param string $date
    * @return type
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/views/includes/views_between_dates_filter_handler.inc
```diff
@@ -60,26 +76,48 @@ class views_between_dates_filter_handler extends date_views_filter_handler_simpl
...
       }
       // Else not between
       else {
         $this->query->add_where_expression($group, "$start_field > $placeholder OR $end_field < $placeholder", array($placeholder => $value));
       }
    }
+  }
 
+  /**
+   * Get date value formatted according to date format set up by input field.
```
davialexandre: Can you elaborate this comment a little more? What is the date format set up by the input field? Why is the original date returned if $parts == 1?
Is not clear from and to what format the dates are converted. Assuming the original date is in dd/mm/yyyy format, what about changing the code on lines 108 to something like?:
```php
list($day, $month, $year) = explode('/', $date);
```

coldrunKacper: In this case 'get_date_value_formatted()' method returns date by assuming that input date is in 'd/m/Y' format (as it's currently defined in front-end datepicker) (so that's why there is 'explode' stuff). In the other hand, when we don't have datepicker customizing our input format, the most reasonable date format seem to be 'Y-m-d' (for example by using input date filter on View Admin page). Anyway the best solution would be to use the same date format defined for Drupal on front-end and back-end but I have to talk with Bruno and ask him to implement Drupal date format.
Regarding the list() function - it causes PHP notice message if the input array is too short so I wanted to bypass this case.

PR #128, opened 2016-05-30
Creating custom Aggregator (Count Unique Values (not empty)) for Pivot Table.

coldrunKacper:
I've created "Count Unique Values (not empty)" Pivot Table Aggregator which works similarly to "Count Unique Values" but doesn't count empty values such as NULL, 0 or empty string. This allows to count unique Role IDs excluding empty Role IDs which appear on Pivot Table where Job Contract doesn't have any Role created.


jamienovick:
I think a better name might be "count unique values excluding null, 0 or empty"

omarabuhussein:
Approved

PR #133, opened 2016-06-14
PCHR-1167 - Implementing ajax callbacks handling for State/Province Emergency Contact webform.

coldrunKacper:
- Using hook_form_alter for implementing ajax callbacks on State/Province fields.
- Using hook_node_view for implementing default values of State/Province fields.
- Creating a function returning a list of State/Province key-value pairs by given Country ID.
- Creating callback functions for handling ajax refresh of State/Province fields after changing Country field values.


nishant-bhorodia:
Approved!

PR #138, opened 2016-06-20
PCHR-920 Custom module to load civicrm resource drupal

vidhatanand:
The condition to select js/css when wildcard parameters (*.css / *.js )was being passed was not correct. The css files were being added as js scripts too. 
Added an additional condition to check filenames with regex.


coldrunKacper:
Approved.

PR #139, opened 2016-06-24
PCHR-1136: Handle sickness absence with different name

AkA84:
#### Problem
The "Sick" absence type is a special case of an absence, as it is displayed separately from the other types and it has a dedicated modal to request it.

The problem is that in order to distinguish it from the other types, we rely on its name, which is always assumed to be "Sick"

```php
if ($absence_type['title'] == 'Sick') {
  // ...
}
```
That means that the moment someone changes the name even slightly, say from "Sick" to "Sick Leave", the system breaks

before

#### Solution
The solution is still temporary, and only because we have a new Leave&Absences system coming in the near future. Basically instead of checking for the exact "Sick" name, we just check if "Sick" is present
```php
return preg_match('/^sick/i', $absenceName);
```
Also some views had to have the filters updated to make them less strict about the absence type name

after

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -1317,7 +1317,7 @@ function _get_civicrm_contact_id_by_drupal_uid($uid) {
 
 /**
  * Get node ID (nid) by its title and type. If no node is found, return null.
- * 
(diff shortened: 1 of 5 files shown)


vidhatanand:
Approved.

PR #140, opened 2016-06-27
PCHR-1158: Add js to trigger auto-submit on views data on My leave block.

vidhatanand:
The views data field has  an embedded view which is supposed to inherit the contextual filters. Looks like something wrong with auto submit. The auto submit is not called on first page load. Fixed that by forcing form to submit on page load.


omarabuhussein:
looks good ..

PR #142, opened 2016-06-30
PCHR-1166 : Add Maxlength validation on postal code fields in webforms. 

vidhatanand:
1. Removed a warning, "Strict warning: Only variables should be passed by reference in civihr_employee_portal_my_details_callback() (line 1532 of "

2.The address field was not saving when postal_code field failed validation on CiviCRM side.
Adding a field length to this field in webform solves the issue.

On trying to postal code of more than 12 characters saving, we received  this error:
CiviCRM API Error : The CiviCRM "address create" API function returned the error: "Value for postal_code is 13 characters - This field has a maxlength of 12 characters."

Deployment Notes : After pull code, below steps to be executed
- drush refresh-node-export-files
- Update the new nid of "My details" form and "Emergency contact" form on admin/config/system/site-information


coldrunKacper:
Approved.

PR #145, opened 2016-07-06
PCHR-959: Set drupal variables when creating webforms

AkA84:
When using `drush refresh-node-export-files`, the `nid`s of the newly generated webforms weren't then set as variable automatically, this PR fixes that


coldrunKacper:
Approved.

PR #146, opened 2016-07-07
Pchr 1176 ssp task block enter name filter not working

danilo-compucorp:
The "Enter Name" filter wasn't working properly:
before

It's been fixed and refactored to consider case insensitive searches as well:
after

Also, when the user is in the _civihr_manager_ role, the filter will only show reportees (people that have "Line Manager" relationships with the current user).


coldrunKacper:
Approved.

PR #147, opened 2016-07-07
PCHR-860: View leave type title instead of name

omarabuhussein:
My leave block was viewing the name of leave types ( which may contain underscores ) instead of their titles 

**Before :**

selection_068

and now it's fixed 

**Now :**

selection_069


nishant-bhorodia:
Approved!

PR #148, opened 2016-07-11
PCHR-1005: Fix datepicker

brunoqueiros:
This PR update:
- the datepicker identifiers to work with the newest version;
- fix the bootstrapcivicrm path which is responsible to add the SSP style;
- add a different placeholder on the input, `yyyy-mm-dd` instead of `yyyy-MM-dd`;
- add the `ng-click` directivetive on the calendar icon;
- also it was updated the code indentation on `reports.js` to uses 2 spaces.

***Before***
image

***After***
image


AkA84:
Approved

PR #150, opened 2016-07-18
PCHR-1292 - Removing custom page overrided templates.

coldrunKacper:
Removing custom page.tpl.php handling (it was moved into theme repository and handled with a dynamic variable in the template file).


brunoqueiros:
Approved.

PR #151, opened 2016-07-19
PCHR-1092 - Excluding deleted contracts from SQL views.

coldrunKacper:
Excluding deleted Job Contracts when joining with Revision and Role data during SQL views creation.


davialexandre:
Approved

PR #154, opened 2016-07-26
PCHR-920: Changed the path for loading the files and refactor coding standards. 

hitesh-jain:
Problem : Stylesheets were not getting injected into Drupal through api callbacks and also the code written was not following proper Drupal coding standards.
https://compucorp.atlassian.net/browse/PCHR-949

Solution: Changed the path that was getting generated with Drupal ROOT in drupal_add_css function to remove path prefix and make that loaded like other files which are part of Drupal without Drupal ROOT path included . example it should be loaded like sites/all/modules/filename and not like https:civihr/sites/all/modules.
Fixed all the missing Drupal coding standards as we need to follow strict coding standards to make this an generic changes.

### DIFF
--- a/civicrm_resources/civicrm_resources.module
+++ b/civicrm_resources/civicrm_resources.module
@@ -12,50 +12,51 @@
  *   Machine path of civicrm extention.
  * @param array $req_files
  *   An array of file list to be included
+ *
  * @return NULL
  */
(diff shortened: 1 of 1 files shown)


hitesh-jain:
Closing as its not required to follow Fork workflow.

PR #157, opened 2016-08-01
PCHR-1341: Reverting breaking changes from PCHR-1176

danilo-compucorp:
**Issue:**
Error 500 when login with civihr_manager or civihr_staff  without tasks assigned. 

**Solution:**
Reverting the changes made in PCHR-1176 in order to solve the issue.


coldrunKacper:
Approved.

PR #161, opened 2016-08-05
SSP Reports

coldrunKacper:
Extending Self Service Portal with Reports feature basing on Drupal Views. By default it provides two Report pages containing People Report and Leave and Absence Report and also Reports settings page with Age groups management.

The modification provides two new Drupal permissions: "Access HR Reports" and "Manage HR Reports settings".
The main Reports page is created under "/reports" url and also is visible in top navigation for user which has permissions to access the Reports.
Each of Report page contains two tabs: Views Data as a Table and Pivot Table content.


AkA84:
Approve

PR #163, opened 2016-08-05
PCHR-1316: Remove unwanted lines

deb1990:
# Problem

After Radix theme update, 404 errors were seen in the console.

# Solution

Removed the unused files


AkA84:
Approved

PR #167, opened 2016-08-09
PCHR-1403: Remove reference to civihr_admin_theme

AkA84:
The civihr_admin_theme is being deleted, so any reference to it doesn't serve any purpose anymore (and the whitelist was empty to begin with)


omarabuhussein:
approved

coldrunKacper:
Approved.

PR #172, opened 2016-08-12
PCHR-1316: Modify Bootstrap filename

deb1990:
Problem: "bootstrap-civicrm-style.css" does not exist any more.
Fix: Changed it to 'bootstrap.css'


AkA84:
Approved

PR #173, opened 2016-08-12
PCHR-1452: Skipping API GET call on Contact if there is an empty array of Contact IDs.

coldrunKacper:
The issue appears on CiviCRM 4.7 (regarding to API changes) and causes an error when using
'id' => array('IN' => array())
API parameters with empty array (while CiviCRM 4.5 works fine with it).
I've added an additional check and now there is no call executed if ID array is empty.
Also I've removed '$rowContacts' variable initialization as it wasn't used in the script.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -5,6 +5,12 @@ use Drupal\civihr_employee_portal\Blocks\Leave;
 use Drupal\civihr_employee_portal\Forms\AbsenceRequestForm;
 use Drupal\civihr_employee_portal\Helpers\HelperClass;
 
+define('TASK_PAST_DAY', 1);
+define('TASK_TODAY', 2);
+define('TASK_DAY_AFTER_WEEKEND', 4);
+define('TASK_TOMORROW', 5);
+define('TASK_ANY_OTHER_DAY', 3);
+
 /**
  * Implements hook_css_alter().
  */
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6273,3 +6273,71 @@ function _is_hrreports_data_table_view($viewName) {
...
+ * 4 - DAY_AFTER_WEEKEND
+ * 5 - TOMORROW
+ * 3 - ANY_OTHER_DAY
+ * Used in Tasks block template.
+ * 
+ * @param string $date
+ * @return int
+ */
+function _get_task_filter_by_date($date) {
+    $today = date('Y-m-d');
```
nishant-bhorodia: Please can we use `DateTime` objects for all dates in this function for consistency or is there a specific reason why both `date` and `DateTime` are used ? Thank you!

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6273,3 +6273,71 @@ function _is_hrreports_data_table_view($viewName) {
...
+    $tomorrow = new DateTime('tomorrow');
+    $nbDay = date('N', strtotime($today));
+    $sunday = new DateTime($today);
+    $sunday->modify('+' . (7 - $nbDay) . ' days');
+    $weekEnd = $sunday->format('Y-m-d');
+    $taskDate = date('Y-m-d', strtotime(strip_tags($date)));
+
+    if ($taskDate < $today) {
+        return 1;
```
nishant-bhorodia: Perhaps use of constants will make it more readable.

omarabuhussein:
generally looks okay .. but i have one question and i know it's not directly related to your ticket .. If I understand correctly , It appear that the contact names are pre-fetched to be used in javascript for search .. but this doesn't scale very well and ajax calls should be used instead.

coldrunKacper:
It's difficult to say what is better practice as in this case we only want to search within currently shown records so applying ajax calls would cause returning the same set of data which we already have in the template.

omarabuhussein:
> we only want to search within currently shown records

if the number of shown records is small then i guess it's fine ... approved

PR #178, opened 2016-08-16
PCHR-1317: Remove task top filters

AkA84:
The top task filters on the SSP (My Tasks | Delegated Tasks | All) are not needed anymore, as the list should be filtered by default as if the "My Tasks" filter had been selected

The html elements and related javascript has then been removed, and the list had been filtered via php based on the value of the 'task_contacts_1' property (which represents which contact is assigned to the task)

_(Note: I'm not sure this is the best way to filter the list, as i think it would be possible to filter the list directly on the View/Block. Couldn't figure out how to do it though, so I went for the easiest route)_

**Before**
before

**After**
after

### Updates
The `civihr_employee_portal/views/templates/views-view-table--Tasks--block.tpl.php` file had a lot of inline javascript, which has now been moved in `js/tasks.js` under the  `Drupal.behaviors.civihr_employee_portal_tasks` object. The file is now included in the /dashboard page, via the `_init()` function

Speaking about the `_init()` function, it was in a really bad shape and thus had been refactored


civihr_employee_portal/views/templates/views-view-table--Tasks--block.tpl.php
```diff
@@ -40,19 +50,11 @@
...
-    if (strip_tags($row['task_contacts_2']) == $civiUser['contact_id'] && strip_tags($row['task_contacts_1']) != $civiUser['contact_id']):
-        $rowType = 'task-delegated';
-    endif;
-    if (!$rowType):
-        continue;
-    endif;
-    $contactsIds[strip_tags($row['task_contacts'])] = 1;
-    $taskFiltersCount[_get_task_filter_by_date($row['activity_date_time'])]++;
-    $taskFiltersCount[0]++;
+  if (taskAssignedToUser($row, $civiUser['contact_id'])) { continue; }
```
hitesh-jain: Should be
 if (taskAssignedToUser($row, $civiUser['contact_id'])) {
    continue;
  }
we don't follow shorthand format according to Drupal coding standards.

AkA84: > we don't follow shorthand format according to Drupal coding standards.

what would that be?

civihr_employee_portal/views/templates/views-view-table--Tasks--block.tpl.php
```diff
@@ -111,18 +113,10 @@
...
-                        $rowType = 'task-my';
-                    endif;
-                    if (strip_tags($row['task_contacts_2']) == $civiUser['contact_id'] && strip_tags($row['task_contacts_1']) != $civiUser['contact_id']):
-                        $rowType = 'task-delegated';
-                    endif;
-                    if (!$rowType):
-                        continue;
-                    endif;
+                    <?php
+                    if (taskAssignedToUser($row, $civiUser['contact_id'])) { continue; }
```
hitesh-jain: same as above, need to change structure.

civihr_employee_portal/views/templates/views-view-table--Tasks--block.tpl.php
hitesh-jain: Usually we avoid using inline code format with conditions like the one { continue; } should be changed to 
if ( condition ) {
  continue;
}

civihr_employee_portal/views/templates/views-view-table--Tasks--block.tpl.php
```diff
@@ -186,17 +176,13 @@ class="ctools-use-modal ctools-modal-civihr-custom-style ctools-use-modal-proces
     (function($){
```
hitesh-jain: Why we are adding javascript code in tpl.php file , this should go into some .js file.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -91,6 +91,183 @@ function _add_script_to_page($scriptFile, $page) {
...
+                        LEFT JOIN " . $civi_db_name . ".civicrm_hrabsence_type absence_type_credit
+                        ON request.activity_type_id = absence_type_credit.credit_activity_type_id
+                        INNER JOIN " . $civi_db_name . ".civicrm_uf_match drupal_contact
+                        ON tgt.contact_id = drupal_contact.contact_id
+                        WHERE (request.activity_type_id IN (" . rtrim($options, ',') . "))
+                        GROUP BY request.id");
+
+    // Set to false so it will not rebuild until not asked to rebuild
+    variable_set('rebuild_absence_list_view', 'FALSE');
+  } catch (Exception $e) {
```
hitesh-jain: Move : catch (Exception $e) to new line.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -91,6 +91,183 @@ function _add_script_to_page($scriptFile, $page) {
 }
 
 /**
+ * Rebuilds the absence list view
```
hitesh-jain: Can you please end all the documentation comments with period .

hitesh-jain:
Looks good :) thanks

PR #180, opened 2016-08-16
PCHR-1455 - Big notice message on staff directory

coldrunKacper:
Fixing notice messages in removeAdditionalCommas() and removeInactiveDepartments() module functions by adding validation for non empty Views values.


nishant-bhorodia:
Approved!

PR #183, opened 2016-08-19
PCHR-1249: Nav markup changed for consistency

deb1990:
**Problem**: "btn btn-default" classes were not necessary for the UI of the Tabs in http://civihr-dev.civihrhosting.co.uk/reports/leave_and_absence/.

**Fix**: Removed the classes and did minor changes to CSS in https://github.com/compucorp/civihr-employee-portal-theme/pull/95/

#### Problem
The block had a title which we don't need (because the page containing the block already has it), and also was using a layout that wasn't correct, resulting in the wrong paddings that would not make the block align to the title.
#### Solution
Removed the title, and used the one-column layout offered by Radix

#### Problem
Some of the buttons defined in the module templates were using `.btn-default` along with `.btn-primary`, while obviously it should be either one or the other
#### Solution
Removed `.btn-default` and left only `.btn-primary`


AkA84:
Approved

deb1990:
Approved

PR #185, opened 2016-08-23
PCHR-1497: Fix count of "Tomorrow" tasks in SSP block

AkA84:
## Problem
Tasks that had a due date of tomorrow weren't show in the "Due This Week" section of the "My Tasks" block in the SSP

This was due to the fact that the constant (`TASK_TOMORROW`) used to identify the filter type that a given task belonged to referenced an index (`5`) that didn't match any of the available filters:
```php
$taskFilters = array(
  0 => 'All',
  1 => 'Overdue',
  2 => 'Due Today',
  3 => 'Due This Week',
  4 => 'Later',
);
```
because "tomorrow" tasks are not grouped in a specific section, but are put together with the "Due This Week" tasks

## Solution
Initially I simply made `TASK_TOMORROW` reference the "Due This Week" index, `3`.

I thought went on and refactored this part a bit, by removing the `TASK_` constants altogether and by making the `$taskFilter` array more expressive in terms of key, and by putting both the label and the count of each type within it:
```php
$taskFilters = array(
  'all'       => array('label' => 'All', 'count' => 0),
  'overdue'   => array('label' => 'Overdue', 'count' => 0),
  'today'     => array('label' => 'Due Today', 'count' => 0),
  'week'      => array('label' => 'Due This Week', 'count' => 0),
  'later'     => array('label' => 'Later', 'count' => 0)
);
```
this made other optimizations/adjustments follow, both in php and js

### DIFF
--- a/civihr_employee_portal/views/templates/views-view-table--Tasks--block.tpl.php
+++ b/civihr_employee_portal/views/templates/views-view-table--Tasks--block.tpl.php
@@ -45,11 +45,11 @@ $taskFilters = array(
     3 => 'Due This Week',
     4 => 'Later',
 );
-$taskFiltersCount = array_combine(array_keys($taskFilters), array_fill(0, count($taskFilters), 0));
+$taskFiltersCount = array_fill(0, count($taskFilters), 0);
 $contactsIds = array();
 $contacts = array();
 
-foreach ($rows as $row):
+foreach ($rows as $row) {
   if (taskAssignedToUser($row, $civiUser['contact_id'])) {
     continue;
   }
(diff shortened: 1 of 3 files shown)


civihr_employee_portal/views/templates/views-view-table--Tasks--block.tpl.php
```diff
@@ -31,64 +31,76 @@ function taskAssignedToUser($task, $user_id) {
...
-$civiUser = get_civihr_uf_match_data($user->uid);
+/**
+ * Wraps the original function that fetches the filter type by task date so that
+ * the resulting filter, if necessary, can be normalized to match the list of filters
+ * offered in the UI ('tomorrow' and 'week' are grouped together)
+ *
+ * @param  [type] $filter [description]
+ * @return [type]         [description]
+ */
+function taskFilterByDate($taskDate, $normalize = false) {
```
coldrunKacper: The best way would be to create this function in php module file (or even modify '_get_task_filter_by_date' to accept this additional $normalize attribute). It's good to not define functions in template files in case that the template would be used twice on the same page (which would cause PHP error regarding double declaration of the same function).

coldrunKacper:
Approved.

PR #190, opened 2016-08-30
PCHR-1458: Fix error 500 when using the keywords filtes on HR Resources

davialexandre:
**Problem:**
Trying to search for HR Resources using a keyword would result on a blank page and a response with status code 500.

The source of this problem is the commit 8427db4c. On it, the hr_resources view was changed to load the short description instead of the body of the node, but the view's combine filter was still trying to use the old body field, resulting on the error.

**Solution**:
The combine filter was updated to use the short description field instead of the body.

Here you can see the filter working after the fix:
hr_resources


omarabuhussein:
approved

PR #192, opened 2016-09-01
PCHR-1398: My details menu permission fix

omarabuhussein:
## Problem 

When removing (View My Details) from a specific role , it will still be able to see ( My details ) menu and access it .

selection_112

## Solution 

Unlike other menu items which is controlled by views , (My Details ) view menu item is defined inside civihr_employee_portal_menu and the permission is given to it was based on roles instead of permission . So I'v changed it and it can be controlled by (View My Details) permission only.

I also fixed the access arguments for (Quick Approve Leave) ajax call which used to generate emails when leave request get approved , so now emails will get generated when an absence get approved by any user with  (access manager approval screen) permission.

Finaly , I've added a new permission ( change document status ) and who have it can change documents status on SPP in case the (document assignee ID) == user ID

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -434,6 +434,10 @@ function civihr_employee_portal_permission() {
       'title' => t('View HR Resources'),
       'description' => t('Availability for the user to view hr resources block and page')
     ),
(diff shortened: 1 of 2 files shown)


nishant-bhorodia:
Looks good to me @omarabuhussein . Approved :)

PR #194, opened 2016-09-02
PCHR-1540 - Reports pages issue (date filters and JSON requests)

coldrunKacper:
The warning messages was caused undefined values of dates in views_between_dates filter (when the date was not specified - especially right after page load) so I set the undefined dates to null (and then all null dates are converted to today dates by default). Also, I've created a module callback for getting JSON reports data (instead of calling View URLs directly) by ajax JS requests. It doesn't change much but creates one entry point for every JSON Reports requests.

- Creating custom callback for Reports JSON requests.
- Modifying People Reports contextual filters.
- Updating Feature module.


davialexandre:
The code looks good. Can you please update the PR to better describe what was the issue and what caused it and how your solution fixes it?

davialexandre:
Approved

PR #195, opened 2016-09-03
PCHR-1483: Fix missing job contract/roles fields on My Details and Staff Directory views

davialexandre:
The SSP Reports merge introduce a new structure to link the job contract and job role entities for views. They are now all linked through a revision.

These two views use these entities and had to have their relationships declarations updated in order to properly fetch the information. Since things are now linked though revisions, I also had to add contextual filters to the views to make sure it will return only the information related to the current revision.

I also made one small fix to the query that creates the hrjc_role view. It was joining with the option value using its ID, but recently this has changed to use the option value value. I changed the query to reflect this.

Below are the before/after screenshots of My Details and Staff Directory. Note that, on Staff Directory, we had a lot of contact on the list before the fix and just a few after. The reason is that, since the relationships were wrong, the query was not joining with the contract and was returning every contact available. After the fix, it only lists contacts with contracts:

### My Details
Before:
captura de tela 2016-09-02 as 20 48 55

After:
captura de tela 2016-09-02 as 20 46 37

### Staff Directory
Before:
captura de tela 2016-09-02 as 20 49 27

After:
captura de tela 2016-09-02 as 20 46 11


coldrunKacper:
Approved.

PR #198, opened 2016-09-06
PCHR-1550: Fix My Details views to show contact details even if it doesn't have a contract

davialexandre:
The contextual filter for the contract period start date was filtering out records without a start date (that is, contacts without a contract), making the My Details tab empty for such contracts.

This fixes the issue by allowing an empty start date. In that case, only the information related to the contact (Name, Work email and Work Phone) will be displayed and the contract related fields will be empty.

This is the My Details block before the changes, for an user without a contract:
captura de tela 2016-09-06 as 11 28 32

And this is after the fix:
captura de tela 2016-09-06 as 11 24 42

For comparison, this is how it looks like for a contact which has a contract:
captura de tela 2016-09-06 as 11 30 12


coldrunKacper:
Approved.

PR #199, opened 2016-09-08
PCHR-1245: Fix vacancy list to show location label instead of value

omarabuhussein:
When viewing vacancies , the location value appear instead of its label .
selection_138

That's because the location value which is retrieved with other vacancy details are not converted to label which i fixed in this PR by passing the Vacancies list to a new function **civihr_employee_portal_convert_array_option_value** that do the conversion .

selection_139

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -2422,6 +2422,12 @@ function civihr_employee_portal_get_vacancy_list() {
         'return' => "id,start_date,end_date,status_id,description,position,salary,location",
     ));
 
+    $result_output['values'] = civihr_employee_portal_convert_array_option_value(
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -2422,6 +2422,12 @@ function civihr_employee_portal_get_vacancy_list() {
...
 
+  $result_output['values'] = civihr_employee_portal_convert_array_option_value(
+    'hrjc_location',
+    'location',
+    $result_output['values']
+  );
+
     // Output json
     drupal_json_output($result_output);
     drupal_exit();
```
davialexandre: Can you please indent the rest of the function? It will look weird to have only the new code indented with two spaces

PR #200, opened 2016-09-19
PCHR-1521: Set default permissions

omarabuhussein:
The default permissions for the default civihr roles should follow this list : 

https://compucorp.atlassian.net/wiki/display/PCHR/Drupal+permissions+configuration .

So I've created a new drupal feature called "**CiviHR Default Permissions**" to handle the setup of the default permissions .

I also opened this related PR (https://github.com/civicrm/civicrm-buildkit/pull/274) against buildkit repo to remove the code that was used to set some permissions for some roles.


nishant-bhorodia:
Looks good to me . Approved!

coldrunKacper:
Approved.

coldrunKacper:
In .user_permission.inc there are permissions for specific roles such as "edit users with role 55120974" (the ones containing "with role" in the name) but in .info file I see only two of them. Is this correct?
Also can you check if on /admin/people/permissions they appear as "Edit users with role civihr_admin" instead of "edit users with role 55120974"?

omarabuhussein:
> Also can you check if on /admin/people/permissions they appear as "Edit users with role civihr_admin" instead of "edit users with role 55120974"?

@coldrunKacper that's fine , when you export these permissions it export them with role IDs since these permissions are dynamically generated based on existing roles .

omarabuhussein:
> In .user_permission.inc there are permissions for specific roles such as "edit users with role 55120974" (the ones containing "with role" in the name) but in .info file I see only two of them. Is this correct?

I don't know to be honest but I generated the info file via the UI and that was the output , I also tested by creating a new civihr installation via buildkit and stuff were fine as expected .

coldrunKacper:
Seems to be fine (as long as nothing "leaks" between two feature modules when updating / reverting). Approved.

PR #201, opened 2016-09-20
PCHR-1576: Add new menu buttons

deb1990:
# Requirement

Add menu items for CiviHR SSP and CiviHR Admin which would be visible on only mobile

# Image

anim
Note: Showing the menu only in mobile is taken care by 
https://github.com/compucorp/civihr-employee-portal-theme/pull/100


JonScreat:
I think it would be worth differentiating the Admin and SSP links from the others as they serve a different purpose and so should sit differently in the hierarchy. Is this something that's possible to do?

AkA84:
@JonScreat i think your comment should be in this PR https://github.com/compucorp/civihr-employee-portal-theme/pull/100

PR #209, opened 2016-09-27
PCHR-1529 - Creating Reports configuration section

coldrunKacper:
- Creating a new configuration section on Reports page:
- The configuration is stored in db table (created as with Features upgrade function).
- Configuration data contains any Pivot Table setting which can be managed by Pivot Table UI (rows, columns, aggregators, renderers, field filters).
- Configuration interface contains a select field with available configurations for current report, and buttons to save (overwrite) currently selected configuration, save currently active configuration as new (with prompt alert which allows to enter configuration label) and delete currently selected configuration.
- Save / delete operations require 'manage hrreports configuration' and view configuration requires 'access hrreports' permission.
- All actions are processed dynamically with Ajax calls so there is no page refreshing.
- Also I've updated sweetalert JS library to newest version to be able to use prompt popups.

A Reports configuration section preview:
canned_reports


omarabuhussein:
okay two points from the description : 

> All operations require 'access hrreports' permission.
is it okay for any one with  'access hrreports' permission to save a new configuration ? maybe saving a new or overwritting a configuration should be avaliable to any one with a permission like "adminsiorr hrreports" or similar permission if there is a one ? 

> Also I've updated sweetalert JS library to newest version to be able to use prompt popups.

is this library only used by reports or is it used any another place , did you make sure that upgrading it doesn't break something else even inside the reports itself ?

coldrunKacper:
Regarding permissions: I think I need to introduce a new permission as the task description says:
"Admin can decide who can create, edit, remove reports and who has view-only access" so yeah, I'll do this.
Regarding sweetalert JS library, it's used also in js/filters.js script but I've tested it and it's working fine with updated library.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -1567,6 +1567,27 @@ function civihr_employee_portal_menu() {
...
+    $items['reports/%/configuration/%'] = array(
+        'title' => 'Get Report configuration',
+        'page callback' => 'civihr_employee_portal_get_report_configuration_json',
+        'page arguments' => array(1, 3),
+        'access arguments' => array('access hrreports'),
+        'type' => MENU_CALLBACK,
+    );
+    $items['reports/%/configuration/%/save'] = array(
+        'title' => 'Save Report configuration',
+        'page callback' => 'civihr_employee_portal_save_report_configuration_json',
```
omarabuhussein: I am not really a drupal expert , but when I worked with employee portal I was thinking that maybe there is a better way to organize stuff instead of creating all the methods in  extension.module file , so these callbacks can be placed in a separate file or something .. don't know actually I will let it for you to decide

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6581,3 +6605,117 @@ function civihr_employee_portal_convert_array_option_value($optionGroup, $fieldN
...
   return $data;
 }
+
+/**
+ * Get an array containing existing configuration labels for specified Report.
+ *
+ * @param string $reportName
+ * @return array
+ */
+function civihr_employee_portal_get_report_configuration_list($reportName) {
```
omarabuhussein: even this method can probably moved to the new file ( in case that was possible ) as I suggested above .

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6581,3 +6605,117 @@ function civihr_employee_portal_convert_array_option_value($optionGroup, $fieldN
...
+  $responseData = array(
+    'status' => 'success',
+  );
+  try {
+    if (!empty($configId)) {
+      $result = db_update('reports_configuration')
+        ->condition('report_name', $reportName)
+        ->condition('id', (int)$configId)
+        ->fields(array(
+          'json_config' => json_encode($_POST['json_config'])
```
omarabuhussein: I think it is better to check if json_encode is returning a valid output and in that case only you proceed , otherwise db_update will not be called

coldrunKacper:
A new 'manage hrreports configuration' permission implemented (required for saving / deleting configuration).

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6581,3 +6605,117 @@ function civihr_employee_portal_convert_array_option_value($optionGroup, $fieldN
...
+        ))
+        ->execute();
+    } else {
+      $result = db_insert('reports_configuration')
+        ->fields(array(
+          'report_name' => $reportName,
+          'label' => $_POST['label'],
```
omarabuhussein: label should be escaped , or does db_insert use prepared statements by default and not need to worry about escaping you data ?

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6581,3 +6605,117 @@ function civihr_employee_portal_convert_array_option_value($optionGroup, $fieldN
...
+}
+
+/**
+ * Save Report configuration in database.
+ *
+ * @param string $reportName
+ * @param int $configId
+ */
+function civihr_employee_portal_save_report_configuration_json($reportName, $configId) {
+  if (empty($configId) && empty($_POST['label'])) {
```
omarabuhussein: does $reportName and $configId came from user input , if so they need to be escaped too before using them in db calls , unless as i said that drupal use prepared statements by default.

omarabuhussein:
@coldrunKacper I was checking the current permissions and just found that there is already ( Manage HR Reports settings ) maybe you can just use it ?

omarabuhussein:
or no I guess a new permission what is needed here ...

coldrunKacper:
Yes, the 'Manage HR Reports settings' permission is for something slightly different (managing Age Groups).

omarabuhussein:
can you just make sure to update (CiviHR Default Permissions) feature so that users with ( civihr_admin) role have this new permission by default.

civihr_employee_portal/civihr_employee_portal.module
coldrunKacper: Fortunately, these db_insert / db_update and db_select methods are safe (as long as Drupal is :D ) as explained here: http://drupal.stackexchange.com/questions/26816/is-db-insert-secure .

omarabuhussein: it is really good to know that drupal methods are just a wrapper for PDO and they use prepared statements by default , thanks for clarifying that :) .

coldrunKacper:
I've included 'manage hrreports configuration' into civihr_default_permissions feature.

coldrunKacper:
Regarding SSP module I think it needs some refactoring (as you can see), moving some code outside, maybe split into few smaller parts. There is a few parts which can be separated from main module file (as absences, my details, reports and probably more) and the module file is getting bigger and bigger and more difficult to manage. But I would probably require a whole separate task to do it in one go. Let's mention it on CiviHR room and as it needs to be done sooner or later.

PR #210, opened 2016-09-29
PCHR-1520: Add new view to show (HR resource type) Vocabulary terms

omarabuhussein:
## Problem 

Even when Edit terms in HR Resource type permission is enabled the user cannot access http://civihr-staging.civihrhosting.co.uk/admin/structure/taxonomy/hr_resource_type/ .

But that's the default drupal behavior , having the permission will allow the user to edit the "terms" added to (HR Resource type) only , if you want to be able to access this page , then you should have (Administer vocabularies and terms) permission.

But to work around this we need to create a new drupal view that show the list of Vocabulary terms in (HR resource type) .

## Solution

The new view get added (hr_resource_types_list) and another PR added https://github.com/compucorp/civihr-employee-portal-theme/pull/104 to add a link for this new view in the gear menu.

selection_032

### DIFF
--- a/civihr_employee_portal/views/views_export/views_hr_resource_types_list.inc
+++ b/civihr_employee_portal/views/views_export/views_hr_resource_types_list.inc
@@ -1,0 +1,81 @@
+<?php
+
+/**
+ * Exported hr_resource_types_list view
+ */
+
+$view = new view();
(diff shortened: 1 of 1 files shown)


nishant-bhorodia:
Approved!

PR #211, opened 2016-09-29
PCHR-1589: Don't redirct guest users to welcome-page from /user page

omarabuhussein:
When a guest user access /user or ?q=user from the browser , it shouldn't be redirected to /welcome-page because that will prevent from logging into drupal in maintenance mode .

Before : 

before

After :

after


coldrunKacper:
What about /user/register and /user/password (which are available from /user page but still redirect to /welcome-page)? Should they be available too?

omarabuhussein:
> What about /user/register and /user/password (which are available from /user page but still redirect to /welcome-page)? Should they be available too?

good point , I added user/password to allowed paths , but for /user/register it is not needed actually , we don't want users to register new accounts during maintenance mode

coldrunKacper:
Approved.

PR #214, opened 2016-10-04
PCHR-1460: Show (no result) if vacancy search shows no result

omarabuhussein:
## Requirements 
When using filters to search for vacancy , a "no result found" message should appear to the user in case there are no results .

## Solution 

A No results behavior is added to vacancy list drupal view .

- Before

selection_036

- After

selection_035


nishant-bhorodia:
Approved !

PR #215, opened 2016-10-04
PCHR-1613: Update Favicon

deb1990:
# Requirement
Change drupal default favicon to civihr, base it on civihr logo, most likely without text.
1. Create favicon for different resolutions (mobile, tablet, desktop), don't forget retina version. 
2. Make sure it works in drupal (dashboard) as well as in civicrm parts

# Screenshot

**CiviCRM**
image

**Other Drupal Pages**
image


omarabuhussein:
looks fine to me but cannot really tell if there is a better way or not

hitesh-jain:
@deb1990 I think above method is not the right way to change favicon if we are using custom theme, I suppose for Drupal we are using employee portal theme so to update favicon on that just replace/add an favicon without any custom hook functions being implemented that will do work for you and also it will follow Drupal standard of updating favicon.

And for Civicrm it seems we are using seven theme which is a part of Drupal core so changing favicon image won't be possible for this, ideally we should try to always use sub-theme for any of the Drupal Core themes it makes life easier if we need to change anything but it seems that won't be possible here. So the above approach in custom module should be used only for seven theme adding an condition to check.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -7196,14 +7196,15 @@ function civihr_employee_portal_delete_report_configuration_json($reportName, $c
...
 }
 
-function civihr_employee_portal_page_alter(&$pages) {
-  $favicon = drupal_get_path('module', 'civihr_employee_portal') . "/images/fav.png";
-  $type = theme_get_setting('favicon_mimetype');
-  drupal_add_html_head_link(array('rel' => 'shortcut icon', 'href' => drupal_strip_dangerous_protocols($favicon), 'type' => $type));
-}
-
+/**
+ * Changes the favicon for both the SSP and CiviHR admin
```
hitesh-jain: Change to 
/**
 * Implements hook_html_head_alter().
 *  Changes the favicon for both the SSP and CiviHR admin.
 */

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -7195,3 +7195,16 @@ function civihr_employee_portal_delete_report_configuration_json($reportName, $c
   print drupal_json_output(array('status' => 'error'));
   drupal_exit();
 }
+
+/**
+ * Changes the favicon for both the SSP and CiviHR admin
```
hitesh-jain: Change to :
/**
  Implements hook_html_head_alter().
  Changes the favicon for both the SSP and CiviHR admin.

hitesh-jain:
Approved.

PR #217, opened 2016-10-05
PCHR-1652: Set default roles for (access manager approval scree) permission

omarabuhussein:
Updated (civihr default permissions drupal  feature) to include Access CiviHR manager approval screen for administrator, civihr_admin and civihr_manager by default .


coldrunKacper:
Approved.

PR #218, opened 2016-10-05
PCHR-1238: Add vacancy apply page to list of allowed paths

omarabuhussein:
'civicrm/vacancy/apply' Added to the list of allowed path in case the site is online . Related PR ( https://github.com/civicrm/civihr/pull/1164)

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -7252,6 +7252,12 @@ function _user_redirection() {
     'request_new_account/ajax',
   ];
 
+  // If we are not in maintenance mode then the guest users should
+  // be able to access vacancy apply public page.
+  if (!variable_get('maintenance_mode', 0)) {
+    $allowed_paths[] = 'civicrm/vacancy/apply';
+  }
+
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -7023,6 +7023,12 @@ function _block_anonymous_access() {
     'user/login',
   ];
 
+  // If we are not in maintenance mode then the guest users should
+  // be able to access vacancy apply public page.
+  if (variable_get('site_offline', 1)) {
+    $allowed_paths = array_merge(['civicrm/vacancy/apply'], $allowed_paths);
```
davialexandre: actually, is this condition correct? Since they should be allowed to access this only if not in maintenance mode, shouldn't it be:
```php
if(variable_get('site_offline', 0)) {
}
```

Sorry if I'm wrong, but I don't know what is the value stored in site_offline for when it's in maintenance mode. I'm assuming it 1 in that case

davialexandre:
Approved

davialexandre:
Closing stale PR

PR #220, opened 2016-10-07
PCHR-1078: Login destination rework

omarabuhussein:
## Problem 

Currently, when you make a new Drupal role (e.g. civihr_admin2), we have to set up login destination and front page for it manually. (See this https://compucorp.atlassian.net/wiki/display/PCHR/Configuring+a+newly+created+user+role)
Rather, this should be tied to CiviCRM access. 
If a role has "CiviHR: access CiviHR" permission enabled, then they should get this behavior: 

1)CiviHR logo and "Home" button should link to /dashboard
2)After logging in, page goes to T&A dashboard
3)On clicking CiviCRM home, page goes to T&A dashboard

## Solution 

Both (Front Page) and ( Loging destination ) module get removed with any refernces to them  , and this get replaced by implemnting hook_user_login drupal hook , so in case the user has ( access civicrm ) it will get redirected to T&A dashboard , otherwise it will get redirected to SSP dashboard .

-----------------------------------------------------
Related Work : https://github.com/civicrm/civicrm-buildkit/pull/278


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5799,6 +5799,16 @@ function civihr_employee_portal_user_login_submit(&$form, &$form_state) {
     }
 }
 
+/**
+ * Implements hook_user_login
+ */
+function civihr_employee_portal_user_login(&$edit, $account) {
+  global $base_url;
+  if (!isset($_POST['form_id']) || $_POST['form_id'] != 'user_pass_reset') {
+    $_GET['destination'] = user_access('access CiviCRM') ?  'civicrm/tasksassignments/dashboard#/tasks' : 'dashboard';
```
davialexandre: Is this really the best way to do this? Usually, it's a bad practice to change the values from global variables such as $_GET, $_POST, $_SERVER etc.

Have you checked the [user_login_destination](https://api.drupal.org/api/drupal/modules!user!user.module/function/user_login_destination/7.x) function? It looks like it does what you're trying to achieve here.

omarabuhussein: also this is how a popular drupal module called ( login desitination ) which we where using do it http://cgit.drupalcode.org/login_destination/tree/login_destination.module#n345

PR #221, opened 2016-10-18
PCHR-1620: Align columns in absence block

kkalamarski:
## Problem

Columns were not aligned correctly i.e. header table cells were wider than content table cells.
### Before

before-align
### After

dashboard-align
report-align


deb1990:
Approved

PR #223, opened 2016-10-25
PCHR-1589: fix user redirection

omarabuhussein:
After new user requests a login,  the reset password link received via email which looks something like this  https://compucorp.civihrhosting.co.uk/user/reset/12/1477310738/At0L5hbbyhOk-XNeXj8kQhxkGoPKMQjzayjsSBaoXwQ  Will redirect to welcome-page instead which is wrong . So any access to a link start with **user** for anymones users should not lead to welcome-page redirection .

Also  when the logged in user access /civicrm it should be redirected to T&A dashboard not to civicrm dashboard .

So I've changed  **_block_anonymous_access**  to **_user_redirection** to handle both situation

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -7026,14 +7026,12 @@ function _block_anonymous_access() {
     // This is the URL for the modal that is displayed when the user
     // clicks on "Click here to request one from your HR administrator"
     'request_new_account/ajax',
-    // PCHR-1589 : The guest user should be able to access these urls
-    // to access default drupal login and rest password page specially for maintenance mode
-    'user',
-    'user/password',
-    'user/login',
   ];
 
-  if ($user_is_anonymous && !in_array($current_path, $allowed_paths)) {
+  if (
+    $user_is_anonymous &&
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -7026,15 +7025,18 @@ function _block_anonymous_access() {
...
-    // to access default drupal login and rest password page specially for maintenance mode
-    'user',
-    'user/password',
-    'user/login',
   ];
 
-  if ($user_is_anonymous && !in_array($current_path, $allowed_paths)) {
-    drupal_add_http_header('Location', "$base_url/welcome-page");
+  if ($user_is_anonymous) {
+    if (!in_array($current_path, $allowed_paths) && substr($current_path, 0, 4) !== 'user') {
```
coldrunKacper: Notice that user can get to "user" page also with "/index.php?q=user" URL. This rule also applies to any url (/index.php?q=dashboard for example). However I'm not sure if we're going to consider this in this case.

coldrunKacper:
Approved regardless of my comment regarding URL schema (as I guess it's more general thing).

coldrunKacper:
Approved.

PR #224, opened 2016-10-26
PCHR-1646 - Creating view handler for absence_list table content.

coldrunKacper:
The SSP absence_list view was built by three separate views (one for Entitlement header, one for table content with absences and one for footer containing balance values). This was causing a various issues regarding view styling (as each of the three views might have different width) and required (re)define absence types columns each time absence types was changed or added.
I've created a new view handler which generate whole content of absence list in one class which should resolve the issues mentioned above.


civihr_employee_portal/views/includes/civihr_employee_portal_handler_absence_table_header_block_area.inc
```diff
@@ -0,0 +1,212 @@
...
+
+/**
+ * This class contains View Handler for overriding absence_list view.
+ * It's used to generate Absence List custom table structure based
+ * on Drupal View's fields settings.
+ */
+class civihr_employee_portal_handler_absence_table_header_block_area extends views_handler_area {
+
+  /**
+   * Render the area.
```
omarabuhussein: can you please define parameter and return value in the doc block .. or is this an auto-generated file ?

civihr_employee_portal/views/includes/civihr_employee_portal_handler_absence_table_header_block_area.inc
```diff
@@ -0,0 +1,212 @@
...
+    return $row;
+  }
+
+  /**
+   * Return an array containing entitlements row for given $entitlements array.
+   *
+   * @param array $entitlements
+   * @return array
+   */
+  private function getEntitlementRow(array $entitlements) {
```
omarabuhussein: I think this is more formatting rather than actually getting the entitlements .. so maybe formatEntitlementRow is a better name ?

civihr_employee_portal/views/includes/civihr_employee_portal_handler_absence_table_header_block_area.inc
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    return $row;
+  }
+
+  /**
+   * Return an array containing entitlements row for given $entitlements array.
+   *
+   * @param array $entitlements
+   * @return array
```
omarabuhussein: there should be a space between @param and @return , this apply to all methods

civihr_employee_portal/views/includes/civihr_employee_portal_handler_absence_table_header_block_area.inc
```diff
@@ -0,0 +1,212 @@
...
+  private function getEntitlementData($periodId) {
+
+    $entitlementData = array();
+    $query = db_select('civicrm_hrabsence_entitlement', 'che')
+      ->fields('che', array('type_id', 'amount'))
+      ->orderBy('che.contact_id');
+
+    $query->where('(che.contact_id = :c_id)', array(':c_id' => $_SESSION['CiviCRM']['userID']));
+    $query->where('(che.period_id = :period_id)', array(':period_id' => $periodId));
+    $query->where('(che.type_id <> 1)');
```
omarabuhussein: is this condition for sick leave types .. I think it would be better to do a join with leave types table and put the condition on the name instead

civihr_employee_portal/views/views_export/views_absence_list.inc
```diff
@@ -441,32 +292,18 @@ $handler->display->display_options['filters']['absence_start_date_period_filter'
...
 $handler->display->display_options['filters']['absence_start_date_period_filter']['group_info']['group_items'] = array(
   1 => array(
     'title' => 'Generated from code (Do not overwrite)',
     'operator' => 'between',
     'value' => array(
       'type' => 'date',
       'value' => '',
-      'min' => $current_year . '-01-01 00:00:00',
-      'max' => $current_year . '-12-31 23:59:59',
+      'min' => '2016-01-01 00:00:00',
```
omarabuhussein: I know this an auto generated .. but why min and max values are hardcoded here ?

civihr_employee_portal/views/views_export/views_absence_list.inc
```diff
@@ -562,31 +399,18 @@ $handler->display->display_options['filters']['absence_start_date_period_filter'
...
```
omarabuhussein: same here

civihr_employee_portal/civihr_employee_portal.info
```diff
@@ -45,6 +45,7 @@ files[] = views/includes/civihr_employee_portal_argument_age_group.inc
 files[] = views/includes/civihr_employee_portal_handler_absence_duration.inc
 files[] = views/includes/civihr_employee_portal_handler_absence_entitlement.inc
 files[] = views/includes/civihr_employee_portal_handler_activity_type.inc
+files[] = views/includes/civihr_employee_portal_handler_absence_table_header_block_area.inc
```
omarabuhussein: maybe I Am missing some stuff here but shouldn't you deleted any reference to the views you replaced in this PR as well any related files.

civihr_employee_portal/views/includes/civihr_employee_portal_handler_absence_table_header_block_area.inc
coldrunKacper: Yeah, I've changed the method to use API anyway (like the other methods) and added a helper function to SSP module which returns 'sick' type ID.

civihr_employee_portal/views/includes/civihr_employee_portal_handler_absence_table_header_block_area.inc
coldrunKacper: It's more like "get a row with Entitlement data" (so it's not about getting entitlement itself).

civihr_employee_portal/views/views_export/views_absence_list.inc
coldrunKacper: It's because for defining date period it requires to specify dates range (min and max). The same is when you define a new period in CiviCRM: /index.php?q=civicrm/absence/period&action=add&reset=1

PR #226, opened 2016-10-26
PCHR-1636: Allow civihr_admin role to access the site in maintenance mode

omarabuhussein:
any user with civihr_admin should be able to access the site in maintenance mode


coldrunKacper:
Approved.

PR #229, opened 2016-10-28
PCHR-1651 - Absence report doesn't pick up absence data from civi

coldrunKacper:
Rewriting absence_activity SQL View to base on leave activity records by leave type IDs  instead of custom absence comment table. The comment table may be empty and therefore SQL View won't create any data.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -492,10 +492,10 @@ function _rebuild_absence_activity_view() {
  * @return array
  */
 function _get_absence_activity_types_ids() {
-  $activityTypesIds = array();
-  $absenceTypes = civicrm_api3('HRAbsenceType', 'get', array(
+  $activityTypesIds = [];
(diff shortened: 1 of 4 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -446,38 +446,67 @@ function _rebuild_absence_activity_view() {
...
+  }
+}
+
+/**
+ * Return an array of absence activity types IDs.
+ *
+ * @return array
+ */
+function _get_absence_activity_types_ids() {
+  $activityTypesIds = array();
```
omarabuhussein: can you use the short array syntax here ..

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -446,38 +446,67 @@ function _rebuild_absence_activity_view() {
...
+  $absenceTypes = civicrm_api3('HRAbsenceType', 'get', array(
```
omarabuhussein: here too

PR #231, opened 2016-10-28
PCHR-1628 - Absence report date filter is showing no results

coldrunKacper:
Creating a new 'civihr_employee_portal_hrreport_get_exposed_filter_default_values()' SSP function to map date filter default values to the view if the values are not specified in the request.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -6295,7 +6295,7 @@ function civihr_employee_portal_hrreport_get_json($reportName) {
 
 /**
  * Return an array with mapped default exposed filter values for given view
- * and params. Used to map date filter default values if they are not specified
+ * and parameters. Used to map date filter default values if they are not specified
  * in request.
  *
  * @param object $view
@@ -6311,13 +6311,17 @@ function civihr_employee_portal_hrreport_get_exposed_filter_default_values($view
   $dateFilter = $view->display['default']->display_options['filters']['date_filter'];
   if (!empty($dateFilter['expose']['identifier'])) {
     $identifier = $dateFilter['expose']['identifier'];
+
     $defaultFromDate = !empty($dateFilter['default_date']) ? $dateFilter['default_date'] : '';
     $defaultToDate = !empty($dateFilter['default_to_date']) ? $dateFilter['default_to_date'] : '';
+
     $from = (new DateTime())->modify($defaultFromDate);
     $to = (new DateTime())->modify($defaultToDate);
+
     if (empty($params[$identifier]['min'])) {
       $params[$identifier]['min'] = $from->format('Y-m-d');
     }
+
     if (empty($params[$identifier]['max'])) {
       $params[$identifier]['max'] = $to->format('Y-m-d');
     }
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -6312,29 +6312,37 @@ function civihr_employee_portal_hrreport_get_exposed_filter_default_values($view
   if (!empty($dateFilter['expose']['identifier'])) {
     $identifier = $dateFilter['expose']['identifier'];
 
-    $defaultFromDate = !empty($dateFilter['default_date']) ? $dateFilter['default_date'] : null;
-    $defaultToDate = !empty($dateFilter['default_to_date']) ? $dateFilter['default_to_date'] : null;
-
-    $from = new DateTime();
-    if ($defaultFromDate) {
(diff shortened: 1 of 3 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6276,14 +6276,56 @@ function civihr_employee_portal_hrreport_custom($reportName) {
...
+  $view->exposed_input = civihr_employee_portal_hrreport_get_exposed_filter_default_values($view, $params);
   $view->set_display('page');
   $view->execute();
   $view->style_plugin->options['using_views_api_mode'] = FALSE;
   return $view->render('page');
 }
 
+/**
+ * Return an array with mapped default exposed filter values for given view
+ * and params. Used to map date filter default values if they are not specified
```
omarabuhussein: change params =(to)> parameters  ..  since this is a method description not a variable name

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6276,14 +6276,56 @@ function civihr_employee_portal_hrreport_custom($reportName) {
...
+ *
+ * @return array
+ */
+function civihr_employee_portal_hrreport_get_exposed_filter_default_values($view, array $params) {
+  if (empty($view->display['default']->display_options['filters']['date_filter'])) {
+    return $params;
+  }
+
+  $dateFilter = $view->display['default']->display_options['filters']['date_filter'];
+  if (!empty($dateFilter['expose']['identifier'])) {
```
omarabuhussein: It would be better to add some spaces :

```php
  if (!empty($dateFilter['expose']['identifier'])) {
    $identifier = $dateFilter['expose']['identifier'];

    $defaultFromDate = !empty($dateFilter['default_date']) ? $dateFilter['default_date'] : '';
    $defaultToDate = !empty($dateFilter['default_to_date']) ? $dateFilter['default_to_date'] : '';

    $from = (new DateTime())->modify($defaultFromDate);
    $to = (new DateTime())->modify($defaultToDate);

    if (empty($params[$identifier]['min'])) {
      $params[$identifier]['min'] = $from->format('Y-m-d');
    }

    if (empty($params[$identifier]['max'])) {
      $params[$identifier]['max'] = $to->format('Y-m-d');
    }
  }
```

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6276,14 +6276,56 @@ function civihr_employee_portal_hrreport_custom($reportName) {
...
+    $identifier = $dateFilter['expose']['identifier'];
+    $defaultFromDate = !empty($dateFilter['default_date']) ? $dateFilter['default_date'] : '';
+    $defaultToDate = !empty($dateFilter['default_to_date']) ? $dateFilter['default_to_date'] : '';
+    $from = (new DateTime())->modify($defaultFromDate);
```
omarabuhussein: so if $defaultFromDate was an empty string then this will return false (I think ) .. and then  $from->format('Y-m-d') below will fail which will cause a corrupted value in $params[$identifier]['min'] .. while I am not sure what exactly the required here but you may need to add some checking here to prevent that.

coldrunKacper: You're right! I've added additional IFs for this.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6312,11 +6312,18 @@ function civihr_employee_portal_hrreport_get_exposed_filter_default_values($view
   if (!empty($dateFilter['expose']['identifier'])) {
     $identifier = $dateFilter['expose']['identifier'];
 
-    $defaultFromDate = !empty($dateFilter['default_date']) ? $dateFilter['default_date'] : '';
-    $defaultToDate = !empty($dateFilter['default_to_date']) ? $dateFilter['default_to_date'] : '';
+    $defaultFromDate = !empty($dateFilter['default_date']) ? $dateFilter['default_date'] : null;
```
omarabuhussein: sorry I just noticed this .. but I guess it would make more sense to orgnize the code like this :

```php
    if (empty($params[$identifier]['min'])) {
      $defaultFromDate = !empty($dateFilter['default_date']) ? $dateFilter['default_date'] : null;

      if ($defaultFromDate) {
        $from = new DateTime();
        $from->modify($defaultFromDate);
        $params[$identifier]['min'] = $from->format('Y-m-d');
      }
    }
```

the same applies for $params[$identifier]['max']

omarabuhussein: and to avoid duplicated you could separate this into a new function and call it twice for min and max values

PR #232, opened 2016-10-31
PCHR-1248: Refractored update hooks naming convention.

hitesh-jain:
Problem : 
Updates hook were given in-proper naming convention under civiHR Employee Portal features module, calling civihr_employee_portal_update_7000 but it should be called as civihr_employee_portal_features_update_7000 according to https://api.drupal.org/api/drupal/modules%21system%21system.api.php/function/hook_update_N/7.x
Above was causing issues as we already have an module named civihr_employee_portal  so while updating it was throwing 

`Fatal error: Cannot redeclare civihr_employee_portal_update_7001() (previously declared in sites/all/modules/civihr-custom/civihr_employee_portal/civihr_employee_portal.install:32) in sites/all/modules/civihr-custom/civihr_employee_portal/features/civihr_employee_portal_features/civihr_employee_portal_features.install on line 141
`

Solution:
Rewritten functions according to naming conventions and added menu checks function so that it doesn't recreate menus while updating again.

### DIFF
--- a/civihr_employee_portal/features/civihr_employee_portal_features/civihr_employee_portal_features.install
+++ b/civihr_employee_portal/features/civihr_employee_portal_features/civihr_employee_portal_features.install
@@ -233,7 +233,7 @@ function civihr_employee_portal_features_uninstall() {
 function _is_menu_link_path_exists($path = '') {
 
   if ($path) {
-    // Lookup fpr link path table if the path does exist.
+    // Lookup for link path table if the path does exist.
     $menu_status = db_select('menu_links', 'ml')
       ->fields('ml', ['mlid'])
       ->condition('ml.link_path', $path)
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/features/civihr_employee_portal_features/civihr_employee_portal_features.install
```diff
@@ -212,4 +219,27 @@ function civihr_employee_portal_features_uninstall() {
...
+ * @param string
+ *      Path of the menu link.
+ *
+ * @return bool
+ *      Status of the menu link path.
+ */
+function _is_menu_link_path_exists($path = '') {
+
+  if ($path) {
+    // Lookup fpr link path table if the path does exist.
```
coldrunKacper: A typo?

coldrunKacper:
Approved.

PR #233, opened 2016-11-02
PCHR-1247: Renamed module name and changed menu ordering.

hitesh-jain:
Renamed feature name to Leave and Absences Features and changed the ordering of My Leave menu item.


coldrunKacper:
Approved.

PR #238, opened 2016-11-17
PCHR-1382: Fixed Quick Approve/Reject Buttons in Manager Absence Approval

MiyaNoctem:
Changed functionality of quick approve/reject buttons, so that instead of making a
direct API Call to CiviCRM, it makes the call to the Drupal civihr_employee_portal
module, where the request is processed.

- Changed quick-button functionality on front-end
- Added manager_approval/%ctools_js/quick_approval route
- Route uses civihr_employee_portal_manager_approval_quick_approval as callback
- Added _invoke_rules_on_absence_decision helper function to build required data
  for rules_invoke_event

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -1625,6 +1625,13 @@ function civihr_employee_portal_menu() {
         'type' => MENU_CALLBACK,
     );
 
+    $items['manager_approval/%ctools_js/quick_approval'] = array(
+      'page callback' => 'civihr_employee_portal_manager_approval_quick_approval',
+      'page arguments' => array(1, 2, 3, 4, 5),
+      'access callback' => TRUE,
+      'type' => MENU_CALLBACK,
+    );
+
     $items['manager_approval/%ctools_js/calendar/show'] = array(
         'page callback' => 'civihr_employee_portal_manager_approval_show_calendar_callback',
         'page arguments' => array(1),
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/js/filters.js
```diff
@@ -25,63 +25,33 @@ Drupal.behaviors.civihr_employee_portal_filters = {
...
-                        // Rebuild filters
-                        loadFilters();
+                            // Remove the row (so it will not be loaded when rebuilding the filters)
+                            $("#act-id-" + clicked_object).closest('tr').remove();
 
-                        console.log(Drupal.settings.basePath + 'ajax/quick_email_notify/' + clicked_object + '/' + type);
+                            // Rebuild filters
+                            loadFilters();
 
-                        // Notify by email
```
omarabuhussein: this is important to send a notification via the email you should not delete it .. I Think it will be even better to approve and send the email in a single ajax call instead of two .. so you maybe can keep it removed and inside **civihr_employee_portal_manager_approval_quick_approval** drupal function you can call this function this function **_ajax_quick_email_notify**  to send the email

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -1625,6 +1625,13 @@ function civihr_employee_portal_menu() {
         'type' => MENU_CALLBACK,
     );
 
+    $items['manager_approval/%ctools_js/quick_approval'] = array(
```
omarabuhussein: maybe I Am missing something here but rout here is 

``php
manager_approval/%ctools_js/quick_approval
```

but in the js file it is :

```php 
manager_approval/ajax/quick_approval/
```

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -2170,6 +2177,117 @@ function civihr_employee_portal_manager_approval_pick_days_callback($ajax, $acti
 }
 
 /**
+ * Function that processes approvals and rejections made by manager from the absence approval widget
+ *
+ * @param int $ajax
+ *   boolean value, 1 if it is an ajax call
+ * @param string $action
+ *   should always be "quick_approval"
```
omarabuhussein: if this always the same value then what is need for it

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -2170,6 +2177,117 @@ function civihr_employee_portal_manager_approval_pick_days_callback($ajax, $acti
...
+}
+
+/**
+ * Helper function that builds required parameters to pass to rules_invoke_event on decision (accept/reject) by manager
+ *
+ * @param $main_activity
+ * @param $type
+ * @param $absence_details
+ */
+function _invoke_rules_on_absence_decision($main_activity, $type, $absence_details) {
```
omarabuhussein: what is this about .. I think the ticket was about activity permissions ? can you explain a bit

civihr_employee_portal/js/filters.js
MiyaNoctem: Well, I implemented the same process flow used on approvals/rejection used on the calendar form.  If I understand correctly, invoking the "approve_all_post_event" and "reject_all_post_event" rule at the end would take care of that e-mail, wouldn't it?  If I call _ajax_quick_email_notify, wouldn't two emails be sent?

civihr_employee_portal/civihr_employee_portal.module
MiyaNoctem: The purpose of this function is to call "approve_all_post_event" and "reject_all_post_event" rules respectively, so the e-mail is sent to all interested parties on each event.  However, I found building required parameters for the rule took a lot of code, so I thought it would be cleaner to refactor all that code into this function.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -2170,6 +2177,121 @@ function civihr_employee_portal_manager_approval_pick_days_callback($ajax, $acti
...
+function civihr_employee_portal_manager_approval_quick_approval($ajax, $action, $activity_id, $status, $type) {
+  // Civi init
+  civicrm_initialize();
+
+  try {
+    // Load main activity leave
+    $main_activity = civicrm_api3('Activity', 'get', ['sequential' => 1, 'id' => $activity_id])['values'][0];
+
+    // Get activities for each day of leave
+    $activityTypes = CRM_Core_OptionGroup::values('activity_type', FALSE, FALSE, FALSE, NULL, 'name');
```
omarabuhussein: you can just to the OptionValues API to get the activity type where the name = 'Absence' instead of fetching all activity types and then filtering the one you want :

```php
$absenceTypeValue= civicrm_api3('OptionValue', 'get', array(
  'sequential' => 1,
  'return' => array("value"),
  'option_group_id' => "activity_type",
  'name' => "Absence",
  'options' => array('limit' => 1),
))['values'][0];

    $search_params = [
      'source_record_id' => $activity_id,
      'activity_type_id' => CRM_Utils_Array::key('Absence', $absenceTypeValue)
    ];
```

omarabuhussein:
Also I just paid an attention for the PR title , it should briefly describe the issue ( take a hint from the ticket title ) ... also can you use the interactive rebase to change the commit message for this commit https://github.com/compucorp/civihr-employee-portal/pull/238/commits/d56db31102223a0f43f17beb05218720b9e5fb7e since also (Fixed Issue) for a commit message is not very descriptive

PR #240, opened 2016-11-23
PCHR-1609: Fixed Problems on Staff Directory View

MiyaNoctem:
Several issues were reported with Staff Directory view, all problems where
due to misconfiguration of parameters to generate the query for the contact
list.

1. Not all contacts shown: view was filtering contacts that either had no manager, or had a manager with the "Leave Aproved By" relationship exclusively. Solved by removing conditions from filters.
2. Managers not listed at all: relationship to contacts table for manager was not configured.  Fixed by adding relationship and associating field to relationship.
3. Manager filter not working: as with problem number 2, relationship was not configured, hence filter could not be used.
4. Ajax error when searching name: problem reported to occur when a nameless contact was present on result set.  Unable to replicate, but tested and works ok.
5. Contacts in trash ar shown: added filter to show only contacts that are not deleted.

*Before*
pchr-1609-before

*After*
pchr-1609-after

### DIFF
--- a/civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
+++ b/civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
@@ -39,9 +39,9 @@ $handler->display->display_options['style_plugin'] = 'table';
 $handler->display->display_options['style_options']['columns'] = array(
   'id' => 'id',
   'image_URL' => 'image_URL',
-  'display_name' => 'display_name',
   'first_name' => 'first_name',
   'last_name' => 'last_name',
+  'display_name' => 'display_name',
   'phone' => 'phone',
   'phone_ext' => 'phone_ext',
   'title' => 'title',
@@ -49,14 +49,14 @@ $handler->display->display_options['style_options']['columns'] = array(
   'location' => 'location',
   'contract_type' => 'contract_type',
   'department' => 'department',
-  'display_name_1' => 'display_name_1',
   'end_date' => 'end_date',
   'is_active' => 'is_active',
   'start_date' => 'start_date',
   'end_date_1' => 'end_date_1',
   'start_date_1' => 'start_date_1',
-  'first_name_1' => 'first_name_1',
-  'last_name_1' => 'last_name_1',
+  'display_name_1' => 'display_name_1',
+  'is_primary' => 'is_primary',
+  'email_1' => 'email_1',
 );
 $handler->display->display_options['style_options']['default'] = '-1';
 $handler->display->display_options['style_options']['info'] = array(
@@ -74,13 +74,6 @@ $handler->display->display_options['style_options']['info'] = array(
     'separator' => '',
     'empty_column' => 0,
   ),
-  'display_name' => array(
-    'sortable' => 1,
-    'default_sort_order' => 'asc',
-    'align' => '',
-    'separator' => '',
-    'empty_column' => 0,
-  ),
   'first_name' => array(
     'sortable' => 0,
     'default_sort_order' => 'asc',
@@ -89,6 +82,13 @@ $handler->display->display_options['style_options']['info'] = array(
     'empty_column' => 0,
   ),
   'last_name' => array(
+    'sortable' => 1,
+    'default_sort_order' => 'asc',
+    'align' => '',
+    'separator' => '',
+    'empty_column' => 0,
+  ),
+  'display_name' => array(
     'sortable' => 1,
     'default_sort_order' => 'asc',
     'align' => '',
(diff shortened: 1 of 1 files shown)


omarabuhussein:
okay first thing and since this is affect what is present in the UI .. can you post before and after screenshot in the PR description .

civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
```diff
@@ -39,24 +39,24 @@ $handler->display->display_options['style_plugin'] = 'table';
...
-  'display_name_1' => 'display_name_1',
   'end_date' => 'end_date',
   'is_active' => 'is_active',
   'start_date' => 'start_date',
   'end_date_1' => 'end_date_1',
   'start_date_1' => 'start_date_1',
-  'first_name_1' => 'first_name_1',
-  'last_name_1' => 'last_name_1',
+  'display_name_1' => 'display_name_1',
+  'is_primary' => 'is_primary',
```
omarabuhussein: I don't think is_primary field is necessary to be shown to the users

civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
```diff
@@ -290,13 +300,13 @@ $handler->display->display_options['fields']['title']['field'] = 'title';
...
 $handler->display->display_options['fields']['title']['label'] = 'Job Title';
 /* Field: CiviCRM Email: Email Address */
-$handler->display->display_options['fields']['email']['id'] = 'email';
-$handler->display->display_options['fields']['email']['table'] = 'civicrm_email';
-$handler->display->display_options['fields']['email']['field'] = 'email';
-$handler->display->display_options['fields']['email']['label'] = 'Email';
-$handler->display->display_options['fields']['email']['location_type'] = '2';
-$handler->display->display_options['fields']['email']['location_op'] = '0';
-$handler->display->display_options['fields']['email']['is_primary'] = 0;
+$handler->display->display_options['fields']['email_1']['id'] = 'email_1';
```
omarabuhussein: are you sure this should be email_1 instead of email ?

civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
```diff
@@ -334,50 +344,29 @@ $handler->display->display_options['fields']['start_date']['field'] = 'start_dat
...
 $handler->display->display_options['fields']['end_date_1']['relationship'] = 'role_jobcontract_id';
 $handler->display->display_options['fields']['end_date_1']['exclude'] = TRUE;
-/* Field: HRJobContract Role entity: Start_date */
+/* Field: Broken/missing handler */
 $handler->display->display_options['fields']['start_date_1']['id'] = 'start_date_1';
 $handler->display->display_options['fields']['start_date_1']['table'] = 'hrjc_role';
 $handler->display->display_options['fields']['start_date_1']['field'] = 'role_star_date';
 $handler->display->display_options['fields']['start_date_1']['relationship'] = 'role_jobcontract_id';
 $handler->display->display_options['fields']['start_date_1']['exclude'] = TRUE;
-/* Field: CiviCRM Contacts: First Name */
```
omarabuhussein: are you sure about removing these fields ?

civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
```diff
@@ -334,50 +344,29 @@ $handler->display->display_options['fields']['start_date']['field'] = 'start_dat
...
-$handler->display->display_options['fields']['first_name_1']['id'] = 'first_name_1';
```
omarabuhussein: are you sure about removing these fields ?

civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
MiyaNoctem: Yeah, it's not shown.  It's just there because it's used in a filter.

civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
MiyaNoctem: No problem, so long as it's consistent throughout the view, as it's the field's name within it.  The "_1" was added because at some point I had both the original field that didn't work and the new field that worked together in the view, to compare their configuration.  So kept the new one (email_1) and deleted the old one (email).

civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
MiyaNoctem: Yeah, those were used to try and show the manager's name, but were not associated with the relationship, so they didn't work.  So I just included display_name (so so first, middle and last names could be used to search) and deleted those.

PR #247, opened 2016-12-12
PCHR-1803: Fix Location Value in Staff Directory View

MiyaNoctem:
Staff directory was currently showing the location value for each contact (as taken from option_value.value), which could be an integer or some other string with no meaningful significance for the user.  

Fix required:

1. Including option_value.label for the contract's location in database view being used to incorporate CiviHR Job Contract Details Entity into drupal.
2. Replacing location field in staff directory view to use location label instead of location value.


MiyaNoctem:
Will add to PCHR-1767 branch.

PR #248, opened 2016-12-13
PCHR-1802: Fix Departments Disappearing From Results on Search

MiyaNoctem:
When a contact in staff directory has roles in different departments, these are shown as comma separted values in staff directory. If you search by a specific department, although the search result shows the correct staff, department column shows only the department you searched for, regardless of how many departments each contact is in.  Also, department column is showing values from all past job roles too.

Fixed by adding a department_list column to hrjc_details view in CMS database, so that a department list for active roles of that contact's contract/revision is always shown, despite filtering of specific roles trough searchable role fields.

Also updated view to show new department_list field in results, and to include only active roles in all queries.


MiyaNoctem:
Will include fixes in PCHR-1767 branch and PR.

PR #249, opened 2016-12-13
PCHR-1767: Fix Staff Directory Problems with Locations and Departments

MiyaNoctem:
**Issues with Departments**

When a contact in staff directory has roles in different departments, these are shown as comma separted values in staff directory. If you search by a specific department, although the search result shows the correct staff, department column shows only the department you searched for, regardless of how many departments each contact is in. Also, department column is showing values from all past job roles too.

Fixed by adding a department_list column to hrjc_details view in CMS database, so that a department list for active roles of that contact's contract/revision is always shown, despite filtering of specific roles trough searchable role fields.

Also updated view to show new department_list field in results, and to include only active roles in all queries.

**Issues with Locations**
Staff directory was currently showing the location value for each contact (as taken from option_value.value), which could be an integer or some other string with no meaningful significance for the user.

Fix required including option_value.label for the contract's location in database view being used to incorporate CiviHR Job Contract Details Entity into drupal, and replacing location field in staff directory view to use location label instead of location value.

**Managers**
Managers had the same problem as departments: when you searched for one, the other managers would disappear from the results of contacts with multiple managers. Fixed by adding a second relationship to contact b, unfiltered by search criteria.

### DIFF
--- a/civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
+++ b/civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
@@ -45,18 +45,17 @@ $handler->display->display_options['style_options']['columns'] = array(
   'phone' => 'phone',
   'phone_ext' => 'phone_ext',
   'title' => 'title',
-  'email' => 'email',
-  'location' => 'location',
+  'email_1' => 'email_1',
+  'location_label' => 'location_label',
   'contract_type' => 'contract_type',
   'department' => 'department',
+  'department_list' => 'department_list',
   'end_date' => 'end_date',
   'is_active' => 'is_active',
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/views/views_export/views_hr_staff_directory.inc
```diff
@@ -184,20 +190,6 @@ $handler->display->display_options['style_options']['info'] = array(
     'separator' => '',
     'empty_column' => 0,
   ),
-  'is_primary' => array(
```
omarabuhussein: I maybe missed this , but why do you removed these lines ?

PR #250, opened 2016-12-19
PCHR-1792: Added Manager Leave block containing AngularJS app.

hitesh-jain:
Problem :
Added Manager Leave block that can fetch custom markup from pre-defined template file and also an angular js app defined for Manager Leave block that will be use to place block in whatever page/view is necessary.

Solution:
Created new block called Manager Leave that contains markup for block from pre-defined template file and angular js app. As this is custom block we can easily export and place in whatever region we require.

### DIFF
--- a/civihr_employee_portal/civihr_leave_absences/civihr_leave_absences.module
+++ b/civihr_employee_portal/civihr_leave_absences/civihr_leave_absences.module
@@ -14,6 +14,10 @@ function civihr_leave_absences_block_info() {
     'info' => t('My Leave'),
     'cache' => DRUPAL_NO_CACHE,
   ];
+  $blocks['manager_leave'] = [
+    'info' => t('Manager Leave'),
+    'cache' => DRUPAL_NO_CACHE,
+  ];
   return $blocks;
 }
 
@@ -27,6 +31,9 @@ function civihr_leave_absences_block_view($delta = '') {
   switch ($delta) {
     case 'my_leave':
       $block['content'] = civihr_leave_absences_my_leave_get_markup();
+      break;
+    case 'manager_leave':
+      $block['content'] = civihr_leave_absences_manager_leave_get_markup();
       break;
   }
   return $block;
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/civihr_leave_absences/civihr_leave_absences.module
```diff
@@ -60,3 +68,23 @@ function civihr_leave_absences_my_leave_get_markup() {
...
+ *    Custom markup from template file.
+ */
+function civihr_leave_absences_manager_leave_get_markup() {
+
+  // @ TODO -> Pointing to Angular JS App once its ready for Manager Leave block.
+  // Loading angular js application for Manager Leave block using CiviCRM Resources API.
+  // eg: civicrm_resources_load('org.civicrm.hrjobcontract', array('gulpfile.js', 'hrjc.css', 'contact.js'));
+
+  // Fetching custom markup from pre-defined template file.
+  $file_path = drupal_get_path('module', 'civihr_leave_absences') . '/templates/manager_leave.html';
```
AkA84: You are starting to have duplicated code. Please move this
```php
$file_path = drupal_get_path('module', 'civihr_leave_absences') . "/templates/#{tplName}.html";

if (file_exists($file_path)) {
   return @file_get_contents($file_path);
 } else {
  return '';
}
```
in a private function so that both `civihr_leave_absences_manager_leave_get_markup` and `civihr_leave_absences_my_leave_get_markup` can call it

(i'm pretty sure that the whole thing could be abstracted even more, but let's start with this at least)

civihr_employee_portal/civihr_leave_absences/templates/manager_leave.html
```diff
@@ -0,0 +1,3 @@
+<section data-leave-absences-manager-leave>
+    <p>Manager Leave Block Template.</p>
```
AkA84: We're using 2 spaces indentation, not 4

civihr_employee_portal/civihr_leave_absences/templates/manager_leave.html
```diff
@@ -0,0 +1,3 @@
...
+</section>
```
AkA84: Please set up your IDE to always leave a new line when a file is saved

PR #252, opened 2016-12-21
PCHR-1561: Hide Appraisal menu

omarabuhussein:
## Problem

The appraisals extension is not comeplete yet , so its SSP menu and page need to be disabled for now

selection_187

## Solution

I disabled the appraisals view page so it will not be shown on the SSP anymore.

selection_186


davialexandre:
This PR is not relevant anymore and its has been closed, so I'm closing it

PR #254, opened 2017-01-04
PCHR-1818: Push the L&A permissions to the global CRM.permissions var

davialexandre:
In order to be able to check permissions on the frontend code, we need to make civi export them to the global CRM.permissions global javascript var. This can be done by passing a list of permissions to the `CRM_Core_Resources::addPermissions()` method, which is called on the hook_init() implementation of the "CiviHR Leave and Absences" Drupal module.

The exported permissions are:
- access leave and absences
- administer leave and absences
- access leave and absences in ssp
- manage leave and absences in ssp


coldrunKacper:
Approved.

PR #259, opened 2017-01-09
PCHR-1712: Improve Administration of Content on Welcome Page

MiyaNoctem:
## Problem
Content in welcome page holding logo, site name and slogan was being included using a custom module with the raw HTML left to be edited to update any information, which is not ideal if it is supposed to be administered by regular users.

## Solution
Removed custom block and added separate blocks with administrable drupal site logo, site name and slogan to the page.  Each block has its own CSS ID, so each can be styled to match expected look & feel for welcome page.


coldrunKacper:
Approved.

PR #264, opened 2017-01-19
PCHR-1764: Get Staff Contact Information by Location Name

MiyaNoctem:
## Problem
E-mail and phone data shown on My Details view and Staff Directory view was being queried from database using ID's for location that were hard-coded on the views.  This meant if the ID's changed for any reason, information for contacts was lost on those views.

### Staff Directory - Before
staff-before

### Dashboard - Before
dashboard-before

### My Details - Before
mydetails-before

## Solution
Fixed by creating a CiviHR entity built on a database view that queries work and home e-mail and phone information using location names instead of ID's.

### Staff Directory - After
staff-after

### Dashboard - After
dashboard-after

### My Details - After
mydetails-after

### DIFF
--- a/civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
+++ b/civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
@@ -1196,4 +1196,3 @@ function civihr_hrjobcontract_entities_views_api() {
     );
 }
 
-
--- a/civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
+++ b/civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
@@ -251,6 +251,43 @@ function civihr_hrjobcontract_entities_init() {
                     WHERE c.deleted = 0
                     ORDER BY t.job_contract_id");
 
+
+        db_query('DROP VIEW IF EXISTS hrjc_contact_details');
+        db_query("CREATE OR REPLACE VIEW hrjc_contact_details AS
+          SELECT c.id AS contact_id,
+            GROUP_CONCAT(DISTINCT CASE WHEN phone_location.name = 'Work' THEN c_phone.phone ELSE '' END
+              ORDER BY c_phone.phone DESC SEPARATOR ' ') AS work_phone,
+            GROUP_CONCAT(DISTINCT CASE WHEN phone_location.name = 'Work' THEN c_phone.phone_ext ELSE '' END 
+              ORDER BY c_phone.phone_ext DESC SEPARATOR ' ') AS work_phone_ext,
+            GROUP_CONCAT(DISTINCT CASE WHEN email_location.name = 'Work' THEN c_email.email ELSE '' END 
+              ORDER BY c_email.email DESC SEPARATOR ' ') AS work_email,
+
+            GROUP_CONCAT(DISTINCT CASE WHEN phone_location.name = 'Home' THEN c_phone.phone ELSE '' END
+              ORDER BY c_phone.phone DESC SEPARATOR ' ') AS home_phone,
+            GROUP_CONCAT(DISTINCT CASE WHEN phone_location.name = 'Home' THEN c_phone.phone_ext ELSE '' END 
+              ORDER BY c_phone.phone_ext DESC SEPARATOR ' ') AS home_phone_ext,
+            GROUP_CONCAT(DISTINCT CASE WHEN email_location.name = 'Home' THEN c_email.email ELSE '' END 
+              ORDER BY c_email.email DESC SEPARATOR ' ') AS home_email
+
+          FROM {$civi_db_name}.civicrm_contact c
+          LEFT JOIN {$civi_db_name}.civicrm_phone c_phone ON c_phone.contact_id = c.id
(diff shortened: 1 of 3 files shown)


civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
```diff
@@ -251,6 +251,43 @@ function civihr_hrjobcontract_entities_init() {
...
+          )
+          LEFT JOIN {$civi_db_name}.civicrm_email c_email ON c_email.contact_id = c.id
+          LEFT JOIN {$civi_db_name}.civicrm_location_type email_location ON (
+            c_email.location_type_id = email_location.id
+            AND (email_location.name = 'Work' OR email_location.name = 'Home')
+          )
+
+          WHERE phone_location.name = 'Work'
+          OR email_location.name = 'Work'
+          OR phone_location.name = 'Home'
```
omarabuhussein: I think we don't want to display home phone or email , but better to check with Mike

civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
```diff
@@ -251,6 +251,43 @@ function civihr_hrjobcontract_entities_init() {
                     WHERE c.deleted = 0
                     ORDER BY t.job_contract_id");
 
+
+        db_query('DROP VIEW IF EXISTS hrjc_contact_details');
+        db_query("CREATE OR REPLACE VIEW hrjc_contact_details AS
+          SELECT c.id AS contact_id,
+            GROUP_CONCAT(DISTINCT CASE WHEN phone_location.name = 'Work' THEN c_phone.phone ELSE '' END
```
omarabuhussein: hmmm not sure if this would work but isn't there is a way in drupal views to display a default value if nothing found for that column .. shouldn't it work instead of using CASE in mysql ( not sure really )

civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
```diff
@@ -251,6 +251,43 @@ function civihr_hrjobcontract_entities_init() {
...
+              ORDER BY c_phone.phone DESC SEPARATOR ' ') AS work_phone,
+            GROUP_CONCAT(DISTINCT CASE WHEN phone_location.name = 'Work' THEN c_phone.phone_ext ELSE '' END 
+              ORDER BY c_phone.phone_ext DESC SEPARATOR ' ') AS work_phone_ext,
+            GROUP_CONCAT(DISTINCT CASE WHEN email_location.name = 'Work' THEN c_email.email ELSE '' END 
+              ORDER BY c_email.email DESC SEPARATOR ' ') AS work_email,
+
+            GROUP_CONCAT(DISTINCT CASE WHEN phone_location.name = 'Home' THEN c_phone.phone ELSE '' END
```
omarabuhussein: I think you can use IN statement above instead of duplicating them for work and home .. though as I said I don't think we need home phone and email at all (to be checked with mike)

civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
MiyaNoctem: The idea of the case expression is to filter out phones that are not in Work location or Home location respectively for each column.  If I take it out, for example, if there is someone who has home phone, but not work phone, the home phone will appear in the column for work phone too.  So the case expression replaces phones not in work location for an empty string on work_phone column.

civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
MiyaNoctem: I could use an IN statement, but if I did, I'd end up having different records for work phone and home phone. Also for work e-mail and home e-mail. Worse case scenario, I'd have 4 different records to contend with.

On the other hand, to use the information in the views I'd have to include the relationship to the entity and then filter by what I need.  But if I filter on the view, all other records will not appear.  For example, if I filter for records with work e-mail, only records with work e-mail would appear, because views makes me filter adding clauses to the WHERE clause. What I need is that records with no work e-mail still appear, but with empty information. This could be done if I could add filters to the JOIN clause, but I found no easy way of doing this.  So I just built the query so that I get one record per contact with all the information I need.

civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
MiyaNoctem: Home phone and e-mail are seen on SSP Dashboard and My Details pages.

civihr_hrjobcontract_entities/civihr_hrjobcontract_entities.module
```diff
@@ -1194,6 +1194,4 @@ function civihr_hrjobcontract_entities_views_api() {
         'api' => 3,
         'path' => drupal_get_path('module', 'civihr_hrjobcontract_entities') . '/views',
     );
-}
-
-
+}
```
omarabuhussein: there is should be only one new line at the end of any file (unless there is a reason not to ) , please fix this .

omarabuhussein:
@MiyaNoctem just added one final note about end of file new lines.

PR #268, opened 2017-02-02
PCHR-1870: Adding destination parameters while user login and redirection.

hitesh-jain:
**Problem**
In civihr after login you always get redirected to T&A dashboard (if you are admin) or SSP (if you don't have access civicrm permission) as required.
This should only happen when entered URL was /. In other cases after successful login user should be redirected to the URL they tried before login page redirect.

**Solution**
Added destination parameter to _user_redirection function that redirects depending upon the user state and permissions. And after login it should redirect to last page they navigated from thats handled using hook  user login function to route to required destination after checking permissions for that page and user.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -5918,14 +5918,14 @@ function civihr_employee_portal_user_login_validate(&$form, &$form_state) {
 }
 
 function civihr_employee_portal_user_login_submit(&$form, &$form_state) {
+  $destination = !empty($_REQUEST['destination']) ? rawurldecode($_REQUEST['destination']) : '';
 
-    $destination = !empty($_REQUEST['destination']) ? $_REQUEST['destination'] : '';
-    if ($form_state['input']['forgot-password']) {
-        user_pass_submit($form, $form_state);
-        _drupal_session_write('custom_login_success_message', t('Details sent!'));
-        drupal_goto($destination);
-        return true;
-    }
+  if ($form_state['input']['forgot-password']) {
+    user_pass_submit($form, $form_state);
+    _drupal_session_write('custom_login_success_message', t('Details sent!'));
+    drupal_goto($destination);
+    return true;
+  }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5929,12 +5929,17 @@ function civihr_employee_portal_user_login_submit(&$form, &$form_state) {
...
 /**
- * Implements hook_user_login
+ * Implements hook_user_login().
  */
 function civihr_employee_portal_user_login(&$edit, $account) {
-  global $base_url;
+
   if (!isset($_POST['form_id']) || $_POST['form_id'] != 'user_pass_reset') {
-    $_GET['destination'] = user_access('access CiviCRM') ?  'civicrm/tasksassignments/dashboard#/tasks' : 'dashboard';
+    if (current_path() != drupal_get_destination()['destination']) {
```
davialexandre: This can be rewritten without the else and with less duplication:
```php
$sspEntryPath = 'dashboard';
$civicrmEntryPath = 'civicrm/tasksassignments/dashboard#/tasks';
if (current_path() != drupal_get_destination()['destination']) {
   $civicrmEntryPath = $_GET['destination'];
}

$_GET['destination'] = user_access('access CiviCRM') ? $civicrmEntryPath : $sspEntryPath;
```

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -7239,7 +7244,7 @@ function _user_redirection() {
   if ($user_is_anonymous) {
     if ((!in_array($current_path, $allowed_paths) && substr($current_path, 0, 4) !== 'user') ||
       $current_path == 'user/register') {
-      $redirect_path = 'welcome-page';
+      $current_path ? $redirect_path = 'welcome-page?destination=' . $current_path : $redirect_path = 'welcome-page';
```
davialexandre: This is more readable:
```php
$redirect_path = 'welcomen-page';
if ($current_path) {
  $redirect_path .= "?destination={$current_path}";
}
```

PR #270, opened 2017-02-08
PCHR-1953: Adding custom template for showing number of pending leave requests.

hitesh-jain:
**Task**
Add custom template for "Manager Leave" menu item that would act as placeholder to angular js files to display number of pending leave requests using template file.

**Solution**
Added function to load the angular js app and custom template to show number of pending leave requests and this function would be called from theme template.php to add markup to Manager Leave menu item using menu_link function.

### DIFF
--- a/civihr_employee_portal/civihr_leave_absences/civihr_leave_absences.module
+++ b/civihr_employee_portal/civihr_leave_absences/civihr_leave_absences.module
@@ -30,10 +30,10 @@ function civihr_leave_absences_block_view($delta = '') {
   $block = [];
   switch ($delta) {
     case 'my_leave':
-      $block['content'] = civihr_leave_absences_my_leave_get_markup();
+      $block['content'] = civihr_leave_absences_get_markup('my-leave');
       break;
     case 'manager_leave':
-      $block['content'] = civihr_leave_absences_manager_leave_get_markup();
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_leave_absences/civihr_leave_absences.module
```diff
@@ -81,6 +81,21 @@ function civihr_leave_absences_manager_leave_get_markup() {
 }
 
 /**
+ * Internal function for fetching custom markup from template/Angular Js to
+ * display number of pending leave requests.
+ *
+ * @return String
+ *    Custom markup from template file.
+ */
+function civihr_leave_absences_pending_leave_request_get_markup() {
```
AkA84: `civihr_leave_absences_pending_leave_request_get_markup`, `civihr_leave_absences_my_leave_get_markup`, `civihr_leave_absences_manager_leave_get_markup` are doing the same thing
```php
civicrm_resources_load('org.civicrm.reqangular', array('reqangular.min.js'));
civicrm_resources_load('uk.co.compucorp.civicrm.hrleaveandabsences', array('<module-name>.min.js'));

return civihr_leave_absences_load_file('<module-name-with-underscore>');
```

so I think you could refactor this and have just one function

PR #272, opened 2017-02-21
PCHR-1975: Added new role civihr_admin_local and updated required permissions.

hitesh-jain:
**Task**
Create civihr_admin_local role for all the new and existing civihr sites as the new default role with required permissions from page https://compucorp.atlassian.net/wiki/display/PCHR/Drupal+permissions+configuration.

**Solution**
1. Created role using hook_update to add to existing sites with the required permissions packaged into default_permissions feature module.
2. For adding new role for all the new civihr sites its handled using buildkit hr16 config https://github.com/civicrm/civicrm-buildkit/pull/303.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.install
+++ b/civihr_employee_portal/civihr_employee_portal.install
@@ -91,3 +91,16 @@ function civihr_employee_portal_update_7005() {
   $schema = module_invoke('civihr_employee_portal', 'schema');
   db_create_table('reports_configuration', $schema['reports_configuration']);
 }
+
+/**
+ * Create civihr_admin_local Role for existing sites.
+ */
+function civihr_employee_portal_update_7006() {
+  if (!array_search('civihr_admin_local', user_roles())) {
+    $role = new stdClass();
+    $role->name = 'civihr_admin_local';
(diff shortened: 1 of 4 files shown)


civihr_employee_portal/civihr_employee_portal.install
```diff
@@ -91,3 +91,16 @@ function civihr_employee_portal_update_7005() {
   $schema = module_invoke('civihr_employee_portal', 'schema');
   db_create_table('reports_configuration', $schema['reports_configuration']);
 }
+
+/**
+ * Create civihr_admin_local Role for existing sites.
```
omarabuhussein: the problem of drupal upgraders that they require you to update all the modules too (right ? ) so I just feel this is not need , what do you think ?

hitesh-jain: ya we need to update all the updates that are available, we need this for existing sites to add role.

omarabuhussein: does this will also update existing modules to their latest stable version ?

hitesh-jain: nope.

PR #274, opened 2017-02-22
PCHR-1976: Log State of Periods Cache on Start Date Validation of Leave Request

MiyaNoctem:
On undetermined occasions, trying to add a leave request will show an error message saying "This date period is not yet defined", even if the period for the given start date is properly defined.

Validation of leave request start dates against configured periods in CiviHR is done using data stored in drupal cache, which is built the first time a date is validated and expires every 6 minutes.  Watchdog log calls were added when setting and retrieving the data to cache, and on validation failure of leave request start date, to gain further inisght on the problem and plan a proper solution.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -839,39 +839,50 @@ function civihr_employee_portal_views_default_views() {
 }
 
 /**
- * Function for caching date periods returned from CiviCRM (avoiding expensive DB calls for each dashboard page hit)
+ * Function for caching date periods returned from CiviCRM (avoiding expensive 
+ * DB calls for each dashboard page hit)
  */
 function get_civihr_date_periods() {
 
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -839,39 +839,50 @@ function civihr_employee_portal_views_default_views() {
...
+            "Error obtaining periods from DB: $error ",
+            array(),
+            WATCHDOG_ALERT
+          );
+      }
 
-            // Cache the date periods for 5 minutes
-            cache_set('civihr_date_periods', $periods_data, 'cache', time() + 360);
-        }
+      // Cache the date periods for 5 minutes
```
davialexandre: The comment says 5 minutes, but it's actually 6. Can you please do this:
- Remove the comment
- Replace the `time() + 360` with `strtotime('+6 minutes')`

PR #278, opened 2017-03-16
PCHR-1890: Fix filtering on staff directory

mickadoo:
## Problem

A contextual filter was added to ensure only active roles are shown [in a previous ticket](https://compucorp.atlassian.net/browse/PCHR-1767). This had the side effect of only showing users with active roles in the staff directory.

The desired logic is to show all users with an active contract regardless of roles. We only care about the role if filtering by department.

## Solution

- Remove the contextual filter for role end.
- Add a hook to add this filter only if the department filter is set

### Notes 

- I wasn't sure where to put the `get_where_fields` function. If there was some util file it would probably be better to put it there.
- I ran the tests, but they are very basic. Of course it would be nice to have access to the `\view` class, but I'm not sure if the tests should depend on classes outside of the project.

### DIFF
--- a/civihr_employee_portal/tdd/HelperFunctionsTest.php
+++ b/civihr_employee_portal/tdd/HelperFunctionsTest.php
@@ -1,0 +1,43 @@
+<?php
+
+require_once __DIR__ . '/../views/civihr_employee_portal.views.inc';
+
+class HelperFunctionsTest extends PHPUnit_Framework_TestCase
+{
+  protected $sampleWherePart = [
+    [
+      'conditions' => [
+        [
+          'field' => 'foo1',
+          'value' => 'bar1'
+        ],
+        [
+          'field' => 'foo2',
+          'value' => 'bar2'
+        ],
+      ]
+    ],
+    [
+      'conditions' => [
+        [
+          'field' => 'foo3',
+          'value' => 'bar3'
+        ],
+        [
+          'field' => 'foo4',
+          'value' => 'bar4'
+        ],
+      ]
+    ]
+  ];
+
+  public function testGetWhereParts() {
+    $parts = [];
+    get_where_parts($this->sampleWherePart, $parts);
+    $expectedCount = 4;
+    $this->assertCount($expectedCount, $parts);
+    for ($i = 1; $i < $expectedCount + 1; $i++) {
+      $this->assertEquals('bar' . $i, $parts['foo' . $i]);
+    }
+  }
+}
--- a/civihr_employee_portal/views/civihr_employee_portal.views.inc
+++ b/civihr_employee_portal/views/civihr_employee_portal.views.inc
@@ -1144,13 +1144,36 @@ function removeInactiveManagers($results = array()) {
 }
 
 /**
+ * Implements hook_views_query_alter
+ *
+ * @param view $view
+ * @param views_plugin_query_default $query
+ */
+function civihr_employee_portal_views_query_alter(&$view, &$query) {
+  if ($view->name !== 'civihr_staff_directory') {
+    return;
+  }
+
+  $queryParts = array();
+  get_where_fields($query->where, $queryParts);
+  $departmentField = 'hrjc_role_hrjc_revision.role_department';
+  $roleEndField = 'hrjc_role_hrjc_revision.role_end_date';
+
+  // if department is set we must limit to only current roles
+  if (!empty($queryParts[$departmentField])) {
+    $group = $query->set_where_group('OR'); // add a new WHERE group
+    $query->add_where($group, $roleEndField, date('Y-m-d'), '>=');
+    $query->add_where($group, $roleEndField, NULL, 'IS');
+  }
+}
+
+/**
  * Implements hook_views_pre_view().
  * @param $view
  * @param $display_id
  * @param $args
  */
 function civihr_employee_portal_views_pre_view(&$view, &$display_id, &$args) {
-
   // Custom configuration start for items per page
   if (($view->name == 'civihr_staff_directory' && $display_id == 'page') ||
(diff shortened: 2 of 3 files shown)


civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1166,3 +1189,23 @@ function civihr_employee_portal_views_pre_view(&$view, &$display_id, &$args) {
     );
   }
 }
+
+/**
+ * @param mixed $part The $query->where part
```
davialexandre: Please add a description for this function

davialexandre: According to the Drupal code standards, the @param description should be in a new line https://www.drupal.org/node/1354#param

civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1166,3 +1189,23 @@ function civihr_employee_portal_views_pre_view(&$view, &$display_id, &$args) {
...
+ * @param array $fields An array to store the fields
+ */
+function get_where_fields($part, &$fields) {
```
davialexandre: It's fine to keep this function here, but note that all the other helper functions names in this file use camelCase instead of  snake_case, so maybe it would be better to use camelCase here as well for some consistency (I know, the hooks names are all snake_case, but at least we keep consistency with the existing code :man_shrugging: )

civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1166,3 +1189,23 @@ function civihr_employee_portal_views_pre_view(&$view, &$display_id, &$args) {
...
+  if (isset($part['conditions'])) {
+    $part = $part['conditions'];
+  }
+  if (isset($part['field']) && $part['field'] instanceof DatabaseCondition){
```
davialexandre: please add one space before the {

civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1144,13 +1144,36 @@ function removeInactiveManagers($results = array()) {
...
+ *
+ * @param view $view
+ * @param views_plugin_query_default $query
+ */
+function civihr_employee_portal_views_query_alter(&$view, &$query) {
+  if ($view->name !== 'civihr_staff_directory') {
+    return;
+  }
+
+  $queryParts = array();
```
davialexandre: since we're using the short array syntax, change this to `$queryParts = []`;

civihr_employee_portal/tdd/HelperFunctionsTest.php
```diff
@@ -0,0 +1,43 @@
...
+  public function testGetWhereParts() {
+    $parts = [];
+    get_where_parts($this->sampleWherePart, $parts);
+    $expectedCount = 4;
+    $this->assertCount($expectedCount, $parts);
+    for ($i = 1; $i < $expectedCount + 1; $i++) {
+      $this->assertEquals('bar' . $i, $parts['foo' . $i]);
+    }
+  }
+}
```
davialexandre: please add a new line at the end of the file

civihr_employee_portal/tdd/HelperFunctionsTest.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+require_once __DIR__ . '/../views/civihr_employee_portal.views.inc';
+
+class HelperFunctionsTest extends PHPUnit_Framework_TestCase
+{
```
davialexandre: To follow the Drupal code standards, for classes, don't put the opening brace in a newline

civihr_employee_portal/tdd/HelperFunctionsTest.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+require_once __DIR__ . '/../views/civihr_employee_portal.views.inc';
+
+class HelperFunctionsTest extends PHPUnit_Framework_TestCase
```
davialexandre: One of the problems we have is the lack of tests for our Drupal code, so +1 for adding this. I can see some problems with it, but since we don't know exactly what are the "best practices" in terms of Drupal testing, I think it's fine to keep this test. 

Just for the record, these are the problems I see:
- You're testing a helper function. If this was a class, this function would probably be private and we would not need a specific test for it. This might be a hint that we don't need a test for `get_where_parts`
- You're using a require_once to include the views file. This is not such a huge problem, but the test will break if we rename/move the views file or the test
- Having a dir named `tdd` is just an example that we have no idea of what are the "best practices" for tests in Drupal.  This dir will certainly have to be renamed (and then, we might have problems with the require_once)
- The test doesn't cover the case where the field is an instance of `DatabaseCondition` (again, fine, because we don't know the best way to mock this)

davialexandre:
@mickadoo there's no need to link the Jira ticket in the PR description

civihr_employee_portal/tdd/HelperFunctionsTest.php
mickadoo: Agree that it has a lot of problems. I'm not sure how to go about testing code that is an extension of a framework. I think writing Symfony bundles (which is similar to extensions) each bundle had it's own composer dependencies so it would have all the required code even if just testing outside of the framework.

Regarding the `get_where_parts` function visibility. It's common to have methods that don't make sense being public but that you still want to test. I think if we had proper classes here I would have no problem in making a utility class for dealing with queries. I'm in favor of smaller classes with very specific purpose.

mickadoo:
Closed because of wrong branch name in favor of https://github.com/compucorp/civihr-employee-portal/pull/279

PR #279, opened 2017-03-16
PCHR-1890: Fix filtering on staff directory

mickadoo:
## Problem

A contextual filter was added to ensure only active roles are shown [in a previous ticket](https://compucorp.atlassian.net/browse/PCHR-1767). This had the side effect of only showing users with active roles in the staff directory.

The desired logic is to show all users with an active contract regardless of roles. We only care about the role if filtering by department.

In addition the rules for what to show have changed:

- Instead of showing contract title we show role titlles
- Instead of showing contract location we show role locations

Also a bug where managers with relationships starting in the future was fixed as part of this.

## Solution

- Remove the contextual filter for role end.
- Add a hook to add this filter only if the department filter is set

### Notes 

- I wasn't sure where to put the `get_where_fields` function. If there was some util file it would probably be better to put it there.
- I ran the tests, but they are very basic. Of course it would be nice to have access to the `\view` class, but I'm not sure if the tests should depend on classes outside of the project.

This is a duplicate of [PR 278](https://github.com/compucorp/civihr-employee-portal/pull/278#discussion_r106401993) which used the wrong ticket number in the branch and commits

### DIFF
--- a/civihr_employee_portal/tdd/HelperFunctionsTest.php
+++ b/civihr_employee_portal/tdd/HelperFunctionsTest.php
@@ -30,9 +30,9 @@ class HelperFunctionsTest extends PHPUnit_Framework_TestCase {
     ]
   ];
 
-  public function testGetWhereParts() {
+  public function testGetWhereFields() {
     $parts = [];
-    get_where_parts($this->sampleWherePart, $parts);
+    getWhereFields($this->sampleWherePart, $parts);
     $expectedCount = 4;
     $this->assertCount($expectedCount, $parts);
     for ($i = 1; $i < $expectedCount + 1; $i++) {
--- a/civihr_employee_portal/tdd/HelperFunctionsTest.php
+++ b/civihr_employee_portal/tdd/HelperFunctionsTest.php
@@ -1,0 +1,43 @@
+<?php
+
+require_once __DIR__ . '/../views/civihr_employee_portal.views.inc';
+
+class HelperFunctionsTest extends PHPUnit_Framework_TestCase {
+  protected $sampleWherePart = [
+    [
+      'conditions' => [
+        [
+          'field' => 'foo1',
+          'value' => 'bar1'
+        ],
+        [
+          'field' => 'foo2',
+          'value' => 'bar2'
+        ],
+      ]
+    ],
+    [
+      'conditions' => [
+        [
+          'field' => 'foo3',
+          'value' => 'bar3'
+        ],
+        [
+          'field' => 'foo4',
+          'value' => 'bar4'
+        ],
+      ]
+    ]
+  ];
+
+  public function testGetWhereParts() {
+    $parts = [];
+    get_where_parts($this->sampleWherePart, $parts);
+    $expectedCount = 4;
+    $this->assertCount($expectedCount, $parts);
+    for ($i = 1; $i < $expectedCount + 1; $i++) {
+      $this->assertEquals('bar' . $i, $parts['foo' . $i]);
+    }
+  }
+
+}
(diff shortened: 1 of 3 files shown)


civihr_employee_portal/tdd/HelperFunctionsTest.php
```diff
@@ -0,0 +1,43 @@
...
+          'field' => 'foo4',
+          'value' => 'bar4'
+        ],
+      ]
+    ]
+  ];
+
+  public function testGetWhereParts() {
+    $parts = [];
+    get_where_parts($this->sampleWherePart, $parts);
```
davialexandre: You forgot to update the function name here (and also the test method name)

civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1117,29 +1117,61 @@ function removeAdditionalCommas($view) {
   return $fieldsToClean;
 }
 
-function removeInactiveDepartments($results = array()) {
-  foreach ($results as $key => $value) {
-    $startDate = !empty($value->hrjc_role_civicrm_contact_start_date) ? $value->hrjc_role_civicrm_contact_start_date : NULL;
-    $endDate = !empty($value->hrjc_role_civicrm_contact_end_date) ? $value->hrjc_role_civicrm_contact_end_date : NULL;
+/**
+ * Sets the name to NULL for relationships starting in future or ending in past
```
davialexandre: The description says you're setting the name to NULL, but the function is actually removing the results with inactive managers. Is the description correct?

civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1117,29 +1117,61 @@ function removeAdditionalCommas($view) {
...
+ *
+ * @param array $results
+ * @return array
```
davialexandre: Please add one empty line between the param and return sections

civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1117,29 +1117,61 @@ function removeAdditionalCommas($view) {
...
 
-function removeInactiveManagers($results = array()) {
-  foreach ($results as $key => $value) {
-    if ($value->civicrm_relationship_is_active == 0 ||
-            ($value->civicrm_relationship_end_date < date('Y-m-d') && $value->civicrm_relationship_end_date != NULL) ||
-            $value->civicrm_relationship_start_date > date('Y-m-d')) {
-      $value->civicrm_contact_civicrm_relationship_first_name = NULL;
-      $value->civicrm_contact_civicrm_relationship_last_name = NULL;
+/**
+ * @param \stdClass[] $results
```
davialexandre: even though both `removeInactiveRoleDetails` and `removeInactiveManagers` receive the same param, the param type on the their docblocks is different. Please use the same for both.

Please also add a description to this function

civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1148,6 +1149,8 @@ function removeInactiveManagers($results = []) {
 }
 
 /**
+ * Removes titles and locations from inactive roles fro the result
```
davialexandre: s/fro/from/

civihr_employee_portal/views/civihr_employee_portal.views.inc
mickadoo: Originally it was just setting the manager name to NULL but this resulted in fields with extra commas like ", Contact B"

I saw the query returns multiple results and removing those rows with the inactive relationships got rid of the data I didn't want while preserving everything else.

PR #280, opened 2017-03-20
PCHR-2074: Allow unauthenticated access to /yoti-connect 

mickadoo:
## Problem 

Anonymous users need to be able to access `/yoti-connect` to login with this plugin

## Solution

Add all routes starting with `/yoti-connect` to the whitelist for anonymous users

## Notes

- Since these rules were getting more complicated I used regular expression instead of simpler matching. I'm aware regex is a heavy operation, but since the check will only be done on unauthenticated routes I think it won't be noticeable
- It's not optimal initializing a class on each request, but after investigating usage of the dependency injection container I found out I cannot access the `\Civi` class at this point in the code. I don't think static methods are a good practice either, but I'm open to discussion.
- I added simple unit tests, but since there is no autoloader here I had to manually require the class to be tested. This is not ideal and I'm open to suggestions for changing it.


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -7475,4 +7460,4 @@ function get_civihr_version() {
   }
 
   return $civihrVersion;
-}
+}
```
davialexandre: missing newline at the end of the file (TIP: we have an .editorconfig file in this repo that, among other things, helps you to always have a newline at the end of the file. If you use PhpStorm, please install the editorconfig plugin to make this work)

civihr_employee_portal/src/Security/PublicFirewall.php
```diff
@@ -0,0 +1,56 @@
...
+  public static $publicRoutes = [
+    '/^welcome-page$/',
+    '/^sites\/default\/files\/logo.jpg$/', // if logo is missing
+    '/^request_new_account\/ajax$/', // from login page
+    '/^user((?!\/register).)*$/', // user* except user/register
+    '/^yoti-connect*/' // yoti login plugin
+  ];
+
+  /**
+   * @param \stdClass $user the user to check access for
```
davialexandre: Please add a description to this method and also remember that in Drupal code standard, the param description should be in a newline

civihr_employee_portal/src/Security/PublicFirewall.php
```diff
@@ -0,0 +1,56 @@
...
+namespace Drupal\civihr_employee_portal\Security;
+
+/**
+ * This class is responsible for checking if anonymous users can access
+ * certain routes
+ */
+class PublicFirewall {
+
+  /**
+   * Anonymous users can access these paths
```
davialexandre: For properties, the description should be below the @var tag

civihr_employee_portal/src/Security/PublicFirewall.php
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Check if route matches against all public routes
+   *
+   * @param string $route
+   * @return bool
+   */
+  private function allowAnonymousAccess($route) {
+    foreach (self::$publicRoutes as $pattern) {
+      if (preg_match($pattern, $route)) {
```
davialexandre: given that most URL's are fixed and don't really required a regex, I'm thinking of a little improvement: For fixed routes, we just use a simple string, but for regex, we should prefix it with something (and @, for example):

```php
public static $publicRoutes = [
    'welcome-page',
    'sites/default/files/logo.jpg', // if logo is missing
    'request_new_account/ajax', // from login page
    '@/^user((?!\/register).)*$/', // user* except user/register
    '@/^yoti-connect*/' // yoti login plugin
];
```

Then this method would be something like:
```php
foreach (self::$publicRoutes as $routes) {
  if ($route[0] == '@') {
     // preg_match with the string without the @
  } else {
    // just a normal comparison
  }
}
```

The advantages:
- you don't need the more complex syntax and escaping for fixed URLs
- you avoid unnecessary regex checks

civihr_employee_portal/tdd/PublicFirewallTest.php
```diff
@@ -0,0 +1,73 @@
...
+    $firewall = new PublicFirewall();
+    $this->assertFalse($firewall->canAccess($user, $route));
+  }
+
+  /**
+   * @test
+   * @dataProvider publicRouteProvider
+   * @param $route
+   */
+  public function willAllowAnonymousUserAsExpected($route) {
```
davialexandre: Instead of using the @test tag, can you please prefix the tests with test (that is, willAllowAnonymousUserAsExpected will be become testWillAllowAnonymousUserAsExpected)? 

Even though I prefer the tag myself, we don't use it for the other tests, so it's best to keep consistency with the rest of the codebase

civihr_employee_portal/civihr_employee_portal.module
mickadoo: ah right, I did see the editorconfig file but I actually thought that this was for some people using vim or something :-D
I just installed it now

civihr_employee_portal/src/Security/PublicFirewall.php
mickadoo: I did a quick benchmark, for 1,000,000 iterations the time was 4.1 seconds for all regex vs 2.9 for part-regex so I think that's worth changing, especially if this list grows in the future.

For our specific use case it would be safe to say that `/` means a regex since the path is coming from request_path() which trims leading and trailing slashes but I also think that using `/` to denote a regex when URLs are involved might be confusing.

I'm a bit wary of this becoming too complex if we introduce our own symbols  How about:

```
$publicRoutes = [
  'welcome-page',
  'sites/default/files/logo.jpg', // if logo is missing
  'request_new_account/ajax', // from login page
  '@^user((?!\/register).)*$@', // user* except user/register
  '@^yoti-connect*@' // yoti login plugin
];
```

Where we would use `@` as the regex delimiter. I think most developers familiar with regex would understand without any further explanation.

civihr_employee_portal/tdd/PublicFirewallTest.php
mickadoo: Sure, I just thought I'd chance my arm putting it in to see the reaction. We had a discussion about this where I worked before, I was pro using 'test' prefix first but they won me over :-)

civihr_employee_portal/src/Security/PublicFirewall.php
davialexandre: Sounds good for me. You might want to add a comment saying we're using @ as delimiter instead of /, though. I'm afraid not all developers are aware that we can use different delimiters (to be honest, before I didn't remember it myself until you mentioned it)

PR #283, opened 2017-03-22
PCHR-2097: Allow overlapping absence when absence rejected

mickadoo:
## Problem

When you create a leave request and it is rejected you cannot create another leave request for the same period.

This is because the query to check for duplicates only ignores cancelled requests.

```
 $absencesDataQuery = db_select('absence_list', 'al')
  ->condition('contact_id', $_SESSION['CiviCRM']['userID'])
  ->condition('absence_status', self::CANCELLED_LEAVE_TYPE_ID, '<>')
  ->fields('al', array('absence_start_date_timestamp', 'absence_end_date_timestamp'));
```

## Solution

Add a condition so absences with the rejected status will also be ignored in the validation

### DIFF
--- a/civihr_employee_portal/src/Forms/AbsenceRequestForm.php
+++ b/civihr_employee_portal/src/Forms/AbsenceRequestForm.php
@@ -9,7 +9,7 @@ class AbsenceRequestForm {
     protected $form_state;
 
     const CANCELLED_LEAVE_TYPE_ID = 3;
-    const DENIED_LEAVE_TYPE_ID = 9;
+    const REJECTED_LEAVE_TYPE_ID = 9;
 
     /**
      * Constructor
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/src/Forms/AbsenceRequestForm.php
```diff
@@ -418,7 +419,7 @@ protected function duplicate_dates_exist($requestedStartTimestamp, $requestedEnd
         $absencesDataQuery = db_select('absence_list', 'al')
             ->condition('contact_id', $_SESSION['CiviCRM']['userID'])
             ->condition('absence_status', self::CANCELLED_LEAVE_TYPE_ID, '<>')
-            //->condition('absence_end_date_timestamp', strtotime('today'), '>')
+            ->condition('absence_status', self::DENIED_LEAVE_TYPE_ID, '<>')
```
davialexandre: In the UI, this status is displayed as "Rejected". To keep consistency with the UI, I think it's better to say "Rejected" instead of "Denied"?  (the ticket title and description also use "Rejected")

mickadoo: Good point

PR #286, opened 2017-03-29
PCHR-2083: Added yoti_connect module.

hitesh-jain:
**Task**
Add Yoti Login button to User edit page and Login (welcome) page, that should allow the user to login using yoti service.

**Solution**
1. Added yoti_connect module from https://compucorp.atlassian.net/wiki/display/PCHR/Yoti+connect to sites/all/modules/civihr-custom repo.
2. Enable yoti_connect module and configure yoti_connect module from admin/config/people/yoti-connect page to provide required settings to connect to yoti account like App ID , SDK ID.


omarabuhussein:
for the next time , but adding a module and adding other stuff should be in separate commits

PR #289, opened 2017-03-30
PCHR-2105 Restrict to Work Emails in Staff Directory

mickadoo:
## Problem

The staff-directory view currently uses the primary email, regardless of whether it is personal or not. It should use only the work e-mail(s) and be blank if there are none.

In addition, rendering a list of emails will show them as plain text unless the `civicrm_handler_field_email` is used. However, this handler does not support multiple e-mails - for example 'a@gmail.com b@gmail.com' will result in `<a href="a@gmail.com b@gmail.com">a@gmail.com b@gmail.com</a>`

## Solution

Change the view to use an existing field, the 'work_email' field defined on the `hrjc_contact_details` table. This field will be empty if there are no work emails, or a list of them (space separated) if there are some.

In order to render them properly with the links I added a new custom field handler that will make links from each of the e-mails.

### DIFF
--- a/civihr_employee_portal/views/includes/civihr_employee_portal_handler_email.inc
+++ b/civihr_employee_portal/views/includes/civihr_employee_portal_handler_email.inc
@@ -1,0 +1,42 @@
+
+<?php
+
+/**
+ * Custom field handler for emails, including multiple emails since
+ * @see civicrm_handler_field_email only works with single mails
+ */
+class civihr_employee_portal_handler_email extends views_handler_field {
+  /**
+   * @param array $values
+   *  All the values to be rendered by the view
+   *
+   * @return string
+   *  The rendered emails separated by spaces
+   */
+  public function render($values) {
+    $emails = $this->get_value($values);
+
+    if (!$emails) {
+      return '';
+    }
+
+    $this->sanitize_value($emails);
+    $emails = array_filter(explode(' ', $emails));
+    array_walk($emails, function (&$email) {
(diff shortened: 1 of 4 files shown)


civihr_employee_portal/views/includes/civihr_employee_portal_handler_email.inc
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Custom field handler for emails, including multiple emails since
+ * @see civicrm_handler_field_email only works with single mails
+ */
+class civihr_employee_portal_handler_email extends views_handler_field {
+  /**
+   * @param array $values
+   *  All the values to be rendered by the view
+   * @return string
```
davialexandre: Please add a empty line between the `@param` and `@return` sections

civihr_employee_portal/views/includes/civihr_employee_portal_handler_email.inc
```diff
@@ -0,0 +1,40 @@
...
+      $email = $this->linkify($email);
+    });
+
+    return implode(' ', $emails);
+  }
+
+  /**
+   * @param string $email
+   *  The email to be put in a link
+   * @return string
```
davialexandre: Add an empty line between the sections here as well

davialexandre:
@mickadoo the last commit (36d8380) doesn't really add anything useful to the commit history. Can you please squash it together with 8c5a32d ?

PR #291, opened 2017-04-05
PCHR-2120: Rewrite the Query to build Absence Activity View to pick data from L&A related tables

tunbola:
The Absence Activity view was picking data from the HRabsence extension related tables but since the extension will soon be deprecated in favour of the Leave and Absence extension, 
This PR makes the change by modifying the query to build the Absence activity view to pick data from L&A related tables.

The Leave and Absence report view now has its data populated from L&A tables although some view handlers need to be updated as some of the columns are showing incorrect values the columns are Absence status (Picking statues from HRabsence leave request statuses), Absence Duration In Days, Absence Amount Taken, Absence Amount Accrued, Absence Absolute Amount . This will be fixed in subsequent PR's.

Here is a sample cross-section of the Leave and Absence Reports Page.

civihr custom report - civihr 1 6 demo 2017-04-05 18-47-18

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -315,50 +315,32 @@ function _rebuild_absence_activity_view() {
   $civi_settings = parse_url(CIVICRM_DSN);
   $civi_db_name = trim($civi_settings['path'], '/');
 
-  $activityTypesIds = _get_absence_activity_types_ids();
-  if (!empty($activityTypesIds)) {
-    db_query('DROP VIEW IF EXISTS absence_activity');
-    db_query("CREATE VIEW absence_activity AS
-              SELECT ac.contact_id AS absence_contact_id,
-                absence_activity.id AS absence_activity_id,
-                ov.label AS absence_type,
-                DATE(a2.activity_date_time) AS absence_date,
-                DATE_FORMAT(DATE(a2.activity_date_time), '%Y-%m') AS absence_month,
-                (SELECT MIN(DATE(activity_date_time)) FROM {$civi_db_name}.civicrm_activity WHERE source_record_id = a2.source_record_id) AS absence_start_date,
-                DATE_FORMAT((SELECT MIN(DATE(activity_date_time)) FROM {$civi_db_name}.civicrm_activity WHERE source_record_id = a2.source_record_id), '%Y-%m') AS absence_start_date_month,
-                (SELECT MAX(DATE(activity_date_time)) FROM {$civi_db_name}.civicrm_activity WHERE source_record_id = a2.source_record_id) AS absence_end_date,
-                DATE_FORMAT((SELECT MAX(DATE(activity_date_time)) FROM {$civi_db_name}.civicrm_activity WHERE source_record_id = a2.source_record_id), '%Y-%m') AS absence_end_date_month,
-                CONCAT(IF (DATE_FORMAT(a2.activity_date_time, '%w') = 0, 7, DATE_FORMAT(a2.activity_date_time, '%w')), '. ', DATE_FORMAT(a2.activity_date_time, '%W')) AS absence_day_of_week,
-                a2.duration AS absence_duration,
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -315,50 +315,32 @@ function _rebuild_absence_activity_view() {
...
-              AND
-              absence_activity.activity_type_id IN (" . implode(',', $activityTypesIds) . ")");
+  db_query('DROP VIEW IF EXISTS absence_activity');
+  db_query("CREATE VIEW absence_activity AS
+            SELECT lr.contact_id AS absence_contact_id,
+              lr.id AS absence_activity_id,
+              at.title as absence_type,
+              lrd.date AS absence_date,
+              DATE_FORMAT(DATE(lrd.date), '%Y-%m') AS absence_month,
+              (SELECT MIN(DATE(date)) FROM {$civi_db_name}.civicrm_hrleaveandabsences_leave_request_date WHERE leave_request_id = lr.id) AS absence_start_date,
```
davialexandre: wouldn't it make more sense to just use `lr.from_date`?

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -315,50 +315,32 @@ function _rebuild_absence_activity_view() {
...
+              DATE_FORMAT((SELECT MIN(DATE(date)) FROM {$civi_db_name}.civicrm_hrleaveandabsences_leave_request_date WHERE leave_request_id = lr.id), '%Y-%m') AS absence_start_date_month,
+              (SELECT MAX(DATE(date)) FROM {$civi_db_name}.civicrm_hrleaveandabsences_leave_request_date WHERE leave_request_id = lr.id) AS absence_end_date,
```
davialexandre: and here just use `lr.to_date`

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -315,50 +315,32 @@ function _rebuild_absence_activity_view() {
...
+              DATE_FORMAT((SELECT MAX(DATE(date)) FROM {$civi_db_name}.civicrm_hrleaveandabsences_leave_request_date WHERE leave_request_id = lr.id), '%Y-%m') AS absence_end_date_month,
+              CONCAT(IF (DATE_FORMAT(lrd.date, '%w') = 0, 7, DATE_FORMAT(lrd.date, '%w')), '. ', DATE_FORMAT(lrd.date, '%W')) AS absence_day_of_week,
+              ABS(bc.amount) as absence_duration,
+              IF(lr.request_type != 'toil', ABS(bc.amount), NULL) AS absence_amount_taken,
+              IF(lr.request_type = 'toil', bc.amount, NULL) AS absence_amount_accrued,
+              IF(lr.request_type = 'toil', -bc.amount, -bc.amount) AS absence_absolute_amount,
```
davialexandre: On the previous query, the third param of the IF didn't have the - sign

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -315,50 +315,32 @@ function _rebuild_absence_activity_view() {
...
+              lr.status_id AS absence_status,
+              IF(lr.request_type = 'toil', 1, 0) AS absence_is_credit
+            FROM {$civi_db_name}.civicrm_hrleaveandabsences_leave_request lr
+            LEFT JOIN {$civi_db_name}.civicrm_hrleaveandabsences_leave_request_date lrd ON lr.id = lrd.leave_request_id
```
davialexandre: I think we can use INNER JOINS instead of LEFT. If a leave request doesn't have leave request dates and balance changes, than it should not be part of the results

civihr_employee_portal/civihr_employee_portal.module
tunbola: For some reason, On the old view, absolute amount for TOIL is always displayed in negative and positive for leave requests. 
To achieve this I had to negate toil since its always has positive balance changes on L&A and make that of leave requests positive since it always has negative balance changes.

PR #296, opened 2017-04-13
PCHR-1728: Add SMTP Module to Dependencies

MiyaNoctem:
## Problem
SMTP module had to be installed manually on every CiviHR installation.

## Solution
Added installation of SMTP module to drush makefile for CiviHR in https://github.com/civicrm/civicrm-buildkit/pull/310. Added SMTP module to dependencies of employee portal, so it gets enabled on installation of SSP.


omarabuhussein:
but is this really a dependency for civihr employee portal , if it just to enable it then you can do that in the buildkit

MiyaNoctem:
Discussed this approach with Omar.  Since SMTP module is not relly a dependency, enabling will be handled in installation profile in buildkit.

PR #299, opened 2017-04-27
PCHR-2143: Always Show Documents Widgets

mickadoo:
## Overview
The document and document manager widget s are hidden on the SSP dashboard if you don't have documents assigned to you. They should always be shown, with a message saying there are no documents.

## Before

If you didn't have documents assigned you wouldn't see the "Documents" or "Document Manager" widgets.

## After
You will always see the "Documents" and "Document Manager" widgets, regardless of whether you have anything assigned to you.

screenshot from 2017-04-27 15-38-02

## Technical Details
This PR makes changes to a view so reverting the view will be necessary.

## Comments
Changing the view was a small issue, but the template does not expect empty results so I needed to fix the warnings from PHP.

---

- [ ] Tests Pass

It looks like the tests haven't been run for a while and are broken.


tunbola:
@mickadoo, Just a small edit on the PR description s/widget s/widgets

PR #303, opened 2017-05-22
PCHR-2269: Fix permission check for view

mickadoo:
## Overview

The document block should be viewable by all staff roles. Currently it's restricted to managers / admin.

This will be accomplished by using the existing "View My Tasks Blocks", but since it now encompassed Document Blocks it will be renamed to "View My Tasks And Documents Blocks"

The documents manager block should be viewable only by managers or admins so instead of using the default view permissions it will use a role-based permission check.

## Before
The documents block and documents manager block is only viewable by managers or admins.

## After
The documents block permission check doesn't use role, but instead usees the "view my tasks and documents block" permission, which is available to all authenticated users.

The documents manager block uses a role check for CiviHR Manager, CiviHR Admin and CiviHR Local Admin.

---

- [ ] Tests Pass
StaffDirectoryResultTest.php and UsernameToExternalIdTest.php do not work and did not work before. The other tests pass.


civihr_employee_portal/views/views_export/views_documents.inc
```diff
@@ -351,6 +351,13 @@ $handler->display->display_options['arguments']['target_contact_id']['summary_op
 $handler = $view->new_display('block', 'Documents Dashboard Manager block', 'block_1');
 $handler->display->display_options['defaults']['title'] = FALSE;
 $handler->display->display_options['title'] = 'Documents Manager';
+$handler->display->display_options['defaults']['access'] = FALSE;
+$handler->display->display_options['access']['type'] = 'role';
+$handler->display->display_options['access']['role'] = array(
+  55120974 => '55120974',
+  37588037 => '37588037',
+  57573969 => '57573969',
```
tunbola: These are roleID's that won't change right?
Is it not possible to use their human readable name equivalents?

mickadoo: It seems like using RIDs in exports is a known issue. There's a whole section about it [here](https://www.drupal.org/docs/7/modules/features/exportables-and-user-role-ids-in-features). Drupal 8 will apparently use the machine name.

The employee portal features use [role export](https://www.drupal.org/project/role_export) which uses a hash of the machine name when exporting. From their page:

> Hashing machine names? That will never be unique, are you out of your mind?
Certainly not. There will not be a large number of roles on a Drupal site (if you have more than 50 you're doing it wrong). So we can assume that the probability of an id clash is very, very, very, very low. If you encounter such a case, please report it in the issue queue.

I tried changing it to use the machine names in the views file but it didn't work.

PR #305, opened 2017-05-26
PCHR-2144: Updates to SSP documents widget

bmnepali:
## Overview
Currently in the view of Documents Widget in SSP, modal is displayed using `ctools`. And this needs to address latest requirements to New Document Modal as in T&A. for this an angular app needs to be injected to the view. Also with reusing the modal form T&A, respective fields are to be hidden for the 'staff' role.

## Before
screen shot 2017-05-26 at 2 24 13 pm

## After
screen shot 2017-05-31 at 11 06 22 pm

## Technical Details
1. Pass 'staff' as role to second argument to 'modalDocument(id, role)' in  `views/templates/views-view-table--Documents--block.tpl.php` which will be passed to Modal Document Controller in T&A through `modalDocument()` in `js/ta-documents-app.js` and using the role to hide the respective fields in the document modal in file `uk.co.compucorp.civicrm.tasksassignments/views/modal/document.html`

2. The angular app injected in the SSP was created originally to work only on the admin, meaning it assumed the user would always have admin permissions. Now that users with the `civihr_staff` role also have access, some permission changes were required: 
- The `view all activities` was added because the Activity API is used internally by the Document API
- The `access AJAX API` was added because it's the very minimum necessary for the frontend to use any API
- The `access Tasks and Assignments Files ` permission was created so users can have access to and upload files attached to a Document
- The `access uploaded files` permission was added so users can download files attached to a Document
3. In order to inject the angular app inside a Drupal page, we need the `CiviCRM Resources` module. This was added as a dependecy of the CiviHR Employee portal module
4. Some settings data required to bootstrap angular app is required, which is collected and passed to view in file `civihr_employee_portal/civihr_employee_portal.module`
5. Also in file `civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php`
-> For angular app to be injected to the view, additional `data-ta-documents`  and `ng-controller="ModalController"` are attached to the rot markup.
-> Settings data are made available by passing data form `Drupal` to global `CRM` js object
-> Added minimal angular code in file`js/ta-documents-app.js`
-> Finally, bootstrapped angular app when T&A modules are loaded completely
```js
// Listen for ready event when T&A finishes loading all modules
    document.addEventListener('taReady', function (e) {
      angular.bootstrap(angular.element("[data-ta-documents]"), ['taDocuments']);
    });
```
6. The following code is removed:
```js
modalInstance.result.then(function (results) {
          $scope.$broadcast('documentFormSuccess', results, data);

          if (results.open) {
            $rootScope.modalDocument(data);
          }
        }, function () {
          $log.info('Modal dismissed');
        });
```
Because, the above function is needed to broadcast the `documentFormSucess` event, on which the listener will be checking the status of the updated document and will remove it if it is approved or rejected form the list of documents otherwise keeps them in the list. But, there is nothing to be processed in document list case of SSP. So, removing it.

Also open a document, if $scope.document.open was set true. But, Since we have removed "Save and New" button as per the wireframe, $scope.document.open won't be set true. So, we wont be needing to open new modal once the document is saved. So, we won't be needing it any more. It is kept intact in T&A as it is appropriate.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -549,7 +549,6 @@ function civihr_employee_portal_init() {
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/lib/sweetalert/sweetalert.css");
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/lib/tablesaw/tablesaw.css");
 
-
     _add_script_to_page('/js/tasks.js', 'dashboard');
 
     _add_script_to_page([
@@ -563,6 +562,9 @@ function civihr_employee_portal_init() {
       '/lib/sweetalert/sweetalert.min.js'
     ]);
 
+    civicrm_resources_load('org.civicrm.reqangular', ['reqangular.min.js']);
+    _add_script_to_page('/js/ta-documents-app.js', 'dashboard');
+
     _rebuild_view('absence_list');
     _rebuild_view('length_of_service');
     _rebuild_view('absence_activity');
@@ -572,13 +574,12 @@ function civihr_employee_portal_init() {
     _rebuild_view('documents');
 
     _load_ta_settings();
-
-    // Page load scripts
-    civicrm_resources_load('org.civicrm.reqangular', ['reqangular.min.js']);
-    _add_script_to_page('/js/ta-documents-app.js', 'dashboard');
   }
 }
 
+/**
+ * Pass tasksAssignments data to Global Drupal.settings object
+ */
 function _load_ta_settings() {
   $taSettings = civicrm_api3('TASettings', 'get')['values'];
 
@@ -587,15 +588,15 @@ function _load_ta_settings() {
   }
 
   drupal_add_js([
-    'Tasksassignments' => array(
+    'tasksAssignments' => [
       'extensionPath' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.tasksassignments'),
       'case_extension' => !empty(CRM_Core_Component::get('CiviCase')),
       'settings' => $taSettings,
-      'permissions' => array(
+      'permissions' => [
         'delete_tasks_and_documents' => CRM_Core_Permission::check('delete Tasks and Documents'),
-      ),
+      ],
       'debug' => $config->debug
-    ),
+    ],
     'adminId' => CRM_Core_Session::getLoggedInContactID(),
     'contactId' => CRM_Utils_Request::retrieve('cid', 'Integer')
   ], 'setting');
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -581,6 +581,7 @@ function civihr_employee_portal_init() {
  * Pass tasksAssignments data to Global Drupal.settings object
  */
 function _load_ta_settings() {
+  $config = CRM_Core_Config::singleton();
   $taSettings = civicrm_api3('TASettings', 'get')['values'];
 
   if (!empty($taSettings['days_to_create_a_document_clone'])) {
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -569,6 +569,32 @@ function civihr_employee_portal_init() {
     _rebuild_view('appraisal');
     _rebuild_view('tasks');
     _rebuild_view('documents');
+
+    $extensionPath = CRM_Extension_System::singleton()->getMapper()->keyToUrl('uk.co.compucorp.civicrm.tasksassignments');
+
+    $allSettings = array();
+    $taSettings = civicrm_api3('TASettings', 'get');
+    foreach ($taSettings['values'] as $key => $value) {
+      $allSettings[$key] = $value['value'];
+    }
+    if (!empty($allSettings['days_to_create_a_document_clone'])) {
+      $allSettings['days_to_create_a_document_clone'] = (int)$allSettings['days_to_create_a_document_clone'];
+    }
+
+    $settings = [
+      'Tasksassignments' => array(
+        'extensionPath' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.tasksassignments'),
+        'case_extension' => !empty(CRM_Core_Component::get('CiviCase')),
+        'settings' => $allSettings,
+        'permissions' => array(
+          'delete_tasks_and_documents' => CRM_Core_Permission::check('delete Tasks and Documents'),
+        ),
+      ),
+      'adminId' => CRM_Core_Session::getLoggedInContactID(),
+      'contactId' => _get_civicrm_contact_id_by_drupal_uid($user->uid),
+      'debug' => $config->debug
+    ];
+    drupal_add_js($settings, 'setting');
   }
 }
 
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -130,43 +130,30 @@ function _setup_modals() {
 }
 
 /**
- * Adds a script to a specific page form other extensions
- *
- * @param array $resources (key as Module and value as array of scripts)
- * @param string $page
- */
-function _add_module_script_to_page($resources, $page = NULL) {
-
-  if (!is_null($page) && arg(0) != $page) {
-    return;
-  }
-
-  if (is_array($resources)) {
-    foreach ($resources as $key => $value) {
-      civicrm_resources_load($key, $value);
-    }
-  }
-}
-
-/**
  * Adds a script to a specific page.
  *
- * @param string $scriptFile the name of the script file (extension included)
+ * @param string $resources the name of the script file (extension included)
+ * @param string $extension
  * @param string $page
  */
-function _add_script_to_page($scriptFile, $page = NULL) {
+function _add_script_to_page($resources, $extension = NULL, $page = NULL) {
 
   if (!is_null($page) && arg(0) != $page) {
     return;
   }
 
-  if (is_array($scriptFile)) {
-    foreach ($scriptFile as $value) {
-      drupal_add_js(drupal_get_path('module', 'civihr_employee_portal') . $value, 'file');
-    }
+  if ($extension) {
+    civicrm_resources_load($extension, $resources);
   }
   else {
-    drupal_add_js(drupal_get_path('module', 'civihr_employee_portal') . $scriptFile, 'file');
+    if (is_array($resources)) {
+      foreach ($resources as $value) {
+        drupal_add_js(drupal_get_path('module', 'civihr_employee_portal') . $value, 'file');
+      }
+    }
+    else {
+      drupal_add_js(drupal_get_path('module', 'civihr_employee_portal') . $resources, 'file');
+    }
   }
 }
 
@@ -569,8 +556,6 @@ function civihr_employee_portal_init() {
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/lib/sweetalert/sweetalert.css");
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/lib/tablesaw/tablesaw.css");
 
-    _add_script_to_page('/js/tasks.js', 'dashboard');
-
     _add_script_to_page([
       '/js/scripts.js',
       '/js/filters.js',
@@ -582,13 +567,8 @@ function civihr_employee_portal_init() {
       '/lib/sweetalert/sweetalert.min.js'
     ]);
 
-    _add_module_script_to_page([
-      'org.civicrm.reqangular' => [
-        'reqangular.min.js'
-        ]
-      ], 'dashboard');
-
-    _add_script_to_page('/js/ta-documents-app.js', 'dashboard');
+    _add_script_to_page(['reqangular.min.js'], 'org.civicrm.reqangular', 'dashboard');
+    _add_script_to_page(['/js/tasks.js', '/js/ta-documents-app.js'], NULL, 'dashboard');
 
     _rebuild_view('absence_list');
     _rebuild_view('length_of_service');
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -130,6 +130,25 @@ function _setup_modals() {
 }
 
 /**
+ * Adds a script to a specific page form other extensions
+ *
+ * @param array $resources (key as Module and value as array of scripts)
+ * @param string $page
+ */
+function _add_module_script_to_page($resources, $page = NULL) {
+
+  if (!is_null($page) && arg(0) != $page) {
+    return;
+  }
+
+  if (is_array($resources)) {
+    foreach ($resources as $key => $value) {
+      civicrm_resources_load($key, $value);
+    }
+  }
+}
+
+/**
  * Adds a script to a specific page.
  *
  * @param string $scriptFile the name of the script file (extension included)
@@ -140,6 +159,7 @@ function _add_script_to_page($scriptFile, $page = NULL) {
   if (!is_null($page) && arg(0) != $page) {
     return;
   }
+
   if (is_array($scriptFile)) {
     foreach ($scriptFile as $value) {
       drupal_add_js(drupal_get_path('module', 'civihr_employee_portal') . $value, 'file');
@@ -562,7 +582,12 @@ function civihr_employee_portal_init() {
       '/lib/sweetalert/sweetalert.min.js'
     ]);
 
-    civicrm_resources_load('org.civicrm.reqangular', ['reqangular.min.js']);
+    _add_module_script_to_page([
+      'org.civicrm.reqangular' => [
+        'reqangular.min.js'
+        ]
+      ], 'dashboard');
+
     _add_script_to_page('/js/ta-documents-app.js', 'dashboard');
 
     _rebuild_view('absence_list');
@@ -598,8 +623,7 @@ function _load_ta_settings() {
       ],
       'debug' => $config->debug
     ],
-    'adminId' => CRM_Core_Session::getLoggedInContactID(),
-    'contactId' => CRM_Utils_Request::retrieve('cid', 'Integer')
+    'adminId' => CRM_Core_Session::getLoggedInContactID()
   ], 'setting');
 }
 
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -549,6 +549,7 @@ function civihr_employee_portal_init() {
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/lib/sweetalert/sweetalert.css");
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/lib/tablesaw/tablesaw.css");
 
+
     _add_script_to_page('/js/tasks.js', 'dashboard');
 
     _add_script_to_page([
@@ -570,32 +571,34 @@ function civihr_employee_portal_init() {
     _rebuild_view('tasks');
     _rebuild_view('documents');
 
-    $extensionPath = CRM_Extension_System::singleton()->getMapper()->keyToUrl('uk.co.compucorp.civicrm.tasksassignments');
+    _load_ta_settings();
 
-    $allSettings = array();
-    $taSettings = civicrm_api3('TASettings', 'get');
-    foreach ($taSettings['values'] as $key => $value) {
-      $allSettings[$key] = $value['value'];
-    }
-    if (!empty($allSettings['days_to_create_a_document_clone'])) {
-      $allSettings['days_to_create_a_document_clone'] = (int)$allSettings['days_to_create_a_document_clone'];
-    }
-
-    $settings = [
-      'Tasksassignments' => array(
-        'extensionPath' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.tasksassignments'),
-        'case_extension' => !empty(CRM_Core_Component::get('CiviCase')),
-        'settings' => $allSettings,
-        'permissions' => array(
-          'delete_tasks_and_documents' => CRM_Core_Permission::check('delete Tasks and Documents'),
-        ),
-      ),
-      'adminId' => CRM_Core_Session::getLoggedInContactID(),
-      'contactId' => _get_civicrm_contact_id_by_drupal_uid($user->uid),
-      'debug' => $config->debug
-    ];
-    drupal_add_js($settings, 'setting');
+    // Page load scripts
+    civicrm_resources_load('org.civicrm.reqangular', ['reqangular.min.js']);
+    _add_script_to_page('/js/ta-documents-app.js', 'dashboard');
   }
+}
+
+function _load_ta_settings() {
+  $taSettings = civicrm_api3('TASettings', 'get')['values'];
+
+  if (!empty($taSettings['days_to_create_a_document_clone'])) {
+    $taSettings['days_to_create_a_document_clone'] = (int)$taSettings['days_to_create_a_document_clone'];
+  }
+
+  drupal_add_js([
+    'Tasksassignments' => array(
+      'extensionPath' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.tasksassignments'),
+      'case_extension' => !empty(CRM_Core_Component::get('CiviCase')),
+      'settings' => $taSettings,
+      'permissions' => array(
+        'delete_tasks_and_documents' => CRM_Core_Permission::check('delete Tasks and Documents'),
+      ),
+      'debug' => $config->debug
+    ),
+    'adminId' => CRM_Core_Session::getLoggedInContactID(),
+    'contactId' => CRM_Utils_Request::retrieve('cid', 'Integer')
+  ], 'setting');
 }
 
 function _rebuild_reports_views() {
--- a/civihr_employee_portal/features/civihr_default_permissions/civihr_default_permissions.features.user_permission.inc
+++ b/civihr_employee_portal/features/civihr_default_permissions/civihr_default_permissions.features.user_permission.inc
@@ -27,6 +27,7 @@ function civihr_default_permissions_user_default_permissions() {
       'civihr_admin' => 'civihr_admin',
       'civihr_admin_local' => 'civihr_admin_local',
       'civihr_manager' => 'civihr_manager',
+      'civihr_staff' => 'civihr_staff',
     ),
     'module' => 'civicrm',
   );
@@ -2099,6 +2100,7 @@ function civihr_default_permissions_user_default_permissions() {
       'civihr_admin' => 'civihr_admin',
       'civihr_admin_local' => 'civihr_admin_local',
       'civihr_manager' => 'civihr_manager',
+      'civihr_staff' => 'civihr_staff',
     ),
     'module' => 'civicrm',
   );
--- a/civihr_employee_portal/features/civihr_default_permissions/civihr_default_permissions.features.user_permission.inc
+++ b/civihr_employee_portal/features/civihr_default_permissions/civihr_default_permissions.features.user_permission.inc
@@ -99,6 +99,17 @@ function civihr_default_permissions_user_default_permissions() {
     'module' => 'civicrm',
   );
 
+  // Exported permission: 'access Tasks and Assignments Files'.
+  $permissions['access Tasks and Assignments Files'] = array(
+    'name' => 'access Tasks and Assignments Files',
+    'roles' => array(
+      'civihr_admin' => 'civihr_admin',
+      'civihr_admin_local' => 'civihr_admin_local',
+      'civihr_staff' => 'civihr_staff',
+    ),
+    'module' => 'civicrm',
+  );
+
   // Exported permission: 'access Vacancy'.
   $permissions['access Vacancy'] = array(
     'name' => 'access Vacancy',
@@ -380,6 +391,7 @@ function civihr_default_permissions_user_default_permissions() {
       'administrator' => 'administrator',
       'civihr_admin' => 'civihr_admin',
       'civihr_admin_local' => 'civihr_admin_local',
+      'civihr_staff' => 'civihr_staff',
     ),
     'module' => 'civicrm',
   );
@@ -2226,17 +2238,17 @@ function civihr_default_permissions_user_default_permissions() {
   );
 
   // Exported permission: 'view my tasks and documents blocks'.
-  $permissions['view my tasks and documents blocks'] = [
+  $permissions['view my tasks and documents blocks'] = array(
     'name' => 'view my tasks and documents blocks',
-    'roles' => [
+    'roles' => array(
       'administrator' => 'administrator',
       'civihr_admin' => 'civihr_admin',
       'civihr_admin_local' => 'civihr_admin_local',
       'civihr_manager' => 'civihr_manager',
       'civihr_staff' => 'civihr_staff',
-    ],
+    ),
     'module' => 'civihr_employee_portal',
-  ];
+  );
 
   // Exported permission: 'view own manual batches'.
   $permissions['view own manual batches'] = array(
--- a/civihr_employee_portal/js/ta-documents-app.js
+++ b/civihr_employee_portal/js/ta-documents-app.js
@@ -24,14 +24,14 @@
           .catch(function (reason) {
             CRM.alert(reason, 'Error', 'error');
           });
-      }
+      };
 
-      (function init ($window) {
+      (function init () {
         // Reloads page on 'document-saved' event
         $rootScope.$on('document-saved', function () {
           $window.location.reload();
         });
-      })($window);
+      })();
 
       /**
        * Opens Document Modal
--- a/civihr_employee_portal/js/ta-documents-app.js
+++ b/civihr_employee_portal/js/ta-documents-app.js
@@ -1,8 +1,8 @@
 /* globals angular */
 
 (function (angular) {
-  angular.module('taDocuments', ['civitasks.appDocuments']).controller('ModalController', ['$scope', '$rootScope', '$rootElement', '$log', '$uibModal', 'DocumentService', 'FileService', 'config',
-    function ($scope, $rootScope, $rootElement, $log, $modal, DocumentService, FileService, config) {
+  angular.module('taDocuments', ['civitasks.appDocuments']).controller('ModalController', ['$scope', '$rootScope', '$window', '$rootElement', '$log', '$uibModal', 'DocumentService', 'FileService', 'config',
+    function ($scope, $rootScope, $window, $rootElement, $log, $modal, DocumentService, FileService, config) {
       var vm = {};
 
       /**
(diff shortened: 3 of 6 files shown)


civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -155,5 +169,29 @@ function filterTable(statusId) {
...
+
+    taInit();
+
+    document.addEventListener('taReady', function(){
+      taInit();
+    });
   }(CRM.$));
 </script>
+
+<script src="<?php echo getModuleUrl('uk.co.compucorp.civicrm.tasksassignments').'js/dist/tasks-assignments.min.js';?>"></script>
```
AkA84: shouldn't you use `civicrm_resources_load` here?

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,6 +569,32 @@ function civihr_employee_portal_init() {
     _rebuild_view('appraisal');
     _rebuild_view('tasks');
     _rebuild_view('documents');
+
+    $extensionPath = CRM_Extension_System::singleton()->getMapper()->keyToUrl('uk.co.compucorp.civicrm.tasksassignments');
```
bmnepali: done!

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,6 +569,32 @@ function civihr_employee_portal_init() {
...
+
+    $allSettings = array();
+    $taSettings = civicrm_api3('TASettings', 'get');
```
bmnepali: done!

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,6 +569,32 @@ function civihr_employee_portal_init() {
...
+    foreach ($taSettings['values'] as $key => $value) {
+      $allSettings[$key] = $value['value'];
+    }
+    if (!empty($allSettings['days_to_create_a_document_clone'])) {
+      $allSettings['days_to_create_a_document_clone'] = (int)$allSettings['days_to_create_a_document_clone'];
```
bmnepali: Typecasting as `(int)` returns `0` for `empty`, `string` with alphanumeric  values.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,6 +569,32 @@ function civihr_employee_portal_init() {
...
+        'settings' => $allSettings,
+        'permissions' => array(
+          'delete_tasks_and_documents' => CRM_Core_Permission::check('delete Tasks and Documents'),
+        ),
+      ),
+      'adminId' => CRM_Core_Session::getLoggedInContactID(),
+      'contactId' => _get_civicrm_contact_id_by_drupal_uid($user->uid),
+      'debug' => $config->debug
+    ];
+    drupal_add_js($settings, 'setting');
```
bmnepali: yes, It attaches the `$settings`  to global `Drupal.settings` variable.

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -17,6 +17,9 @@
  *   field id, then row number. This matches the index in $rows.
  * @ingroup views_templates
  */
+// loading required resources
+civicrm_resources_load('org.civicrm.reqangular', ['reqangular.min.js']);
```
bmnepali: More appropriate to load in `.module` file. Will fix it.

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -34,20 +37,20 @@
   $statusesCount[0] ++;
 endforeach;
 ?>
-<div class="chr_table-w-filters chr_table-w-filters--documents row">
+<div id="tnadocuments" ng-controller="MainCtrl" class="chr_table-w-filters chr_table-w-filters--documents row">
```
bmnepali: done!

bmnepali: Yes, and it is the controller containing the `modalDocument(...)` function, which is responsible to display document modal.

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -95,26 +98,28 @@
...
                   <button
                     class="btn btn-sm btn-default ctools-use-modal ctools-modal-civihr-default-style ctools-use-modal-processed"
                     disabled="disabled">
-                    <i class="fa fa-upload"></i> Upload
+                    <i class="fa fa-upload"></i> Open
+                  </button>
+                <?php else: ?>
+                  <button
+                    ng-click="modalDocument({
+                      id: <?php print strip_tags($row['id']); ?>
```
bmnepali: sure! will do

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -136,10 +150,10 @@ function filterTable(statusId) {
     }
 
     var $tableFilters = $('.chr_table-w-filters--documents'),
-            $filtersDropdown = $tableFilters.find('.chr_table-w-filters__filters__dropdown'),
-            $filtersNav = $tableFilters.find('.chr_table-w-filters__filters__nav'),
-            $tableDocStaff = $tableFilters.find('.chr_table-w-filters__table'),
-            $tableDocStaffRows = $tableDocStaff.find('.document-row');
+      $filtersDropdown = $tableFilters.find('.chr_table-w-filters__filters__dropdown'),
```
bmnepali: got it!

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -155,5 +169,29 @@ function filterTable(statusId) {
     $filtersDropdown.on('change', function (e) {
       filterTable($(this).val());
     });
+
+    // Trigger event to trigger appDocuments forom T&A
```
bmnepali: oops! will fix

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
bmnepali: I can use `civicrm_resources_load(...)` here as order of execution of scripts is important. As `civicrm_resources_load(...)` tries to load the scripts before the inline scripts are loaded, thus the settings variables 
```

    CRM.contactId = Drupal.settings.contactId;
    CRM.adminId = Drupal.settings.adminId;
    CRM.Tasksassignments = Drupal.settings.Tasksassignments;
```
will not be available for `tasks-assignments.min.js` causing following exception:
screen shot 2017-05-29 at 2 02 09 pm

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,6 +569,32 @@ function civihr_employee_portal_init() {
...
+    $allSettings = array();
+    $taSettings = civicrm_api3('TASettings', 'get');
+    foreach ($taSettings['values'] as $key => $value) {
+      $allSettings[$key] = $value['value'];
+    }
+    if (!empty($allSettings['days_to_create_a_document_clone'])) {
+      $allSettings['days_to_create_a_document_clone'] = (int)$allSettings['days_to_create_a_document_clone'];
+    }
+
+    $settings = [
```
bmnepali: These settings are required to bootstrap T&A app, actually, I have taken these settings form T&A in `CRM/Tasksassignments/Page/Tasksassignments.php` file.

civihr_employee_portal/civihr_employee_portal.module
bmnepali: The main purpose to the type casting is to make sure `days_to_create_a_document_clone` always get the integer value.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,6 +569,32 @@ function civihr_employee_portal_init() {
...
+        'extensionPath' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.tasksassignments'),
+        'case_extension' => !empty(CRM_Core_Component::get('CiviCase')),
+        'settings' => $allSettings,
+        'permissions' => array(
+          'delete_tasks_and_documents' => CRM_Core_Permission::check('delete Tasks and Documents'),
+        ),
+      ),
+      'adminId' => CRM_Core_Session::getLoggedInContactID(),
+      'contactId' => _get_civicrm_contact_id_by_drupal_uid($user->uid),
+      'debug' => $config->debug
```
bmnepali: yes it is. And i agree with your suggestion to make it available under `Tasksassignments` array.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,6 +569,32 @@ function civihr_employee_portal_init() {
...
+    $settings = [
+      'Tasksassignments' => array(
+        'extensionPath' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.tasksassignments'),
+        'case_extension' => !empty(CRM_Core_Component::get('CiviCase')),
+        'settings' => $allSettings,
+        'permissions' => array(
+          'delete_tasks_and_documents' => CRM_Core_Permission::check('delete Tasks and Documents'),
+        ),
+      ),
+      'adminId' => CRM_Core_Session::getLoggedInContactID(),
```
bmnepali: I had a little chat with @coldrunKacper . Here `adminId` is actually a currently logged in user contact id. Currently, however it does not check weather it is a admin or not.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,6 +569,32 @@ function civihr_employee_portal_init() {
...
+      'contactId' => _get_civicrm_contact_id_by_drupal_uid($user->uid),
```
bmnepali: Actually, it should be `contactId' => CRM_Utils_Request::retrieve('cid', 'Integer')`. This value is taken form request url. Here `contactId` can be different as per different contacts. So, it is not dependent on currently logged-in contact it.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -549,6 +549,7 @@ function civihr_employee_portal_init() {
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/lib/sweetalert/sweetalert.css");
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/lib/tablesaw/tablesaw.css");
 
+
```
dhavalsharma: empty line

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,7 +570,35 @@ function civihr_employee_portal_init() {
...
+    _load_ta_settings();
+
+    // Page load scripts
+    civicrm_resources_load('org.civicrm.reqangular', ['reqangular.min.js']);
+    _add_script_to_page('/js/ta-documents-app.js', 'dashboard');
+  }
+}
+
+function _load_ta_settings() {
```
dhavalsharma: document function?

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,50 @@
+angular.module('taDocuments', ['civitasks.appDocuments']).controller('myController', ['$scope', '$rootScope', '$rootElement', '$log', '$uibModal', 'DocumentService', 'FileService', 'config',
```
dhavalsharma: can myController be renamed to better name?

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,50 @@
...
+  function($scope, $rootScope, $rootElement, $log, $modal, DocumentService, FileService, config) {
+    $scope.modalDocument = function(data, e) {
+      e && e.preventDefault();
+      var DocumentData = DocumentService.get({
```
dhavalsharma: documentData?

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,50 @@
...
+        'id': data.id
+      });
+
+      DocumentData.then(function(data) {
```
dhavalsharma: why not chain it to .get call?

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Opens Document Modal with or without document data
+     *
+     * @param {object} data
+     */
+    function openModalDocument(data) {
+      var modalInstance = $modal.open({
+        appendTo: $rootElement.find('div').eq(0),
+        templateUrl: config.path.TPL + 'modal/document.html?v=3',
```
dhavalsharma: can path be in some constant or configuration?

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -34,20 +34,20 @@
...
 endforeach;
 ?>
-<div class="chr_table-w-filters chr_table-w-filters--documents row">
+<div data-ta-documents ng-controller="myController" class="chr_table-w-filters chr_table-w-filters--documents row">
   <div class="chr_table-w-filters__filters col-md-3">
     <div class="chr_table-w-filters__filters__dropdown-wrapper">
       <div class="chr_custom-select chr_custom-select--full">
         <select class="chr_table-w-filters__filters__dropdown skip-js-custom-select">
-<?php foreach ($statuses as $key => $value): ?>
+          <?php foreach ($statuses as $key => $value): ?>
```
dhavalsharma: why would ng-repeat not work here?

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -125,6 +125,11 @@ class="btn btn-sm btn-default ctools-use-modal ctools-modal-civihr-default-style
   (function ($) {
     'use strict';
 
+    CRM.contactId = Drupal.settings.contactId;
+    CRM.adminId = Drupal.settings.adminId;
+    CRM.Tasksassignments = Drupal.settings.Tasksassignments;
```
dhavalsharma: taskAssignments?

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
bmnepali: @dhavalsharma `$statuses` is a php array and it's the php template file. i think we can tweak to use `ng-repeat` but i prefer not to mess with the template that is already working fine. At the end we are only attaching angular app to display modal on clicking "Open" button.

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
bmnepali: got it! will fix

civihr_employee_portal/js/ta-documents-app.js
bmnepali: `config.path` is created as  follows:
```
  angular.module('civitasks.config',[]).constant('config', {
        ...
        path: {
            EXT: CRM.Tasksassignments.extensionPath,
            TPL: CRM.Tasksassignments.extensionPath + 'views/'
        },
```
And for views path we are using `templateUrl: config.path.TPL`.

civihr_employee_portal/js/ta-documents-app.js
bmnepali: done!

civihr_employee_portal/civihr_employee_portal.module
bmnepali: done!

civihr_employee_portal/civihr_employee_portal.module
bmnepali: added!

dhavalsharma: `for to` can be replaced by `to` only

civihr_employee_portal/js/ta-documents-app.js
dhavalsharma: please rename `DocumentData` to `documentData`

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -34,7 +34,7 @@
   $statusesCount[0] ++;
 endforeach;
 ?>
-<div data-ta-documents ng-controller="myController" class="chr_table-w-filters chr_table-w-filters--documents row">
+<div data-ta-documents ng-controller="ModalController" class="chr_table-w-filters chr_table-w-filters--documents row">
```
dhavalsharma: there are so many modals, which model's controller is this?

bmnepali: It's newly  created `ModalController` in `civihr_employee_portal/js/ta-documents-app.js`

civihr_employee_portal/js/ta-documents-app.js
bmnepali: Actually, we don't need it anymore! fixed.

civihr_employee_portal/civihr_employee_portal.module
bmnepali: fixed

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,9 +569,40 @@ function civihr_employee_portal_init() {
     _rebuild_view('appraisal');
     _rebuild_view('tasks');
     _rebuild_view('documents');
+
+    _load_ta_settings();
+
+    // Page load scripts
+    civicrm_resources_load('org.civicrm.reqangular', ['reqangular.min.js']);
+    _add_script_to_page('/js/ta-documents-app.js', 'dashboard');
```
davialexandre: please move these two lines adding scripts closer to the existing `_add_script_to_page` calls, before the `_rebuild_view` calls.

The 'Page load scripts' comment isn't adding any useful information. Please remove it.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -569,9 +569,40 @@ function civihr_employee_portal_init() {
...
+ */
+function _load_ta_settings() {
+  $taSettings = civicrm_api3('TASettings', 'get')['values'];
+
+  if (!empty($taSettings['days_to_create_a_document_clone'])) {
+    $taSettings['days_to_create_a_document_clone'] = (int)$taSettings['days_to_create_a_document_clone'];
+  }
+
+  drupal_add_js([
+    'tasksAssignments' => array(
```
davialexandre: Always use the PHP's short array syntax. Replace the array() calls with []

bmnepali: got it! will fix

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -562,16 +562,46 @@ function civihr_employee_portal_init() {
...
+
+  drupal_add_js([
+    'tasksAssignments' => [
+      'extensionPath' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.tasksassignments'),
+      'case_extension' => !empty(CRM_Core_Component::get('CiviCase')),
+      'settings' => $taSettings,
+      'permissions' => [
+        'delete_tasks_and_documents' => CRM_Core_Permission::check('delete Tasks and Documents'),
+      ],
+      'debug' => $config->debug
```
davialexandre: You're using the `$config` object here but you're not instantiating it anywhere. I assume you based yourself on the code of some other extension to write this, right? So check in that code where and how the `$config` object is instantiated.

civihr_employee_portal/features/civihr_default_permissions/civihr_default_permissions.features.user_permission.inc
```diff
@@ -99,6 +99,17 @@ function civihr_default_permissions_user_default_permissions() {
     'module' => 'civicrm',
   );
 
+  // Exported permission: 'access Tasks and Assignments Files'.
+  $permissions['access Tasks and Assignments Files'] = array(
```
mickadoo: Ah, there it is :-)

civihr_employee_portal/civihr_employee_portal.module
bmnepali: Thanks! for catching, fixed it.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -562,16 +562,47 @@ function civihr_employee_portal_init() {
       '/lib/sweetalert/sweetalert.min.js'
     ]);
 
+    civicrm_resources_load('org.civicrm.reqangular', ['reqangular.min.js']);
```
AkA84: I don't think this can stay like this. You are adding `reqangular.min.js` everywhere. You should add it only in the dashboard

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -562,16 +562,47 @@ function civihr_employee_portal_init() {
...
+      'extensionPath' => CRM_Core_Resources::singleton()->getUrl('uk.co.compucorp.civicrm.tasksassignments'),
+      'case_extension' => !empty(CRM_Core_Component::get('CiviCase')),
+      'settings' => $taSettings,
+      'permissions' => [
+        'delete_tasks_and_documents' => CRM_Core_Permission::check('delete Tasks and Documents'),
+      ],
+      'debug' => $config->debug
+    ],
+    'adminId' => CRM_Core_Session::getLoggedInContactID(),
+    'contactId' => CRM_Utils_Request::retrieve('cid', 'Integer')
```
AkA84: Can you explain to me what you need `adminId` and `contactId` for?  I saw your explanations for both of them but it still doesn't make much sense to me

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,52 @@
+angular.module('taDocuments', ['civitasks.appDocuments']).controller('ModalController', ['$scope', '$rootScope', '$rootElement', '$log', '$uibModal', 'DocumentService', 'FileService', 'config',
```
AkA84: wrap everything in an IIFE
```js
(function (angular) {
  angular.module()

}(angular))

File: civihr_employee_portal/js/ta-documents-app.js [hunk 31]
```diff
@@ -0,0 +1,52 @@
...
+  function ($scope, $rootScope, $rootElement, $log, $modal, DocumentService, FileService, config) {
+    this.modalDocument = function (id, role) {
```
AkA84: you need to document any method/function
```js
/**
 *
 *
 */
 this.moduleDocument
```

AkA84: do not use `this`, use `vm`
```js
.controller('MyController', function () {
  var vm = {};

  vm.moduleDocument = function () {
  }

  // ...

  return vm;
}

File: civihr_employee_portal/js/ta-documents-app.js [hunk 32]
```diff
@@ -0,0 +1,52 @@
...
+
+      DocumentService.get({'id': id})
```
AkA84: did this pass the linter? I think it should be
`{ 'id': id }`

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,52 @@
...
+      .then(function (data) {
```
AkA84: wrong indentation
```js
DocumentService.get()
  .then(function () {

  }
```

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,52 @@
...
+        // Display the modal
```
AkA84: you don't need this, it's self explanatory

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,52 @@
...
+        !!data && openModalDocument(data[0], role);
```
AkA84: or else?
```js
Document.get()
  .then(function (data) {
     if (!data) {
       throw new Error('lorem ipsum dolor'); // this should be caught by the `catch()` below
     }
  })
  .catch(function (reason) {
    CRM.alert(reason, 'Error', 'error');
  });

File: civihr_employee_portal/js/ta-documents-app.js [hunk 36]
```diff
@@ -0,0 +1,52 @@
...
+      }, function (reason) {
+        CRM.alert(reason, 'Error', 'error');
+      });
+    }
+
+    /**
+     * Opens Document Modal with or without document data
```
AkA84: how can it be "without" data?

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,52 @@
...
+     *
+     * @param {object} data
+     */
+    function openModalDocument (data, role) {
+      var modalInstance = $modal.open({
+        appendTo: $rootElement.find('div').eq(0),
+        templateUrl: config.path.TPL + 'modal/document.html?v=3',
```
AkA84: v=3 ?

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,52 @@
...
+        controller: 'ModalDocumentCtrl',
+        resolve: {
+          role: function () {
+            return [role];
```
AkA84: again, why an array? Also, does this work 
```js
resolve: {
  role: role
}
```
?

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,52 @@
...
+              return [];
+            }
+
+            return FileService.get(data.id, 'civicrm_activity');
+          }
+        }
+      });
+
+      modalInstance.result.then(function (results) {
+        $scope.$broadcast('documentFormSuccess', results, data);
```
AkA84: what is this for?

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,52 @@
...
+
+        if (results.open) {
+          $rootScope.modalDocument(data);
```
AkA84: what does this do?

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -155,5 +160,13 @@ function filterTable(statusId) {
     $filtersDropdown.on('change', function (e) {
       filterTable($(this).val());
     });
+
+    // Listen for ready event when T&A finishes loading all modules
+    document.addEventListener('taReady', function (e) {
```
AkA84: who is sending the `taReady` event in this case? What's the initialization flow here (= sequence of events)?

civihr_employee_portal/js/ta-documents-app.js
bmnepali: will do

civihr_employee_portal/js/ta-documents-app.js
bmnepali: I already tried this, if I do so, in ModalDocumentController, it looks for 'staffs' service giving service not found exception so I am using array.

civihr_employee_portal/js/ta-documents-app.js
bmnepali: A query string, taken form T&A, this is to get to the latest changes to `modal/document.html` .

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
bmnepali: `taReady` event is sent by the following code in `js/src/tasks-assignments/app.js` in T&A
```js
define([
  ... all required modules
], function () {
  'use strict';
  ...
  document.dispatchEvent(typeof window.CustomEvent === 'function' ? new CustomEvent('taReady') : (function () {
    var e = document.createEvent('Event');

    e.initEvent('taReady', true, true);

    return e;
  })());
});
```
It triggers the `taReady` when all modules are loaded.

civihr_employee_portal/js/ta-documents-app.js
bmnepali: The comment needs to be fixed.

civihr_employee_portal/civihr_employee_portal.module
bmnepali: Well, as per my understanding about `adminId` and `contactId` in T&A, 
1. `adminId` is simply contact id of currently logged in user (will be admin in case of T&A)
2. `contactId` is retrieved form url. i.e this is cid in url like `/civicrm/contact/view?reset=1&cid=30`. This is useful if admin is logged in and want to access the some other contact tasks and documents relating to T&A.

So, to conclude, in T&A sending the contactId is appropriate. But in SSP indivisula user is only dealing  with their respective documents, so `contactId` is not necessary.

civihr_employee_portal/js/ta-documents-app.js
bmnepali: will check, again

civihr_employee_portal/js/ta-documents-app.js
bmnepali: done!

civihr_employee_portal/js/ta-documents-app.js
bmnepali: I have taken the above code form T&A
```js
modalInstance.result.then(function (results) {
   ...
})
```
I realized, this is only needed in T&A, as there needs to identify the status of the document and should be removed form the list of documents if the document is approved or rejected and keep otherwise. And `$rootScope.modalDocument(data);` is used to display the modal  if the button "Save and New" was clicked. But now that the button is removed, we wont be needing now.

civihr_employee_portal/civihr_employee_portal.module
bmnepali: fixed it!

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -562,14 +582,49 @@ function civihr_employee_portal_init() {
       '/lib/sweetalert/sweetalert.min.js'
     ]);
 
+    _add_module_script_to_page([
+      'org.civicrm.reqangular' => [
+        'reqangular.min.js'
+        ]
+      ], 'dashboard');
+
+    _add_script_to_page('/js/ta-documents-app.js', 'dashboard');
```
AkA84: you already have 
```php
_add_script_to_page('/js/tasks.js', 'dashboard');
```
so you can simply amend it to
```php
_add_script_to_page(['/js/tasks.js', '/js/ta-documents-app.js'], 'dashboard');

File: civihr_employee_portal/civihr_employee_portal.module [hunk 43]
```diff
@@ -562,14 +582,49 @@ function civihr_employee_portal_init() {
       '/lib/sweetalert/sweetalert.min.js'
     ]);
 
+    _add_module_script_to_page([
```
AkA84: you should reuse the already existing `_add_script_to_page`. Just change the signature
```php
function _add_script_to_page($resources, $extension = NULL, $page = NULL) {

}

// ...

_add_script_to_page('reqangular.min.js', 'org.civicrm.reqangular', 'dashboard');
_add_script_to_page(['/js/tasks.js', '/js/ta-documents-app.js'], NULL, 'dashboard');
```
might not be the most elegant solution, but it's better define a whole new function

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,61 @@
...
+       * @param {string} role
+       */
+      vm.modalDocument = function (id, role) {
+        DocumentService.get({ id: id })
+          .then(function (data) {
+             if (!data) {
+               throw new Error('Requested Document is not available');
+             }
+
+             !!data && openModalDocument(data[0], role);
```
AkA84: there is no point checking for `!!data` anymore. If `data` is not present, you are throwing an exception

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,61 @@
...
+       * @param {string} role
+       */
+      function openModalDocument (data, role) {
+        var modalInstance = $modal.open({
+          appendTo: $rootElement.find('div').eq(0),
+          templateUrl: config.path.TPL + 'modal/document.html?v=3',
+          controller: 'ModalDocumentCtrl',
+          resolve: {
+            role: function () {
+              return [role];
```
AkA84: remove the array

civihr_employee_portal/js/ta-documents-app.js
bmnepali: This is only needed in T&A, once the document is updated there needs to identify the status of the document and should be removed form the list of documents if the document is approved or rejected and keep otherwise.  So, `$scope.$broadcast('documentFormSuccess', results, data);` code is simply broadcasting `documentFormSuccess` to its listener which will be updating the list of documents based on the status.

civihr_employee_portal/js/ta-documents-app.js
bmnepali: will do

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -26,6 +26,11 @@
           });
       }
 
+      // Reloads page on 'document-saved' event
+      $rootScope.$on('document-saved', function () {
```
AkA84: wrap this in a `init()` IIFE

```js
(function init() {
  // ...
}())
```

bmnepali: Ahh! right! will do

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -5,6 +5,13 @@
     function ($scope, $rootScope, $window, $rootElement, $log, $modal, DocumentService, FileService, config) {
       var vm = {};
 
+      (function init ($window) {
```
AkA84: put it in the correct order
```js
// public properties 
vm.property1 = 'foo';
vm.property2 = 'bar';

// public methods
vm.method1 = function () {}
vm.method2 = function () {}

(function init() {

}());

// private methods
function private1() {

}

File: civihr_employee_portal/js/ta-documents-app.js [hunk 48]
```diff
@@ -26,12 +26,12 @@
           });
       }
 
-      (function init ($window) {
+      ;(function init () {
```
bmnepali: Adding `;` to fix the issue. As per the solution described here: https://stackoverflow.com/questions/42036349/uncaught-typeerror-intermediate-value-is-not-a-function

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -26,7 +26,7 @@
           });
       }
 
-      (function init ($window) {
+      ;(function init ($window) {
```
bmnepali: Adding `;` to fix the issue. As per the solution described here: https://stackoverflow.com/questions/42036349/uncaught-typeerror-intermediate-value-is-not-a-function

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -0,0 +1,68 @@
...
+             }
+
+             openModalDocument(data[0], role);
+          })
+          .catch(function (reason) {
+            CRM.alert(reason, 'Error', 'error');
+          });
+      }
+
+      ;(function init ($window) {
```
AkA84: why is this semicolon here? the problem is that the semicolon was missing from here
```js
vm.modalDocument = function (id, role) {

}; // <-- missing semicolon

bmnepali: ohh! will fix.

PR #314, opened 2017-06-21
PCHR-2318: Buttons change and console error when modal is open on ssp

bmnepali:
## Overview
During the Document Updates in SSP, angular app was created and injected in SSP which added additional `#/` in the url causing the following error on opening modals.

#### Error:
screen shot 2017-06-21 at 7 33 48 pm

## Before
2318-before-min

## After
2318_after-min

## Technical Details
1. Added following code in `civihr_employee_portal/js/ta-documents-app.js`:
```js
(function(angular) {
  angular.module('taDocuments', ['civitasks.appDocuments', 'civitasks.directives'])
    .config(function ($urlRouterProvider, $locationProvider) {
      $locationProvider.html5Mode(true); // This is required to remove # for the URL
      $urlRouterProvider.otherwise('/dashboard');
    })
    ...
})(angular);
```
This code will be removing the `#/` form the url and add a fallback url to be `/dashboard`.

2. Add `<base href="/"> <!-- This is required to remove # for the URL-->` in file `civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php` which applies the base path to be `/`.

## Comments:
1. Previously the **only the first day of the request is shown in case multiple days are selected in a future** 
Check issue here: https://compucorp.atlassian.net/secure/attachment/41625/41625_2317+dates+selected.mp4
This  issue is resolved as of this PR.Now one can select multiple days in Request leave modal. To see it in action please see gif in **After** section.

### DIFF
--- a/civihr_employee_portal/js/ta-documents-app.js
+++ b/civihr_employee_portal/js/ta-documents-app.js
@@ -2,6 +2,10 @@
 
 (function(angular) {
   angular.module('taDocuments', ['civitasks.appDocuments', 'civitasks.directives'])
+    .config(function ($urlRouterProvider, $locationProvider) {
+      $locationProvider.html5Mode(true); // This is required to remove # for the URL
+      $urlRouterProvider.otherwise('/dashboard');
+    })
     .controller('ModalController', ['$scope', '$rootScope', '$window', '$rootElement', '$log', '$uibModal',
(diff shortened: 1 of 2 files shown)


bmnepali:
@igorpavlov Updated comment section of PR to clarify your comment.

PR #312, opened 2017-06-14
PCHR-2317: Attatch document modal to the root of angular app

bmnepali:
## Overview
This PR is for the hot fix to add name space to temporary styles for the document modal in ssp for staff and manager document sections.

## Before
Previously, modal in request leave was broken ad follows:
screen shot 2017-06-14 at 12 52 49 pm

## After
After name-spacing the temporary styles, the leave request modal looks ok
screen shot 2017-06-14 at 12 48 23 pm

## Technical Details
1. In file `civihr-custom/civihr_employee_portal/js/ta-documents-app.js` change the modal option to attach the modal in the root element as follows:
```js
function openModalDocument(data, role) {
   var modalInstance = $modal.open({
       **appendTo: $rootElement,**
       ...
   });
}
```
2. Name-spacing the temporary styles for document modal in ssp in file `civihr_employee_portal_theme/civihr_default_theme/scss/base/_task-assignments-modal-temp.scss` 
```css
[data-ta-documents],
[data-ta-documents-manager] {
   // All temporary styles for documents modal
}
```
### Updates:
1. When tested in test site 1, sometimes the contacts gets fetched earlier and sometimes take longer. Due to which the target contact in Document modal is available and sometimes not. So to mitigate this issue, function to cache contacts in `Document Service` is passed to modal resolve object, so the modal will be displayed when the contacts gets fetched and cached in SPP.

Update `openModalDocument` function to fetch and cache all contacts before displaying the modal in file `civihr-custom/civihr_employee_portal/js/ta-documents-app.js`.

```js
isContactsCached = DocumentService.cacheContactsAndAssignments(documents, 'contacts');
```

```js
function openModalDocument(data, role) {
  var modalInstance = $modal.open({
    ...
    resolve: {
      ...
      isContactsCached: isContactsCached,
      ...
    }
  });
};
```
Added spinner before loading the Modal.
finalwithspinner

2. In SSP currently documents with approved and rejected status are not fetched as in T&A. But in SSP, including others both approved and rejected documents' contacts needs to be cached. So, removing restriction to fetch documents other than approved(3) and rejected(4) status. Where 3 and 4 are status id respectively.
Change form:
```js
DocumentService.get({
   'status_id': {
       'NOT IN': config.status.resolve.DOCUMENT
   }
}).then(function (documents) {
   ...
});
```
 to
```js
DocumentService.get().then(function (documents) {
   ...
});

- [ ] Tests Pass
No tests added for js changes in ssp.

PR #319, opened 2017-07-03
PCHR-2355: Configure DATE_FORMAT in HR_settings for SSP

bmnepali:
## Overview
Currently, when trying to save the documents form Documents Widget in SSP. A error  is thrown because date format is not configured for angular app in ssp for documents modal.

When a `civihr_manager` changes the status form the documents widget in ssp, the status does not get updated and displays the access denied 403 error. Which was due to the permission issue.

## Before
1. Cannot save document by the civihr_manager in ssp
2355_1_before

2. Changing status does not get saved in documents manager in ssp
2355_before

## After
1. 'civihr_manager` can save edited document form ssp
2355_1_after

2. `civihr_manager` can changed the document status and it gets saved
2355_after

## Technical Details
1. Configure date format for SSP in mini angular app for the Document modals.
```js
.controller('ModalController', ['$scope', ..., 'settings', 'DateFormat',
      function($scope, ..., settings, DateFormat) {
        ...
       (function init() {
          ...
          // Get list of documents
          DocumentService.get().then(function (documents) {
            //sets the date format for HR_settings.DATE_FORMAT
            DateFormat.getDateFormat();
          });
        ...
    })();
});
```
2. Give "change document status" permission for "civihr_manager" user in file "features/civihr_default_permissions/civihr_default_permissions.features.user_permission.inc"
```php
  // Exported permission: 'change document status'.
  $permissions['change document status'] = array(
    'name' => 'change document status',
    'roles' => array(
      ...
      'civihr_manager' => 'civihr_manager',
    ),
    'module' => 'civihr_employee_portal',
  );
```

### Reopened Issue:
When the page in ssp is refreshed and before the page gets scrolled back to top, if clicked the open button in a document on Document Manager in ssp, the modal opens with empty target contact. Fetching and caching of contacts for a document is improved in  PR #312 , and this even needs to be polished to to fix the reopened issue taking following steps

1. Watch for the change in the contacts search array.
```js
$rootScope.$watch('cache.contact', function () {
   ..//other operations
});
```
2. For when there is a change of values in the contacts array, check if the array contains more than one contacts. Because one contact (currently logged in user data) will already be available. So, if there are documents there should be more than one contact including a contacts associated with the document.
```js
$rootScope.$watch('cache.contact', function () {
   contactsLoaded = $rootScope.cache.contact.arrSearch.length > 1;
   if (!contactsLoaded) {
      ... // Display spinner and hide "open" button
   } else {
      ... // Hide spinner and display "open" button
   }
});
```
3. Add a code to display and hide the spinner and "Open " button in documents, in respective place above:
```js
if (!contactsLoaded) {
   $rootScope.$broadcast('ct-spinner-show');
   vm.loadingModalData = true;
} else {
   $rootScope.$broadcast('ct-spinner-hide');
   vm.loadingModalData = false;
}
```
4. We won't be needing to check `isContactsCached` if contact data is resolved as the contacts data will be available when the controller is instantiated in `init()` call.
```js
...
isContactsCached = DocumentService.cacheContactsAndAssignments(documents, 'contacts');
...

...
isContactsCached: isContactsCached,
...
```

### Before:
Displays the open button, before the contacts get fully loaded, and opening the modal displayes modal with empty target contact.
2355-before-iloveimg-compressed

### After:
Displays the spinner and hides the "open" button  until the data gets loaded.
fasdfasdf-iloveimg-compressed

### DIFF
--- a/civihr_employee_portal/js/ta-documents-app.js
+++ b/civihr_employee_portal/js/ta-documents-app.js
@@ -7,13 +7,12 @@
       $urlRouterProvider.otherwise('/dashboard');
     })
     .controller('ModalController', ['$scope', '$rootScope', '$window', '$rootElement', '$log', '$uibModal',
-      'DocumentService', 'FileService', 'config', 'settings', 'HR_settings',
+      'DocumentService', 'FileService', 'config', 'settings', 'DateFormat',
       function($scope, $rootScope, $window, $rootElement, $log, $modal, DocumentService, FileService, config,
-        settings, HR_settings) {
+        settings, DateFormat) {
         var vm = {};
         var isContactsCached = {};
 
-        HR_settings.DATE_FORMAT = "dd/MM/yyyy";
         vm.loadingModalData = false;
 
         /**
@@ -41,12 +40,15 @@
         };
 
         (function init() {
+
           // Reloads page on 'document-saved' event
           $rootScope.$on('document-saved', function () {
             $window.location.reload();
           });
           // Get list of documents
           DocumentService.get().then(function (documents) {
+            //sets the date format for HR_settings.DATE_FORMAT
+            DateFormat.getDateFormat();
             isContactsCached = DocumentService.cacheContactsAndAssignments(documents, 'contacts');
           });
         })();
--- a/civihr_employee_portal/js/ta-documents-app.js
+++ b/civihr_employee_portal/js/ta-documents-app.js
@@ -11,7 +11,7 @@
       function($scope, $rootScope, $window, $rootElement, $log, $modal, DocumentService, FileService, config,
         settings, DateFormat) {
         var vm = {};
-        var isContactsCached = {};
+        var availableContacts = false;
 
         vm.loadingModalData = false;
 
@@ -49,7 +49,7 @@
           DocumentService.get().then(function (documents) {
             //sets the date format for HR_settings.DATE_FORMAT
             DateFormat.getDateFormat();
-            isContactsCached = DocumentService.cacheContactsAndAssignments(documents, 'contacts');
+            DocumentService.cacheContactsAndAssignments(documents, 'contacts');
           });
         })();
 
@@ -74,7 +74,6 @@
               data: function () {
                 return data;
               },
-              isContactsCached: isContactsCached,
               files: function () {
                 if (!data.id || !+data.file_count) {
                   return [];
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -7,11 +7,13 @@
...
     .controller('ModalController', ['$scope', '$rootScope', '$window', '$rootElement', '$log', '$uibModal',
-      'DocumentService', 'FileService', 'config', 'settings',
-      function($scope, $rootScope, $window, $rootElement, $log, $modal, DocumentService, FileService, config, settings) {
+      'DocumentService', 'FileService', 'config', 'settings', 'HR_settings',
+      function($scope, $rootScope, $window, $rootElement, $log, $modal, DocumentService, FileService, config,
+        settings, HR_settings) {
         var vm = {};
         var isContactsCached = {};
 
+        HR_settings.DATE_FORMAT = "dd/MM/yyyy";
```
davialexandre: two things:
- It seems a bit weird to change the settings inside of a controller. It kind of kills the purpose of having a "global" settings var, right? Do we really need to change it in here?
- Please use single quotes to keep consistency with the rest of the code in this file

bmnepali: > It seems a bit weird to change the settings inside of a controller. It kind of kills the purpose of having a "global" settings var, right? Do we really need to change it in here?

Here the `HR_settings` factory comes form the `org.civicrm.reqangular` (civihr_employee_portal/js/ta-documents-app.js) extension. It is set to `null` by default ad following.
```js
define([
    'common/modules/services'
], function (module) {
    'use strict';

    module.factory('HR_settings', function () {
        return {
            DATE_FORMAT: null
        };
    });
});
```
So, I prefer to set the date format here in controller.

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -91,6 +86,25 @@
...
+         * Watch for the changes in the list of $rootScope.cache.contact.arrSearch
+         * Display spinner and hide "open" button until the arrSearch is filled with contacts
+         *
+         * Note: $rootScope.cache.contact.arrSearch will always contain a
+         * contact data (curently logged in contact). So if there are documents,
+         * there must be more that one contacts conidering at aleast a target contact in a document
+         */
+        $rootScope.$watch('cache.contact', function () {
+          availableContacts = $rootScope.cache.contact.arrSearch.length > 1;
+          if (!availableContacts) {
```
igorpavlov:
```js
$rootScope.$broadcast('ct-spinner-' + (availableContacts ? 'hide' : 'show'));
vm.loadingModalData = !availableContacts;
```

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -39,13 +40,16 @@
         };
 
         (function init() {
+
```
igorpavlov: Is this line needed?

bmnepali: I think we need a empty line before comment.

igorpavlov: OK!

PR #324, opened 2017-07-12
PCHR-2358: Add user guide link to be used in account menu

bmnepali:
## Overview
A new menu under account is to be added calling "User Guide" in SSP.

## Before
Not added User guide Menu
screen shot 2017-07-12 at 7 59 46 pm

## After
Added User guide Menu
screen shot 2017-07-12 at 7 33 47 pm

## Technical Details
1. 
```php
/**
 * Implement hook_preprocess_page
 * @param $vars
 */
function civihr_employee_portal_preprocess_page(&$vars) {
    ...
    $user_guide_link = '';
    $image_url = $base_url . '/' . drupal_get_path('module', 'civihr_employee_portal') . '/images/profile-default.png';

    if (isset($_SESSION['CiviCRM']) && $user->uid) {
        ...
        $user_guide_link = l('<i class="fa fa-book"></i>' . t('User guide'), 'http://civihr-documentation.readthedocs.io/en/latest/self-service-portal/login-screen/', array('html' => true));
    ...
    $vars['user_guide_link'] = $user_guide_link;
}
```

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -6768,14 +6768,12 @@ function civihr_employee_portal_preprocess_page(&$vars) {
     $display_name = t('Anonymus');
     $edit_account_link = '';
     $logout_link = '';
-    $user_guide_link = '';
+    $user_guide_link = l('<i class="fa fa-book"></i>' . t('User guide'), 'http://civihr-documentation.readthedocs.io/en/latest/self-service-portal/login-screen/', array('html' => true));
     $image_url = $base_url . '/' . drupal_get_path('module', 'civihr_employee_portal') . '/images/profile-default.png';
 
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6768,11 +6768,14 @@ function civihr_employee_portal_preprocess_page(&$vars) {
     $display_name = t('Anonymus');
     $edit_account_link = '';
     $logout_link = '';
+    $user_guide_link = '';
```
mickadoo: I don't think the $user_guide_link depends on whether the session is set or not. It should always be the same, why not just declare it here?

bmnepali: Sounds reasonable!

PR #325, opened 2017-07-12
PCHR-245: Decode HTML Tags in Vacancy Description

MiyaNoctem:
## Overview
Vacancies list on SSP was showing HTML tags on its description, after clicking on "More details".
hr_vacancies___hrdemo

## Before
Vacancy list view was showing HTML tags in every vacancy's description. This happened because HTML special characters like '<' and '>' were being changed for HTML entities '&lt;' and '&gt;' respectively, breaking the HTML.

## After
Fixed by implementing the _views_post_render hook, replacing the HTML entities for their corresponding characters.

image

### DIFF
--- a/civihr_employee_portal/views/civihr_employee_portal.views.inc
+++ b/civihr_employee_portal/views/civihr_employee_portal.views.inc
@@ -1107,8 +1107,8 @@ function civihr_employee_portal_views_post_render(&$view, &$output, &$cache) {
  * Decodes html entities in description for each vacancy so HTML tags are sent
  * to the browser.
  *
- * @param $view
- * @param $output
+ * @param object $view
+ * @param string $output
  */
 function _civihr_employee_portal_decode_vacancy_description_html(&$view, &$output) {
   $replacements = [];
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1097,6 +1097,27 @@ function civihr_employee_portal_views_post_render(&$view, &$output, &$cache) {
...
+  if ($view->name == 'hr_vacancies') {
+    _civihr_employee_portal_decode_vacancy_description_html($view, $output);
+  }
+}
+
+/**
+ * Decodes html entities in description for each vacancy so HTML tags are sent
+ * to the browser.
+ *
+ * @param $view
```
tunbola: Missing param type

civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1097,6 +1097,27 @@ function civihr_employee_portal_views_post_render(&$view, &$output, &$cache) {
...
+ * @param $output
```
tunbola: Missing param type

PR #328, opened 2017-07-20
PCHR-2417: Add table class

deb1990:
## Overview
As part of this PR `table` class is added to the `age-group` table.

## Technical Details
The `age-group` sections table did not have `table` class and because of that the bootstrap styles were not getting applied. Added the same.

### DIFF
--- a/civihr_employee_portal/templates/civihr-employee-portal-civihr-report-settings-age-group.tpl.php
+++ b/civihr_employee_portal/templates/civihr-employee-portal-civihr-report-settings-age-group.tpl.php
@@ -6,7 +6,7 @@
         <span class="chr_search-result__total__count"><?php echo count($rows); ?></span>
       </div>
     </div>
-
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/templates/civihr-employee-portal-civihr-report-settings-age-group.tpl.php
```diff
@@ -6,7 +6,7 @@
         <span class="chr_search-result__total__count"><?php echo count($rows); ?></span>
       </div>
     </div>
-
+    <?php $classes .= ' table '; ?>
```
AkA84: leave a comment explaining why you are doing this. Also, isn't there a better place to do this other than a template?

deb1990: I discussed with @coldrunKacper and it is not a problem to addd it here because it is a custom template.

PR #330, opened 2017-07-28
PCHR-2449: Add Sickness reason to Leave And Absence Report

tunbola:
## Overview
Currently the Leave and Absence report on the SSP does not have sickness reason as one of its columns. This PR adds the Sickness reason column and this also makes the sickness reason available in the Pivot table.

## After
civihr custom report - l_and_a 2017-07-31 11-31-06

---

- [ ] Tests Pass
Didn't run the tests in employee portal module.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -334,6 +334,7 @@ function _rebuild_absence_activity_view() {
             SELECT lr.contact_id AS absence_contact_id,
               lr.id AS absence_activity_id,
               at.title as absence_type,
+              lr.sickness_reason as sickness_reason,
               lrd.date AS absence_date,
               DATE_FORMAT(DATE(lrd.date), '%Y-%m') AS absence_month,
               lr.from_date AS absence_start_date,
@@ -775,6 +776,45 @@ function get_civihr_absence_statuses($status_id = NULL) {
     $status_id_value = CRM_Utils_Array::value($status_id, $absenceStatuses);
 
     return $status_id_value;
+}
+
+/**
+ * This function returns the sickness reason label for the sickness reason
+ * option value passed to it.
+ * The sickness reason option group is cached to avoid hitting the Db on
+ * every call.
+ *
+ * @param int $sickness_reason_value
+ *
+ * @return string
+ */
+function get_civihr_absence_sickness_reason_label($sickness_reason_value) {
+    $sicknessReasons = &drupal_static(__FUNCTION__);
+
+    if (!isset($sicknessReasons)) {
+        if ($cache = cache_get('civihr_absence_sickness_reasons')) {
+            $sicknessReasons = $cache->data;
+        }
+        else {
+            try {
+                // Civi init
+                civicrm_initialize();
+                $sicknessReasons = CRM_Core_OptionGroup::values('hrleaveandabsences_sickness_reason', FALSE, FALSE, FALSE, NULL, 'label');
+
+                watchdog('DB hit absence sickness reasons', 'DB');
+            }
+
+            catch (CiviCRM_API3_Exception $e) {
+                $error = $e->getMessage();
+            }
+
+            // Cache the absence sickness reasons for 5 minutes
+            cache_set('civihr_absence_sickness_reasons', $sicknessReasons, 'cache', time() + 360);
+        }
+    }
+    $sickness_reason_label = CRM_Utils_Array::value($sickness_reason_value, $sicknessReasons);
+
+    return $sickness_reason_label;
 }
 
 function get_document_statuses($status_id = NULL) {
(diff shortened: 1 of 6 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -777,6 +778,45 @@ function get_civihr_absence_statuses($status_id = NULL) {
...
+            }
+
+            catch (CiviCRM_API3_Exception $e) {
+                $error = $e->getMessage();
+            }
+
+            // Cache the absence sickness reasons for 5 minutes
+            cache_set('civihr_absence_sickness_reasons', $sicknessReasons, 'cache', time() + 360);
+        }
+    }
```
omarabuhussein: can you add a space after this line to make the code a bit more clear

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -777,6 +778,45 @@ function get_civihr_absence_statuses($status_id = NULL) {
...
+    $sickness_reason_label = CRM_Utils_Array::value($sickness_reason_value, $sicknessReasons);
```
omarabuhussein: just noticed !! why using civicrm utility class for drupal module, I can understand if it is something related to civicrm and can only be done with civicrm methods and classes but not in this case, you can replace it with and if(!empty(..etc).

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -777,6 +778,45 @@ function get_civihr_absence_statuses($status_id = NULL) {
...
+function get_civihr_absence_sickness_reason_label($sickness_reason_value) {
+    $sicknessReasons = &drupal_static(__FUNCTION__);
+
+    if (!isset($sicknessReasons)) {
+        if ($cache = cache_get('civihr_absence_sickness_reasons')) {
+            $sicknessReasons = $cache->data;
+        }
+        else {
+            try {
+                // Civi init
```
omarabuhussein: no need for this comment.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -777,6 +778,45 @@ function get_civihr_absence_statuses($status_id = NULL) {
...
+                civicrm_initialize();
+                $sicknessReasons = CRM_Core_OptionGroup::values('hrleaveandabsences_sickness_reason', FALSE, FALSE, FALSE, NULL, 'label');
```
omarabuhussein: can you replace this with API call, it is more guaranteed to work in future civicrm versions.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -777,6 +778,45 @@ function get_civihr_absence_statuses($status_id = NULL) {
...
+
+                watchdog('DB hit absence sickness reasons', 'DB');
+            }
+
+            catch (CiviCRM_API3_Exception $e) {
+                $error = $e->getMessage();
```
omarabuhussein: you are not handling the error so why catching it? also I don't feel it really needs to catch anything here so just getting the results from civicrm API and checking for count >0 would be enough I think. At the end sickness reason is not something vital for the user to see.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -777,6 +778,45 @@ function get_civihr_absence_statuses($status_id = NULL) {
...
+            }
+
+            // Cache the absence sickness reasons for 5 minutes
+            cache_set('civihr_absence_sickness_reasons', $sicknessReasons, 'cache', time() + 360);
```
omarabuhussein: 360 is a magic number, always try to avoid magic numbers in code, it should be defined as constant right before calling cache_set() 

```php
 const secondsInFiveMinutes = 300; 
 cache_set('civihr_absence_sickness_reasons', $sicknessReasons, 'cache', time() + secondsInFiveMinutes);
```

also hence 5 minutes are 300, not 360 seconds.

also since 'civihr_absence_sickness_reasons' cache key is used twice, and because magic strings are not usually that different from magic numbers, you can also define it as constant at the top of the function and use it instead.

```php
const sicknessReasonsCacheKey = 'civihr_absence_sickness_reasons';
```

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -777,6 +778,45 @@ function get_civihr_absence_statuses($status_id = NULL) {
...
+        }
+    }
+    $sickness_reason_label = CRM_Utils_Array::value($sickness_reason_value, $sicknessReasons);
+
+    return $sickness_reason_label;
+}
```
omarabuhussein: so basically, I would replace this function with something like this : 

```php
function get_civihr_absence_sickness_reason_label($sicknessReasonValue) {
  $sicknessReasons = &drupal_static(__FUNCTION__);
	const sicknessReasonsCacheKey = 'civihr_absence_sickness_reasons';
	
    if (!isset($sicknessReasons)) {
	    $cachedsicknessReasons = cache_get(sicknessReasonsCacheKey);
        if ($cachedsicknessReasons) {
            $sicknessReasons = $cachedsicknessReasons->data;
        }
        else {
		    civicrm_initialize();
            $sicknessReasons = ;//Replace with API call and check for count > 0 otherwise default to empty array;
			
            // Cache the absence sickness reasons for 5 minutes
			const secondsInFiveMinutes = 300; 
            cache_set(sicknessReasonsCacheKey, $sicknessReasons, 'cache', time() + secondsInFiveMinutes);
        }
    }
	
	$sicknessReasonLabel = '';
	if (!empty($sicknessReasons[$sicknessReasonValue]) {
	    $sicknessReasonLabel = $sicknessReasons[$sicknessReasonValue];
	}

    return $sicknessReasonLabel;
}
```

PR #336, opened 2017-08-10
PCHR-2489: Repurpose Document View

mickadoo:
## Overview

There are a number of changes to the appearance and behaviour of the documents widget:

- The filter on the documents view will only show documents assigned to the current user
- The following columns will be shown:
  - ID: (hidden) required for modal click trigger
  - Staff: the person this document belongs to (target contact) 
  - Type: document type (also a link that will open the modal below)
  - Due Date: the deadline for the document
  - Expiry Date: the date the document will expire
  - Status: status of the document
- All columns in view to be sortable
- The documents manager view will be removed. Documents not assigned to you, but otherwise related to you, will be instead viewable from the admin area only.
- The order of the filters will be:
  - Awaiting upload
  - Awaiting approval
  - Approved
  - Rejected
  - All
- Awaiting upload will be selected by deafult
- When status is 'awaiting upload' an upload button will be shown
- In other statuses these buttons will be shown
  - View
  - Download (if document has files)

## Before
The document manager exists and the documents widget appeared like this:

image

## After
The document manager is gone and the documents widget has different columns, which are sortable, as well as different buttons depending on status.

image

## Comments
The "view" mode for the modal will be added in another ticket by a FE developer, most likely Buddha. For now a new argument, "mode" is passed to the modal creation.

---

- [ ] Tests Pass

Unfortunately another case of "what was broken is still broken, what was working is still working". The changes don't affect what few tests we have for this repo


civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -78,43 +104,54 @@
...
+                    $attribute = drupal_attributes($field_attributes[$field][$row_count]);
+                  }
 
-                <?php print $content; ?>
-                </td>
-                <?php endforeach; ?>
-              <td ct-spinner ct-spinner-id="document-<?php print strip_tags($row['id']); ?>">
-                <?php if (strip_tags($row['status_id']) == 3): ?>
+                  printf('<td %s %s>%s</td>', $class, $attribute, $content);
+                endforeach; ?>
```
davialexandre: Since this is PHP only, there's no need to use the alternative syntax for the foreach loop. Also, it looks a bit weird the fact that the foreach uses the alternative syntax and the ifs don't

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -3,57 +3,76 @@
...
+ *   An array of classes to apply to each row, indexed by row number.
+ * @var array $rows
+ *   An array of row items. Each row is an array of content. $rows are keyed by
+ *   row number, fields within rows are keyed by field ID.
+ * @var array $field_classes
+ *   An array of classes to apply to each field, indexed by field id, then row.
+ * @var array $field_attributes
+ *   Array of arrays of attributes indexed by field id and then row.
+ * @var string $attributes
+ *   Empty
```
mickadoo: It wasn't included in the original list and I have no information on what it represents except that it appears to always be empty in this template. If you have an idea what it is I can change it.

mickadoo: I've found that it at least contains the table summary if it is set, so I've updated the description.

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -224,9 +224,18 @@ function filterTable(statusId) {
 
     // Listen for ready event when T&A finishes loading all modules
     document.addEventListener('taReady', function () {
+      var selectedFilter = window.location.hash.substring(1);
```
mickadoo: Angular bootstrap removes the window hash and query string, so had to do it before that happened.

civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php
```diff
@@ -96,66 +96,79 @@
...
+            $mode = 'edit';
+            $label = 'Upload';
+            $faIcon = 'fa-upload';
+          }
+
+          ?>
+          <!-- no results row -->
+          <tr id = "no-results-row" style="display:none;">
+            <td colspan="100%">There are no documents to display.</td>
+          </tr>
```
davialexandre: If this is inside the rows loop, how it will be displayed when there are no documents? Won't the $rows array be empty in that case?

mickadoo: I would have expected that too, but when you have a "no results behavior" in the view, and you have no results, the "no results text" is stored in $rows as 

```
$rows = [
  ["There are no documents to display."]
];
```
As you've probably noticed we don't need this text stored in the view anymore, because we're adding it in the template. However removing the "no results behaviour" from the view will result in an empty screen. 

I needed to move it to the template because of the mix of angular filters and drupal views going on here. Drupal views will show the no results text only when there are zero overall results. But for this ticket we need to show it when there are no results matching the selected filter too. I've fixed that an pushed my changes.

This comment did make me notice that I was putting the no results row in the wrong place. It should be outside the loop. Originally it was part of an if statement in the loop checking for `$row[0]` which would indicate no results, but since we always need it, it should be outside of the loop altogether.

PR #343, opened 2017-08-21
PCHR-2474: Add Floating Signup Tab Block

mickadoo:
## Overview
In order to attract more people to sign up, the demo sites will have a floating tab that, when clicked, shows links to the signup page. The tab will be minimized by default.

## Before
There was no tab on demo sites.

## After
A block has been added that shows the tab. It is disabled by default and will only be enabled on demo sites.

anim

## Notes

All the files changes in this were created by exporting features.

---

- [ ] Tests Pass
Doesn't affect any tests in civihr_employee_portal, some of which are broken.


am2100:
Functionality looks great. A couple of styling points:

1. I think you've reused the CiviHR styles but the tab, button and hyperlink colours should be #18BCB9
2. The text for the button should be uppercase, bold, and turned 180 degrees
3. The padding on the left and right looks like it needs to be increased (design has 30px)
4. Title 'Become an early adopter' should be a much larger font size than the text. Whilst the design is using museo sans, we're probably not going to get the same paragraph to header sizing but if please choose a header/style a class to get something closer to the design (if it's helpful - design has: body copy is 14px, title is 26px). 

That's all from me!

mattlinduk:
Functionality looks great. A couple of styling points:

I think you've reused the CiviHR styles but the tab, button and hyperlink colours should be #18BCB9
The text for the button should be uppercase, bold, and turned 180 degrees
The padding on the left and right looks like it needs to be increased (design has 30px)
Title 'Become an early adopter' should be a much larger font size than the text. Whilst the design is using museo sans, we're probably not going to get the same paragraph to header sizing but if please choose a header/style a class to get something closer to the design (if it's helpful - design has: body copy is 14px, title is 26px).
That's all from me!

mickadoo:
@mattlinduk I've updated based on your recommendations. Please re-review

mickadoo:
@mattlinduk I forgot the shadow effect, I've added a box shadow. Here's a gif comparing it with a normal border, and borderless.

anim

mattlinduk:
- Let's go with the borderless option - it's in the design and fits the theme better. 
- Regarding the box shadow, it looks much heavier than the shadows on the rest of the page. In the design, the shadow on the popup looks as if it is an equivalent weight to those already on the page. 
Could you look into those?

mickadoo:
@mattlinduk Updated the shadow box to be lighter, matching the design:

anim

mickadoo:
Closing as this was moved to a separate repo, [CiviHR Signup](https://github.com/compucorp/civihr-signup)

PR #350, opened 2017-09-14
(WIP) PCHR-2560: Onboarding Email

mickadoo:
## Overview
For new users invited to the system by either the bulk "Create User Accounts" action from the advanced search, or by clicking "Create User Account" from the user profile page, the welcome email should match the new format. The email content will be generated in the civihr hremails extension using Foundation. This PR will use that content to replace the System email body.

## Before
The "account activated" system email is generated by Drupal and uses the Drupal template.

image

## After
The "account activated" system email uses the template from hremails. In this PR the only change is the new template and partials.

image

## Notes

- The email content is not final yet so this PR is still WIP

---

- [ ] Tests Pass
Only some tests work in civihr_employee_portal. This changes does not affect the existing tests.


mickadoo:
Closed in favor of #365

PR #352, opened 2017-09-27
PCHR-2657: Allow users with (edit terms in HR Resource Types) permission to add new types

omarabuhussein:
## Overview
Allow users with "**edit terms in HR Resource Types**" to add new types

## Before
Users with "**edit terms in HR Resource Types**"  permission were not able to access "**site-url/admin/structure/taxonomy/hr_resource_type**" to see the current types which was fixed by creating a new view to show this taxonomy vocabulary terms at "**site-url/hr-resource-types-list**" long time ago in this PR "https://github.com/compucorp/civihr-employee-portal/pull/210".

But the problem is that the user with this permission and access to this view cannot add a new term, but only editing or viewing the current terms which is not very useful. The only way was to achieve that is to give the user "**Administer vocabularies and terms**" permission which is not a good solution.

## After
The button is added by overriding "**views-view--hr-resource-types-list--page.tpl.php**" and adding the button markup to it. it points to the form that allows you to add new HR Resource type term "**site-url/admin/structure/taxonomy/hr_resource_type/add**" : 

2017-10-04 19_26_02-hr resource types list _ hr16_11

2017-10-04 19_33_36-hr resource types list _ hr16_11

Also the permissions to "**site-url/admin/structure/taxonomy/hr_resource_type/add**" path where changed so if the user has "**edit terms in HR Resource Types**"  permission the add form can be accessed.

2017-09-27 18_33_15-add term _ hr16_11

## Technical Details

The permissions of "Add HR Resource Type" form is altered by using **civihr_employee_portal_menu_alter** hook "**hook_menu_alter**", the path used is "'admin/structure/taxonomy/%taxonomy_vocabulary_machine_name/add'" since using the direct path "admin/structure/taxonomy/hr_resource_type/add" won't work and an **access callback** was defined "**_civihr_employee_portal_taxonomy_form_access_callback**" that ensures that the current user is either has **administer taxonomy** permission, or **edit terms in HR Resource Types** permission and the add form is "HR Resource Type" only and not other taxonomy vocabulary. 

---

- [ ] Tests Pass (No tests to run)


civihr_employee_portal/views/views_export/views_hr_resource_types_list.inc
```diff
@@ -26,6 +26,14 @@ $handler->display->display_options['exposed_form']['type'] = 'basic';
 $handler->display->display_options['pager']['type'] = 'none';
 $handler->display->display_options['pager']['options']['offset'] = '0';
 $handler->display->display_options['style_plugin'] = 'table';
+/* Header: Global: Text area */
+$handler->display->display_options['header']['area']['id'] = 'area';
+$handler->display->display_options['header']['area']['table'] = 'views';
+$handler->display->display_options['header']['area']['field'] = 'area';
+$handler->display->display_options['header']['area']['label'] = 'Add HR Resource Type';
+$handler->display->display_options['header']['area']['empty'] = TRUE;
+$handler->display->display_options['header']['area']['content'] = '<a class="btn btn-default btn-primary" href="/admin/structure/taxonomy/hr_resource_type/add"><span class="fa fa-plus"></span> Add HR Resource Type</a>';
```
davialexandre: I think we should add this button using a template instead. It would give us more flexibility to change its position/styling without having to update the view itself

omarabuhussein:
> I checked the ticket and I couldn't find any discussion about the position of the button. Did you check with Mike if the position you choose is ok?

No to be honest since I found this approach while investigating the problem, I put it in the place where buttons get usually added, but I will tag @mi-kon

civihr_employee_portal/views/views_export/views_hr_resource_types_list.inc
omarabuhussein: do you have any example for a view that use template since I am not really that familiar with such stuff?

JonScreat:
I think the position of the button is fine - 'Add new' usually goes in the top left. My only comments would be that there is that the margin above the button looks too big (there should only be 20px between the top of the button and the edge of the panel).

civihr_employee_portal/views/views_export/views_hr_resource_types_list.inc
davialexandre: Here is an example of how it is used by the staff directory view: https://github.com/compucorp/civihr-employee-portal/blob/c6ac2e215f87c868b7345bd9e6755d4774e960ab/civihr_employee_portal/views/templates/views-view--civihr-staff-directory--page.tpl.php

Better confirm this with a Drupal dev, but if I'm not wrong, the file follows this structure:
`views-view--<view name>--<display name>.tpl.php`

JonScreat:
Only comment would be if you could add an additional 20px margin to the bottom of the button so that there's a little more space between the button and table. Thanks

civihr_employee_portal/views/templates/views-view-table--hr-resource-types-list--page.tpl.php
```diff
@@ -0,0 +1,55 @@
...
+
+<table <?php if ($classes) { print 'class="'. $classes . '" '; } ?><?php print $attributes; ?>>
+  <?php if (!empty($title) || !empty($caption)) : ?>
+    <caption><?php print $caption . $title; ?></caption>
+  <?php endif; ?>
+  <?php if (!empty($header)) : ?>
+    <thead>
+    <tr>
+      <?php foreach ($header as $field => $label): ?>
+        <th <?php if ($header_classes[$field]) { print 'class="'. $header_classes[$field] . '" '; } ?> scope="col">
```
davialexandre: I wonder if it wouldn't be better to wrap the if condition in an `empty()` to avoid a `E_NOTICE` in case the index doesn't exist. If you think that it's better, please do the same to the other ifs in this template

omarabuhussein: my mistake, I should added this file in two separate commits to avoid confusion, but this is actually the default view page syntax which is generated by drupal, I only added this div to it https://github.com/compucorp/civihr-employee-portal/blob/47016e5a60831e166d7ab108e078b4d072563d0f/civihr_employee_portal/views/templates/views-view-table--hr-resource-types-list--page.tpl.php#L23

and I would rather not to touch drupal generated syntax unless we discovered there is a reason to do that.

civihr_employee_portal/views/templates/views-view--hr-resource-types-list--page.tpl.php
```diff
@@ -0,0 +1,97 @@
...
+  </div>
+<?php endif; ?>
+
+<?php if ($feed_icon): ?>
+  <div class="feed-icon">
+    <?php print $feed_icon; ?>
+  </div>
+<?php endif; ?>
+
+  </div><?php /* class view */ ?>
```
davialexandre: what is this `<?php /* class view */ ?>` for?

omarabuhussein: drupal stuff, this is just a straight copy from drupal auto generated template

PR #353, opened 2017-09-28
FAR-198: Allow requesting leave for dates matching the end/beginning of Daylight Saving Time

davialexandre:
## Overview

It was not possible to request leave if one of the request dates were the start or the end of the Daylight Saving Time in the timezone of the current user, set in Drupal. Whenever someone tried that, the system would return a server error and the list of dates in the Leave Request modal wouldn't be updated.

The reason was that in some places in the SSP code was required to add 1 day to a date. This was implemented with a logic that would convert the date to a unix timestamp and then add 86400 seconds to it (the number of seconds in a day), which works fine for days with 24 hours. However, in countries/timezones with Daylight Saving Time, we can have days with 25 or 23 hours, that will make the code fail.

In 2017, for example, the UK Daylight Saving Time ends on 29th Oct. This day has 25h long. If PHP's timezone is set to `Europe/London`, and we add 86400 to it, the resulting date will still be 2017-10-29, as it can be seen here: https://3v4l.org/i5HEB

## Before

The logic to add 1 day to dates didn't work for days that were not exactly 24h long. This made it impossible to request leave containing either the start or the end date of Daylight Saving Time period.

## After

It's now possible to request leave containing the start or the end date of a Daylight Saving Time period. The logic to add 1 day to dates was improved. Instead of doing the math manually, we now use a `DateTime` object and it's `modify` method to add 1 day to it. The DateTime object is aware of timezones and Daylight Saving Times and can do the proper adjustments automatically. 

## Comments
Whenever you don't use a DateTime object for date calculations, you cause a rupture in space time's fabric. If done continuously, the effects could be catastrophic. 

---

- [ ] Tests Pass
We don't have test coverage for this Part of the system.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -6980,21 +6980,27 @@ function civihr_employee_portal_module_implements_alter(&$implementations, $hook
 
 /**
  * Check if a passed date is weekend day or not
+ *
+ * @param \DateTime $date
+ *
+ * @return bool
  */
-function _isWeekend($date_passed) {
-
-    $date = explode("-", $date_passed);
-
-    $time = mktime(0, 0, 0, $date[1], $date[2], $date[0]);
-    $weekday = date('w', $time);
-    return ($weekday == 0 || $weekday == 6);
+function _isWeekend(\DateTime $date) {
+    $saturday = 6;
+    $sunday = 0;
+    $weekday = $date->format('w');
+    return ($weekday == $saturday || $weekday == $sunday);
 }
 
 /**
  * @param $public_holidays
- * Check if the requested day is holiday or not working day
+ *  Check if the requested day is holiday or not working day
+ * @param \DateTime $date
+ *  The requested day
+ *
+ * @return array
  */
-function _checkRequestedDay($public_holidays, $date) {
+function _checkRequestedDay($public_holidays, \DateTime $date) {
 
     $not_working_day = t('Weekend');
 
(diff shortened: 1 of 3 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6980,21 +6980,27 @@ function civihr_employee_portal_module_implements_alter(&$implementations, $hook
...
-
-    $date = explode("-", $date_passed);
-
-    $time = mktime(0, 0, 0, $date[1], $date[2], $date[0]);
-    $weekday = date('w', $time);
-    return ($weekday == 0 || $weekday == 6);
+function _isWeekend(\DateTime $date) {
+    $saturday = 6;
+    $sunday = 0;
+    $weekday = $date->format('w');
```
tunbola: Please leave a line before the return statement.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6980,21 +6980,27 @@ function civihr_employee_portal_module_implements_alter(&$implementations, $hook
...
+    return ($weekday == $saturday || $weekday == $sunday);
```
tunbola: What about  `return $date->format('N') >= 6`

davialexandre:
@tunbola thanks for the review, but would it be ok for you if I didn't make the changes you've suggested? The reason is that all this code will be gone once we merge la-milestone-3 into staging in +- 1 month, so there's no point in spending more time to make it look good.

tunbola:
@davialexandre , You're right. In that case, PR approved.

PR #358, opened 2017-10-09
PCHR-2457: Document open buttons take very long to load on ssp

bmnepali:
## Overview
This PR makes some code refactoring and displays the spinner in Document modal removing the spinner in place of "upload" button in a document list, when trying to open a document modal.

## Before
- The spinner was displayed in place of "upload" button, when clicking the button in document list.
2457-issue

## After
- The spinner was displayed in modal popup, when clicking the "upload" button in document list.
2457-fix

## Technical Details
1. Display the spinner after the modal is loaded in file `civihr_employee_portal/js/ta-documents-app.js`
```js
function openModalDocument(data, role, mode) {
   var modalInstance = $modal.open({
      ...
   });
   modalInstance.opened.then(function () {
        $rootScope.isLoading = true;
   });
}
```
2. Hide the spinner when the contacts are loaded and cached.
```js
...
vm.cacheContacts(data).then(function () {
    $rootScope.isLoading = false;
});
```
3. Remove the previous spinner related codes from js file.
```js
$rootScope.$broadcast('...');
...
vm.loadingModalData = false;
...
```
4. Remove/Change the button markup for a document in a list in SSP in file `civihr_employee_portal/views/templates/views-view-table--Documents--block.tpl.php` form:
```html
<td data-ct-spinner data-ct-spinner-id="document-<?php print $rowID; ?>">
     <button
         ng-show='!document.loadingModalData'
         ng-click="document.modalDocument('<?php print $rowID ?>', 'staff', '<?php print $mode ?>')"
         class="btn btn-sm btn-default">
         <i class="fa <?php print $faIcon ?>"></i>
         <?php print $label ?>
      </button>
    ...
</td>
```
to
```html
<td>
    <button ng-click="document.modalDocument('<?php print $rowID ?>', 'staff', '<?php print $mode ?>')" class="btn btn-sm btn-default">
         <i class="fa <?php print $faIcon ?>"></i>
         <?php print $label ?>
     </button>
    ...
<td>
```
That is to remove the spinner related code form button markup.

### DIFF
--- a/civihr_employee_portal/js/ta-documents-app.js
+++ b/civihr_employee_portal/js/ta-documents-app.js
@@ -16,8 +16,6 @@
         var availableContacts = false;
         var vm = this;
 
-        vm.loadingModalData = false;
-
         vm.modalDocument = modalDocument;
         vm.openModalDocument = openModalDocument;
         vm.cacheContacts = cacheContacts;
@@ -46,17 +44,14 @@
          * @param {string} mode
          */
         function modalDocument (id, role, mode) {
-          $rootScope.$broadcast('ct-spinner-show', 'document-' + id);
-          vm.loadingModalData = true;
-
           DocumentService.get({ id: id })
             .then(function(data) {
               if (!data) {
                 throw new Error('Requested Document is not available');
               }
-
+              vm.openModalDocument(data[0], role, mode);
               vm.cacheContacts(data).then(function () {
-                vm.openModalDocument(data[0], role, mode);
+                $rootScope.isLoading = false;
               });
             })
             .catch(function(reason) {
@@ -98,8 +93,7 @@
           });
 
           modalInstance.opened.then(function () {
-            $rootScope.$broadcast('ct-spinner-hide');
-            vm.loadingModalData = false;
+            $rootScope.isLoading = true;
           });
         }
 
--- a/civihr_employee_portal/js/ta-documents-app.js
+++ b/civihr_employee_portal/js/ta-documents-app.js
@@ -17,8 +17,6 @@
         var vm = this;
 
         vm.modalDocument = modalDocument;
-        vm.openModalDocument = openModalDocument;
-        vm.cacheContacts = cacheContacts;
 
         (function init() {
           // Sets the date format for HR_settings.DATE_FORMAT
@@ -45,16 +43,19 @@
          */
         function modalDocument (id, role, mode) {
           DocumentService.get({ id: id })
-            .then(function(data) {
-              if (!data) {
-                throw new Error('Requested Document is not available');
-              }
-              vm.openModalDocument(data[0], role, mode);
-              vm.cacheContacts(data).then(function () {
-                $rootScope.isLoading = false;
-              });
+            .then(function (data) {
+              return {
+                data: data,
+                isOpened: openModalDocument(data[0], role, mode)
+              };
             })
-            .catch(function(reason) {
+            .then(function (result) {
+              return cacheContacts(result.data);
+            })
+            .then(function () {
+              $rootScope.isLoading = false;
+            })
+            .catch(function (reason) {
               CRM.alert(reason, 'Error', 'error');
             });
         };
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -46,17 +44,14 @@
...
-          $rootScope.$broadcast('ct-spinner-show', 'document-' + id);
-          vm.loadingModalData = true;
-
           DocumentService.get({ id: id })
             .then(function(data) {
               if (!data) {
                 throw new Error('Requested Document is not available');
               }
-
+              vm.openModalDocument(data[0], role, mode);
```
AkA84: if you make `openModalDocument` return the promise of `modalInstance.opened.then()`, then you can write a more expressive code

```js
vm.openModalDocument(data[0], role, mode)
  .then(function () {
    return vm.cacheContact(data);
  })
  .then(function () {
    $rootScope.isLoading = false;
  });
```

AkA84: why is `openModal` published on the `vm` object? Here you are just calling it internally. Is the function called also on the view?

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -46,17 +44,14 @@
...
               vm.cacheContacts(data).then(function () {
-                vm.openModalDocument(data[0], role, mode);
+                $rootScope.isLoading = false;
```
AkA84: why are you using `$rootScope`?

civihr_employee_portal/js/ta-documents-app.js
AkA84: Even better, i think
```js
  DocumentService.get({ id: id })
    .then(function (data) {
      return data;  // or if not present, throw an exception
    })
    .then(function (data) {
      return vm.openModalDocument();
    })
    // etc..
    .catch(function (reason) {
      // the code should go here even if the first `then` throws an error, please double check
    });
```

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -46,17 +44,14 @@
...
               });
             })
             .catch(function(reason) {
```
AkA84: the fact that there is no space between `function` and `(reason)` should have thrown an error in the linter, are you sure you have the git hook configured correctly

civihr_employee_portal/js/ta-documents-app.js
bmnepali: Since `isLoading` is used to display or hide the loader in document modal. In case, when the document modal is loaded form SSP and if the `isLoading` is bound to `vm` in SSP 
it cannot be accessed in modal controller  T&A and vice-versa. So I have used `$rootScope` so, that the changes to `isLoading` in either SSP or T&A  will be reflected in document modal.

civihr_employee_portal/js/ta-documents-app.js
bmnepali: You mean `openModalDocument()`? No, its not used any where else. I will make it private.

civihr_employee_portal/js/ta-documents-app.js
bmnepali: I had permission issue in executing the pre-commit hook. Thanks for pointing it out! fixed!

civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -45,16 +43,19 @@
...
           DocumentService.get({ id: id })
-            .then(function(data) {
-              if (!data) {
-                throw new Error('Requested Document is not available');
-              }
-              vm.openModalDocument(data[0], role, mode);
-              vm.cacheContacts(data).then(function () {
-                $rootScope.isLoading = false;
-              });
+            .then(function (data) {
```
AkA84: where did the "throw new Error" go?

bmnepali: It's not required not as the exception (if no data is found) will be handled by`.catch(function (reason) {` block.

PR #365, opened 2017-10-16
PCHR-2560: Onboarding Email

mickadoo:
## Overview
The onboarding epic requires two new email templates; one for the system invitation mail and the other for the password reset mail.

## Before

Both emails used the default Drupal mail.

#### System Invite
image

#### Password reset
image

## After

Two new templates exist created with Foundation

#### System Invite
image

#### Password reset
image

## Comments

- Content in civihr_default_mail_content is auto generated by features.
- The templates are generated using Foundation from inside the CiviHR repository.
- The actual content of these mails should be configurable from the admin area, so these templates are just providing a button and a placeholder for the mail content. 
- The default content will be imported using the new civihr_default_mail_content feature

- tests pass [x]


mickadoo:
Closed automatically since 2339 branch was merged. Recreated pointing to 2112 in #366

PR #369, opened 2017-10-23
PCHR-2801: Profile Image Resize

mickadoo:
## Overview

When uploading profile images of a non-standard ratio they become distorted and affect the layout on the dashboard.

## Before

Images that don't "fit" for profile pictures become distorted.

#### Dashboard

image

#### Thumbnail in menu

image

#### CiviCRM profile

image

#### Dashboard

image

#### Thumbnail in menu

image

#### CiviCRM profile

image

Images will be resized and scaled. There will also be a recommended size in the dashboard user edit form.

## Comments

It was planned to add the recommended size text to the CiviCRM contact edit page but this will be delayed for now.

From the images in the PR you can see it doesn't magically fix images that don't fix, but the aim is to not have these images break the layout.

---

- [x] Tests Pass


civihr_employee_portal/src/Helpers/ImageHelper.php
```diff
@@ -0,0 +1,16 @@
...
+
+class ImageHelper {
+  public static function resizeForProfile($imagePath) {
+    $image = image_load($imagePath);
+
+    if (!$image) {
+      return;
+    }
+
+    image_scale_and_crop($image, 400, 400);
```
omarabuhussein: Avoid magic numbers, I would define two constants (PROFILE_IMAGE_DEFAULT_WIDTH and PROFILE_IMAGE_DEFAULT_WIDTH & PROFILE_IMAGE_DEFAULT_HIGHT) and use them instead.

civihr_employee_portal/src/Helpers/ImageHelper.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Drupal\civihr_employee_portal\Helpers;
+
+class ImageHelper {
+  public static function resizeForProfile($imagePath) {
```
omarabuhussein: add docblock to this method

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -4675,6 +4676,52 @@ function civihr_employee_portal_form_user_profile_form_alter(&$form, &$form_stat
...
+    ImageHelper::resizeForProfile($file->uri);
+  }
+}
+
+/**
+ * @param string $formName
+ * @param CRM_Contact_Form_Contact $form
+ */
+function civihr_employee_portal_civicrm_postProcess($formName, &$form) {
+  if ($formName !== 'CRM_Contact_Form_Contact') {
```
omarabuhussein: I think this would be more clean : 
```php
switch($formName){
case 'CRM_Contact_Form_Contact':
   _civihr_employee_portal_resize_civicrm_contact_image($formName, &$form);
break;
}

```

where the rest of code below go into _civihr_employee_portal_resize_civicrm_contact_image() ..

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -4675,6 +4676,52 @@ function civihr_employee_portal_form_user_profile_form_alter(&$form, &$form_stat
...
+  if (!$imageChanged) {
+    return;
+  }
+
+  $targetContact = civicrm_api3('Contact', 'getsingle', ['id' => $contactID]);
+  // Since this is postProcess image_URL is already pointing to new image
+  $imageUrl = $targetContact['image_URL'];
+
+  // Get the full path to the image
+  $queryString = parse_url($imageUrl, PHP_URL_QUERY);
```
omarabuhussein: the lines from 4714 to 4719 can go into separate function **_civihr_employee_portal_get_image_path_from_contact_image_url($imageUrl)**

civihr_employee_portal/civihr_employee_portal.install
```diff
@@ -166,6 +166,36 @@ function civihr_employee_portal_update_7008() {
 }
 
 /**
+ * Update My Details webform image description
+ */
+function civihr_employee_portal_update_7010() {
```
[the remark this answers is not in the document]
mickadoo: I'm actually working on another ticket that adds 7009 so I didn't want to have the conflict there later :smile:

civihr_employee_portal/civihr_employee_portal.module
mickadoo: I don't like the ever expanding size of the module file, so I moved this into a static method in another class.

civihr_employee_portal/civihr_employee_portal.module
mickadoo: Moved them to ImageHelper class, I already know another ticket where I'll be using this function.

civihr_employee_portal/src/Helpers/ImageHelper.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Drupal\civihr_employee_portal\Helpers;
+
+class ImageHelper {
```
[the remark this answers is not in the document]
mickadoo: Agreed, will change to ImageResizer. I'll move the method to get the path to a contact image to a separate class then too.

civihr_employee_portal/src/Helpers/ContactImagePathfinder.php
```diff
@@ -0,0 +1,28 @@
...
+
+class ContactImagePathfinder {
+  /**
+   * @param $contactID
+   *   The ID of the target contact
+   *
+   * @return string|NULL
+   *   The image path if it exists, or NULL if not set or doesn't exist.
+   */
+  public static function getPathToContactImage($contactID) {
```
omarabuhussein: there are some repeated words between the class and this method names, I would either change the class name or the method name : 

Example 1)  ImagePathfinder::getPathToContactImage()

Example 2) ContactImagePathfinder:getPath();

what do you think ?

mickadoo: I think you're right. It's a very specific class, but that's not necessarily a bad thing. I prefer the second one because it's clear to and dev looking at php files what services are available

PR #371, opened 2017-10-24
PCHR-2817: Fix bug in fetching webform submissions.

mickadoo:
## Overview

Webform submissions are checked on login. If a user has not completed the onboarding webform they will be redirected to it. There is a bug in submission checking that means webform submissions are not correctly filtered, so if a user has other webform submissions (e.g. edit my details) it will incorrectly assume they have completed onboarding.

## Before

If a user has "edit my details" submissions, but no onboarding webform submissions they won't be redirected to the onboarding form on login.

## After

The user will be redirected to the onboarding form if they haven't completed it, regardless of other webform submissions.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -5036,7 +5036,7 @@ function _civihr_employee_portal_should_do_onboarding($account) {
   $onboardingWebform = WebformHelper::findOneByTitle($formName);
 
   if (!$onboardingWebform) {
-    throw new \Exception("Could not find onboarding webform");
+    throw new \Exception('Could not find onboarding webform');
   }
 
   $canViewOnboardingForm = node_access('view', $onboardingWebform, $account);
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -5033,10 +5033,14 @@ function civihr_employee_portal_user_login(&$edit, $account) {
  */
 function _civihr_employee_portal_should_do_onboarding($account) {
   $formName = OnboardingWebForm::NAME;
-  $nodes = node_load_multiple(NULL, ['title' => $formName]);
-  $node = current($nodes);
-  $canViewOnboardingForm = node_access('view', $node, $account);
-  $submissions = WebformHelper::getUserSubmissionsByName($account, $formName);
+  $onboardingWebform = WebformHelper::findOneByTitle($formName);
+
+  if (!$onboardingWebform) {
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5033,10 +5033,14 @@ function civihr_employee_portal_user_login(&$edit, $account) {
...
 function _civihr_employee_portal_should_do_onboarding($account) {
   $formName = OnboardingWebForm::NAME;
-  $nodes = node_load_multiple(NULL, ['title' => $formName]);
-  $node = current($nodes);
-  $canViewOnboardingForm = node_access('view', $node, $account);
-  $submissions = WebformHelper::getUserSubmissionsByName($account, $formName);
+  $onboardingWebform = WebformHelper::findOneByTitle($formName);
+
+  if (!$onboardingWebform) {
+    throw new \Exception("Could not find onboarding webform");
```
davialexandre: You can use single quotes here

civihr_employee_portal/src/Helpers/WebformHelper.php
```diff
@@ -7,6 +7,21 @@
 class WebformHelper {
 
   /**
+   * @param $title
```
davialexandre: Missing type here

davialexandre: Please add a description to this method

civihr_employee_portal/src/Helpers/WebformHelper.php
```diff
@@ -30,22 +45,25 @@ public static function getWebformComponentsByName($node, $name) {
...
+    $webformId = $webform->nid;
 
-    foreach ($submissions as $submission) {
-      $title = property_exists($submission, 'title') ? $submission->title : NULL;
-      if ($title === $name) {
-        $submissions[] = $submission;
+    foreach ($allUserSubmissions as $submission) {
+      $nodeID = property_exists($submission, 'nid') ? $submission->nid : NULL;
+      if ($webformId === $nodeID) {
+        $targetFormSubmissions[] = $submission;
```
davialexandre: Perhaps I'm slow today, but I can't see what is the difference here. Before we would only return the submissions with the same title as the given name, now we get the webform by title then return the submissions where the node ID is the same as the webform ID. Would you please give more details about this?

mickadoo: I'm confused myself, when I was debugging this I couldn't find any submission which had the property 'title', must have just been a mistake by me from the beginning.

PR #380, opened 2017-10-31
PCHR-2848: Updates to "My Details" Block

mickadoo:
## Overview

The "My Details" block has a few different displays. We're using different ones on the `/hr-details` page and `/dashboard`. These will be unified and old displays removed.

## Before

- `/hr-details` used a different display for the "My Details" view than on `/dashboard`.
- The master display for this view did not show Contact ID and Birth Date
- The address in the "My Address View" showed each address field separately
- The "My Details" view had 4 displays

#### master Display

image

#### hr-details Display

image

## After

- `/hr-details` uses the same "My Details" (master) display as on `/dashboard`.
- The master display for this view adds Contact ID and Birth Date
- The address in the "Contact Information (renamed)" is aggregated to a single field
- 2 obsolete displays are removed from "My Details" 

#### both hr-details and master

image

## Technical Details

Previously we were adding all the address fields to the view, hiding them, and then using a global text field and replacement tokens to show the address as one field. It seems cleaner to use a custom handler so I added one.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/views/includes/civihr_employee_portal_handler_aggregated_address.inc
+++ b/civihr_employee_portal/views/includes/civihr_employee_portal_handler_aggregated_address.inc
@@ -60,13 +60,15 @@ class civihr_employee_portal_handler_aggregated_address extends views_handler_fi
   }
 
   /**
+   * Returns the HTML for the address as it will appear in the view.
+   *
    * @param array $address
    *   The address data
    * @return string
    *   The formatted address
    */
   private function formatAddress($address) {
-    $format = '%s<br/>%s<br/>%s %s<br/>%s<br />%s';
+    $format = '%s<br/>%s<br/>%s %s<br/>%s<br/>%s';
 
     return sprintf(
       $format,
--- a/civihr_employee_portal/views/includes/civihr_employee_portal_handler_aggregated_address.inc
+++ b/civihr_employee_portal/views/includes/civihr_employee_portal_handler_aggregated_address.inc
@@ -1,0 +1,81 @@
+<?php
+
+use CRM_Utils_Array as ArrayHelper;
+
+/**
+ * Handles aggregation of a contacts different address fields
+ */
+class civihr_employee_portal_handler_aggregated_address extends views_handler_field {
+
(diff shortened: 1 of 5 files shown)


civihr_employee_portal/views/includes/civihr_employee_portal_handler_aggregated_address.inc
```diff
@@ -0,0 +1,81 @@
...
+    return array_shift($address['values']);
+  }
+
+  /**
+   * @param array $address
+   *   The address data
+   * @return string
+   *   The formatted address
+   */
+  private function formatAddress($address) {
```
tunbola: Can you add a small description to the function.

civihr_employee_portal/views/includes/civihr_employee_portal_handler_aggregated_address.inc
```diff
@@ -0,0 +1,81 @@
...
+    $format = '%s<br/>%s<br/>%s %s<br/>%s<br />%s';
```
tunbola: Seems the space before the closing tag for the last `<br />` is not consistent with the others.
probably make all to have a space before the closing tag.

civihr_employee_portal/views/includes/civihr_employee_portal_handler_aggregated_address.inc
```diff
@@ -0,0 +1,81 @@
...
+
+    return sprintf(
```
tunbola: Good use of `sprintf` here.

tunbola:
@mickadoo , The before screenshot still shows contact ID and Birth date but the notes say otherwise. Is it the correct one?

mickadoo:
@tunbola 

> The before screenshot still shows contact ID and Birth date but the notes say otherwise. Is it the correct one?

Sorry, it was a bit confusing. Before the change the "master" display did not show contact ID and birth date. After the change it does, and now both `/dashboard` and `/hr-details` use the master display. 

I've updated the description and added another screenshot of the previous master display.

PR #382, opened 2017-11-01
PCHR-2339: Fix group ID on node import

mickadoo:
## Overview

A follow up from the previous change which added some group metadata when exporting nodes. I failed to allow for the fact that the CiviCRM webform part of the export keeps a count of the number of multi value custom data groups are in use for the webform. 

The key for this uses the custom group ID and if this changes then all custom fields for this group will be disabled.

## Before

When importing a node where the custom group changed the individual field mapping would be correct, but the count of custom groups in use would be wrong.

## After

Both custom group and custom field mapping will use the group and field machine name to lookup the new ID on import. The CiviCRM webform count of custom groups in use on the webform will also use the correct custom group ID. 

## Technical Details

CiviCRM webform saves a count of custom groups that are in use like this:

```
  'number_of_cg7' => '0',
  'number_of_cg8' => '0',
  'number_of_cg9' => '1',
```

When importing if the custom group ID does not match any of those it will disable all custom fields in the webform that belong to that group. For example if I have custom group "Inline Custom Data" that has a custom field "NI/SSN", but the custom group ID for "Inline Custom Data" is different on the production server the webform will import the fields, but there will be no link to CiviCRM because the count will not match.

---

- [x] Tests Pass


mickadoo:
@omarabuhussein I think you're right, I'll add some tests for this

PR #385, opened 2017-11-02
PCHR-2339: QA Fixes

mickadoo:
## Overview

Fixes some issues that appeared in QA

## Before

- Work email was not being set as primary if home email was already primary
- Cancel button was navigating to either T&A dashboard or dashboard
- Save button was in the wrong place and have wrong text ("Save Configuration")

image

## After

- Work email is always set as primary on submission if it is set
- Cancel button navigates to `/dashboard`
- Save button was is on the right side and has the right text ("Save")

image

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/src/Forms/OnboardingWebForm.php
+++ b/civihr_employee_portal/src/Forms/OnboardingWebForm.php
@@ -15,16 +15,23 @@ class OnboardingWebForm {
    * Some required processing such as clearing caches and creating tasks.
    *
    * @param \stdClass $node
-   * @param \stdClass $submission
+   *   The webform node
+   * @param \stdClass $values
+   *   The submitted values
    */
-  public function onSubmit($node, $submission) {
+  public function onSubmit($node, $values) {
     $contactID = \CRM_Core_Session::singleton()->getLoggedInContactID();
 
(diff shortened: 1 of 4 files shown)


civihr_employee_portal/src/Forms/OnboardingWebForm.php
```diff
@@ -83,16 +80,34 @@ private function createReminderTask($contactID) {
    * If work email is included then create it here. It will be created on
    * webform submission anyway, but we need to make it primary.
```
tunbola: This description need to be updated to say what the function is now doing. It doesn't create an email anymore but updates the work email if it has previously been set by the webform submission.

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
```diff
@@ -171,4 +176,17 @@ private function clickedCancel($formState) {
 
     return \CRM_Utils_Array::value('#value', $clickedButton) === t('Cancel');
   }
+
+  /**
+   * Sets the title and weight for the save button
+   *
+   * @param array $form
```
tunbola: Please leave a line before return.

PR #387, opened 2017-11-06
PCHR-2864: Add Explanation for Onboarding Wizard for Existing Users

mickadoo:
## Overview

Existing CiviHR users might be a bit confused when required to complete onboarding. To help this we will add a message showing them why they're seeing it.

## Before

Anyone who hadn't completed onboarding would be redirected to the onboarding form without explanation.

image

## After

There will be an explanation message on the first page of the onboarding form _only_ if the user was created before the onboarding form release date.

image

## Notes

- Because the logic for form alterations was becoming more complicated I moved it into the form class itself.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/src/Forms/OnboardingWebForm.php
+++ b/civihr_employee_portal/src/Forms/OnboardingWebForm.php
@@ -174,7 +174,8 @@ class OnboardingWebForm {
    */
   private function userCreatedAfterOnboardingReleased() {
     global $user;
-    $onboardingRelease = (new \DateTime('13 November 2017'))->getTimestamp();
+    $onboardingForm = WebformHelper::findOneByTitle(self::NAME);
+    $onboardingRelease = $onboardingForm->created;
 
     return $user->created > $onboardingRelease;
   }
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/src/Forms/OnboardingWebForm.php
```diff
@@ -111,4 +111,72 @@ private function setWorkEmailAsPrimary($node, $values, $contactID) {
...
+
+  /**
+   * Checks if the current logged in user was created after the onboarding
+   * feature was released.
+   *
+   * @return bool
+   */
+  private function userCreatedAfterOnboardingReleased() {
+    global $user;
+    $onboardingRelease = (new \DateTime('13 November 2017'))->getTimestamp();
```
davialexandre: This should not be done for multiple reasons:
- No business logic should be based on a release process
- Releases can be delayed
- Clients sites don't get updated on the same date as the new version is released. Also, not all client sites are updated on the same day. What happens if I created a user after the release but before my site gets updated?

If we need display different things based on if the user was created via the onboarding wizard or some other way, then we need to store this information somewhere. The first thing that comes to my mind is adding a flag to the database.

PR #388, opened 2017-11-07
PCHR-2862: Add the Absence Duration in Hours field to the SSP Leave and Absence Report.

tunbola:
## Overview
Since Leave can now be requested in Hours, a new field is needed on the SSP Leave and Absence Report as well as on the Pivot table that works the same way the Absence Duration in days field work.

## Before
The Absence duration in hours field does not exist.

## After
- The Absence duration in hours field is added to the Leave and Absence report. 
The field is empty for absences in days. Also the Absence Duration in days field is empty for absences in hours.

civihr custom report - civihr 1 7 demo 2017-11-07 13-17-02

civihr custom report - civihr 1 7 demo 2017-11-07 13-22-08

---
This PR does not affect tests.
- [ ] Tests Pass

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -5,6 +5,7 @@ use Drupal\civihr_employee_portal\Helpers\HelperClass;
 use Drupal\civihr_employee_portal\Security\PublicFirewall;
 use Drupal\civihr_employee_portal\Helpers\ImageResizer;
 use Drupal\civihr_employee_portal\Forms\ContactForm;
+use CRM_HRLeaveAndAbsences_BAO_AbsenceType as AbsenceType;
 
 /**
  * Implements hook_css_alter().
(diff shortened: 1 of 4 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -375,6 +376,8 @@ function _rebuild_length_of_service_view() {
 function _rebuild_absence_activity_view() {
   $civi_settings = parse_url(CIVICRM_DSN);
   $civi_db_name = trim($civi_settings['path'], '/');
+  $calculationUnitOptions = array_flip(AbsenceType::buildOptions('calculation_unit', 'validate'));
+  $hoursUnit = $calculationUnitOptions['hours'];
```
mickadoo: Code looks fine, just a bit confused here. What are the possible values for `$hoursUnit`?

tunbola: The calculation unit option group values are 1 and 2. `1` for Absence types having leave request  in days. `2` for absence type having leave request in hours.
In this case above `$hoursUnit = 2`

mickadoo: I see, so it's just fetching the option value `value` and using it in the view? That makes sense

mickadoo: It's a minor issue, but I think since in this case `$hoursUnit = 2` doesn't really make sense, it might be better to name this variable `$hoursUnitID` or `$hoursUnitCode`. The calculation units are hours and days, so 2 is the identifier for the hours unit. Like I said, it's minor though, so I'll leave it up to you to decide.

PR #391, opened 2017-11-08
PCHR-2715: Swap Reports

deb1990:
## Overview
As part of this PR, bugs related to SSP Reports tabs, has been fixed. 

## Problem
The tabs in the reports, did not support routing, so when the pagination was clicked in Data tab, the page got refreshed and the tabs index were reset.

## Before
image

## After
after

## Solution
1. Added routing to the tabs. 
- Tab hash has been added to the `anchor` tags.
- Added `switchTabsOnLoad()` function, which switches the tabs using url hash, during page load.

## Comments
Semistandard Fixes has been done.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/js/reports.js
+++ b/civihr_employee_portal/js/reports.js
@@ -220,7 +220,7 @@
   /*
    * Finding the filter box for field Employee length of service.
    *
-   *  @param {string} filterData
+   *  @param {String} filterData
    *
    *  @return {bool}
    */
@@ -231,14 +231,14 @@
   /*
    * Finding and replacing the value of Employee length of service.
    *
-   *  @param {string} filterSelector - DOM element that has value.
+   *  @param {String} filterSelector - DOM element that has value.
    */
   HRReport.prototype.lengthOfServiceValue = function (filterSelector) {
     var that = this;
     // Fetching the Employee length of service from filter values.
     var employeeLengthService = $(filterSelector).find('span:first').text();
     if ($.isNumeric(employeeLengthService)) {
-      $(filterSelector).find('span:first').text(that.formatLengthsOfService(employeeLengthService));
+      $(filterSelector).find('span:first').text(that.formatLengthOfService(employeeLengthService));
     }
   };
 
@@ -246,13 +246,14 @@
    * Processes results in people report view to format length of service for each
    * contact in a human readable form using moment lib.
    *
-   * @param {string} employeeLengthService - Employee length of service field value.
+   * @param {String} employeeLengthService - Employee length of service field value.
    *
-   * @returns {string} Date format string for provide length of service.
+   * @returns {String} Date format string for provide length of service.
    */
-  HRReport.prototype.formatLengthsOfService = function (employeeLengthService) {
+  HRReport.prototype.formatLengthOfService = function (employeeLengthService) {
     var dateEnd = moment();
     var dateStart = moment().subtract(employeeLengthService, 'days');
+
     if (!dateStart || !dateEnd) {
       return null;
     }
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/js/reports.js
```diff
@@ -551,7 +545,9 @@
             CRM.$('.report-config-select').append('<option value="' + data['id'] + '">' + data['label'] + '</option>');
             // Sort options by their labels alphabetically.
             CRM.$('.report-config-select').append(CRM.$('.report-config-select option').remove().sort(function (a, b) {
-              var at = $(a).text(), bt = $(b).text();
+              var at = $(a).text();
+              var bt = $(b).text();
```
bmnepali: What does `at` and `bt` mean here?  Can we rename them to more meaningful name?

civihr_employee_portal/js/reports.js
```diff
@@ -223,42 +220,39 @@
...
+   * @param {string} employeeLengthService - Employee length of service field value.
    *
    * @returns {string} Date format string for provide length of service.
    */
-  HRReport.prototype.formatLengthsOfService = function (employee_length_service) {
+  HRReport.prototype.formatLengthsOfService = function (employeeLengthService) {
     var dateEnd = moment();
-    var dateStart = moment().subtract(employee_length_service, 'days');
+    var dateStart = moment().subtract(employeeLengthService, 'days');
     if (!dateStart || !dateEnd) {
```
bmnepali: Leave one empty line after `var`.

civihr_employee_portal/js/reports.js
```diff
@@ -223,42 +220,39 @@
...
    * Processes results in people report view to format length of service for each
    * contact in a human readable form using moment lib.
    *
-   * @param {string} employee_length_service - Employee length of service field value.
+   * @param {string} employeeLengthService - Employee length of service field value.
    *
    * @returns {string} Date format string for provide length of service.
    */
-  HRReport.prototype.formatLengthsOfService = function (employee_length_service) {
+  HRReport.prototype.formatLengthsOfService = function (employeeLengthService) {
```
bmnepali: `formatLengthsOfService` or `formatLengthOfService` ?

civihr_employee_portal/js/reports.js
```diff
@@ -223,42 +220,39 @@
   /*
    * Finding the filter box for field Employee length of service.
    *
-   *  @param {string} filter_data
+   *  @param {string} filterData
```
bmnepali: use `String` instead of `string`.

civihr_employee_portal/js/reports.js
```diff
@@ -223,42 +220,39 @@
...
-    return false;
+  HRReport.prototype.lengthOfServiceFilter = function (filterData) {
+    return $(filterData).find('h4').text().indexOf('Employee length of service (') >= 0;
   };
 
   /*
    * Finding and replacing the value of Employee length of service.
    *
-   *  @param {string} filter_selector - DOM element that has value.
+   *  @param {string} filterSelector - DOM element that has value.
```
bmnepali: use `String` instead of `string`

civihr_employee_portal/js/reports.js
```diff
@@ -223,42 +220,39 @@
...
+      $(filterSelector).find('span:first').text(that.formatLengthsOfService(employeeLengthService));
     }
   };
 
   /**
    * Processes results in people report view to format length of service for each
    * contact in a human readable form using moment lib.
    *
-   * @param {string} employee_length_service - Employee length of service field value.
+   * @param {string} employeeLengthService - Employee length of service field value.
```
bmnepali: Same as above

civihr_employee_portal/js/reports.js
```diff
@@ -579,7 +575,6 @@
       return false;
     }
     var reportName = this.reportName;
-    var configId = this.getReportConfigurationId();
```
bmnepali: Ware you sure you wont need it?

deb1990:
```
var configId = this.getReportConfigurationId();
if (!configId) {
  swal('No configuration selected', 'Please choose configuration to delete.', 'error');
  return false;
}
var reportName = this.reportName;
var configId = this.getReportConfigurationId();
```
Line 1 and 7 is exactly same, so we were initialising the same var with same value twice. And between the 2 lines, no other line is changing the value of `getReportConfigurationId()`, so it does not make sense to keep both lines.

civihr_employee_portal/js/reports.js
deb1990: Changed to `aText` and `bText`

PR #393, opened 2017-11-13
PCHR-2850: Add Delete button

deb1990:
## Overview
As part of this PR, the Delete button is added to the Emergency Contacts and Dependents View.

It also adds a filter to the view to only show emergency contacts related to the current logged in contact.

## Comments
For more details refer https://github.com/compucorp/civihr-employee-portal-theme/pull/273

---

### DIFF
--- a/civihr_employee_portal/views/views_export/views_emergency_contacts.inc
+++ b/civihr_employee_portal/views/views_export/views_emergency_contacts.inc
@@ -86,11 +86,11 @@ $handler->display->display_options['fields']['dependant_s__92']['label'] = '';
 $handler->display->display_options['fields']['dependant_s__92']['exclude'] = TRUE;
 $handler->display->display_options['fields']['dependant_s__92']['element_label_colon'] = FALSE;
 /* Field: Global: Custom text */
-$handler->display->display_options['fields']['nothing']['id'] = 'nothing';
-$handler->display->display_options['fields']['nothing']['table'] = 'views';
-$handler->display->display_options['fields']['nothing']['field'] = 'nothing';
-$handler->display->display_options['fields']['nothing']['label'] = '';
-$handler->display->display_options['fields']['nothing']['element_label_colon'] = FALSE;
+$handler->display->display_options['fields']['emergency_contact_deletion']['id'] = 'something';
+$handler->display->display_options['fields']['emergency_contact_deletion']['table'] = 'views';
+$handler->display->display_options['fields']['emergency_contact_deletion']['field'] = 'nothing';
+$handler->display->display_options['fields']['emergency_contact_deletion']['label'] = '';
+$handler->display->display_options['fields']['emergency_contact_deletion']['element_label_colon'] = FALSE;
 /* Filter criterion: Global: Combine fields filter */
 $handler->display->display_options['filters']['combine']['id'] = 'combine';
 $handler->display->display_options['filters']['combine']['table'] = 'views';
--- a/civihr_employee_portal/views/views_export/views_emergency_contacts.inc
+++ b/civihr_employee_portal/views/views_export/views_emergency_contacts.inc
@@ -85,11 +85,13 @@ $handler->display->display_options['fields']['dependant_s__92']['field'] = 'depe
 $handler->display->display_options['fields']['dependant_s__92']['label'] = '';
 $handler->display->display_options['fields']['dependant_s__92']['exclude'] = TRUE;
 $handler->display->display_options['fields']['dependant_s__92']['element_label_colon'] = FALSE;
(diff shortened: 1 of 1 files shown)


bmnepali:
@deb1990  and @mickadoo  approved from my side. But these are php changes, so i would suggest to get it reviewed from a backend dev as well.

civihr_employee_portal/views/views_export/views_emergency_contacts.inc
```diff
@@ -85,6 +85,22 @@ $handler->display->display_options['fields']['dependant_s__92']['field'] = 'depe
 $handler->display->display_options['fields']['dependant_s__92']['label'] = '';
 $handler->display->display_options['fields']['dependant_s__92']['exclude'] = TRUE;
 $handler->display->display_options['fields']['dependant_s__92']['element_label_colon'] = FALSE;
+/* Field: Global: Custom text */
```
tunbola: Perhaps we should add some comment to describe this field more and that it is used to place a delete button on the Emergency Contacts and Dependents View also because most of the field values are not descriptive.
I didn't get the true picture until I saw the GIF on the other PR in the description.

civihr_employee_portal/views/views_export/views_emergency_contacts.inc
```diff
@@ -85,6 +85,22 @@ $handler->display->display_options['fields']['dependant_s__92']['field'] = 'depe
...
+$handler->display->display_options['fields']['emergency_contact_deletion']['label'] = '';
+$handler->display->display_options['fields']['emergency_contact_deletion']['element_label_colon'] = FALSE;
+/* Contextual filter: CiviCRM Custom: Emergency Contacts: Entity ID */
+$handler->display->display_options['arguments']['entity_id']['id'] = 'entity_id';
+$handler->display->display_options['arguments']['entity_id']['table'] = 'civicrm_value_emergency_contacts_21';
+$handler->display->display_options['arguments']['entity_id']['field'] = 'entity_id';
+$handler->display->display_options['arguments']['entity_id']['default_action'] = 'default';
+$handler->display->display_options['arguments']['entity_id']['default_argument_type'] = 'php';
+$handler->display->display_options['arguments']['entity_id']['default_argument_options']['code'] = 'return \\CRM_Core_Session::getLoggedInContactID();';
+$handler->display->display_options['arguments']['entity_id']['summary']['number_of_records'] = '0';
```
tunbola: Do we really need the summary part for the for adding the arguments? Will the code still work if the section is  removed?

mickadoo: Probably a lot of the auto exported stuff could be removed, but then it will likely be re-added the next time someone exports the view. Or do you mean there's some way to disable exporting this from the views UI?

civihr_employee_portal/views/views_export/views_emergency_contacts.inc
mickadoo: For auto-exported stuff I don't think adding comments is really a good option because it will probably be overwritten by the next person to do an export. I'll check if we can add some description field to the export that will be imported (and included in the export) or maybe just make the field ID clearer about what it does.

civihr_employee_portal/views/views_export/views_emergency_contacts.inc
tunbola: Not really, was just curious. We can leave as is then since its auto exported.

civihr_employee_portal/views/views_export/views_emergency_contacts.inc
tunbola: Either of the suggested options is fine. A description field seems better though.

civihr_employee_portal/views/views_export/views_emergency_contacts.inc
```diff
@@ -138,9 +156,10 @@ $translatables['emergency_contacts'] = array(
   t('Address'),
   t('Relationship with Employee'),
   t('Notes'),
+  t('<!-- This field is totally empty, but uses markup from civihr-employee-portal-theme to display the button. See civihr_default_theme/templates/view/views-view-field--emergency-contacts--non-dependant-emergency-contact--emergency_contact_deletion.tpl.php -->'),
+  t('All'),
   t('Non Dependant Emergency Contacts'),
   t('A list of emergency contacts where the "dependant" flag is false'),
-  t('non_dependant_emergency_contact'),
```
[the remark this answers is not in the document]
mickadoo: That's what the auto exporter tells me, it's been saying that for the last few exports but I ignored it until now. I don't know why it was in the translatables in the first place since it's just the display ID

PR #394, opened 2017-11-15
PCHR-2927: My Details Contract Filter Fix

mickadoo:
## Overview

The "my details" block uses filters for job contract and revision. These filters are problematic because if they don't match then nothing will be shown in the view at all.

## Before

The "my details" block used contextual filters for job contract and revision. Users without contracts only in the past or future would see no data at all on the "My Details" block on `/dashboard`.

image

## After

The contextual filters are gone, and the "My Details" block on `/dashboard` will always show basic info, as well as contract and role data for the current contract _if_ it is available.

image

## Notes

I also moved some code out of the module file into a separate class to represent each view we have. This should help keep the module file smaller and improve encapsulation.

## Technical Details

This uses the `hook_views_query_alter` hook to change the query. By limiting the join on the contract revision table to join only on the current contract we can hide results for inactive contracts.

If there is no current contract an impossible join (jobcontact_id = 0) will be added:

```
LEFT JOIN {hrjc_revision} hrjc_revision_civicrm_contact 
ON civicrm_contact.id = hrjc_revision_civicrm_contact.contact_id 
AND (hrjc_revision_civicrm_contact.jobcontract_id = 0)
```

To restrict the view data to only the current contract revision it adds a conditional to the join. It would be a lot easier to do this using revision ID, but revision ID is not present in the view table. Instead of this it selects by details and role revision IDs to find a single revision record to join on:

```
LEFT JOIN {hrjc_revision} hrjc_revision_civicrm_contact 
ON civicrm_contact.id = hrjc_revision_civicrm_contact.contact_id 
AND (hrjc_revision_civicrm_contact.jobcontract_id = 22 
AND hrjc_revision_civicrm_contact.role_revision_id = 24 
AND hrjc_revision_civicrm_contact.details_revision_id = 25)
```

This ensures the view will always return data, but fields from inactive contracts will not be shown.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/src/View/AbstractView.php
+++ b/civihr_employee_portal/src/View/AbstractView.php
@@ -2,15 +2,25 @@
 
 namespace Drupal\civihr_employee_portal\View;
 
+/**
+ * Contains methods common to views. Extend to encapsulate functionality
+ * specific to certain views, such as what code should be executed for certain
+ * view hooks.
+ */
 abstract class AbstractView {
 
+  /**
+   * This should be defined in child classes.
+   *
+   * @var string|NULL
+   */
   protected static $name = NULL;
 
   /**
    * @param \view $view
    * @param \views_plugin_query_default $query
    */
-  public function alter(&$view, &$query) {
+  public function alter($view, $query) {
     // do nothing
   }
 
--- a/civihr_employee_portal/src/View/AbstractView.php
+++ b/civihr_employee_portal/src/View/AbstractView.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Drupal\civihr_employee_portal\View;
+
+abstract class AbstractView {
+
+  protected static $name = NULL;
+
+  /**
+   * @param \view $view
+   * @param \views_plugin_query_default $query
+   */
+  public function alter(&$view, &$query) {
+    // do nothing
+  }
+
+  /**
+   * @return string
+   */
+  public static function getName() {
+    if (NULL === static::$name) {
+      throw new \Exception(sprintf('Please define $name in %s', static::class));
+    }
+
+    return static::$name;
+  }
+
+  /**
+   * Takes the where part of a query and returns an array of fields => values
+   *
+   * @param mixed $part
+   *  The $query->where part
+   * @param array $fields
+   *  An array to store the fields
+   */
+  public function getWhereFields($part, &$fields) {
+    if (isset($part['conditions'])) {
+      $part = $part['conditions'];
+    }
+    if (isset($part['field']) && $part['field'] instanceof \DatabaseCondition) {
+      $part = $part['field']->conditions();
+    }
+
+    if (isset($part['field'])) {
+      $fields[$part['field']] = $part['value'];
+    }
+    elseif (is_array($part)) {
+      foreach ($part as $item) {
+        $this->getWhereFields($item, $fields);
+      }
+    }
+  }
+}
--- a/civihr_employee_portal/src/View/MyDetailsView.php
+++ b/civihr_employee_portal/src/View/MyDetailsView.php
@@ -23,19 +23,22 @@ class MyDetailsView extends AbstractView {
     }
 
     $currentContactID = \CRM_Core_Session::getLoggedInContactID();
-    $contractID = $this->getCurrentContractID($currentContactID);
+    $revision = $this->getCurrentRevision($currentContactID);
 
     // If no current contract then add an impossible join condition
-    $filterID = $contractID ? $contractID : 0;
+    $contractID = $revision ? $revision['jobcontract_id'] : 0;
+    $roleRevisionID = $revision ? $revision['role_revision_id'] : 0;
+    $detailsRevisionID = $revision ? $revision['details_revision_id'] : 0;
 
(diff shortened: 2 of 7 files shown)


civihr_employee_portal/src/View/StaffDirectoryView.php
```diff
@@ -0,0 +1,28 @@
...
+
+  /**
+   * @var string
+   */
+  protected static $name = 'civihr_staff_directory';
+
+  /**
+   * @inheritdoc
+   */
+  public function alter(&$view, &$query) {
```
davialexandre: references are not necessary here as well

civihr_employee_portal/views/civihr_employee_portal.views.inc
```diff
@@ -1080,20 +1083,20 @@ function removeInactiveRoleDetails($results = []) {
...
+    default:
+      $viewClass = NULL;
+  }
 
-  // if department is set we must limit to only current roles
-  if (!empty($queryParts[$departmentField])) {
-    $group = $query->set_where_group('OR'); // add a new WHERE group
-    $query->add_where($group, $roleEndField, date('Y-m-d'), '>=');
-    $query->add_where($group, $roleEndField, NULL, 'IS');
+  if ($viewClass) {
```
davialexandre: this if can be avoided if return instead of setting `$viewClass` to NULL in default

civihr_employee_portal/tests/HelperFunctionsTest.php
```diff
@@ -32,7 +32,8 @@ class HelperFunctionsTest extends PHPUnit_Framework_TestCase {
 
   public function testGetWhereFields() {
     $parts = [];
-    getWhereFields($this->sampleWherePart, $parts);
+    $view = new \Drupal\civihr_employee_portal\View\MyDetailsView();
+    $view->getWhereFields($this->sampleWherePart, $parts);
```
davialexandre: This is confusing now, as this class is supposed to test Helper functions, but the test is checking a class method. Not only that, but a method of one specific implementation of an abstract class. What happens if I overwrite this method in `MyDetailsView`?

My suggestion here is:
- create a test class for the `AbstractView`
- move this test there
- to get an object with the method, you can mock the Abstract Class but don't mock the method itself, so the original one will be called

civihr_employee_portal/src/View/AbstractView.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Drupal\civihr_employee_portal\View;
+
+abstract class AbstractView {
```
davialexandre: Please add a description explaining why/when one would want to extend this class

civihr_employee_portal/src/View/AbstractView.php
```diff
@@ -0,0 +1,53 @@
...
+
+  protected static $name = NULL;
+
+  /**
+   * @param \view $view
+   * @param \views_plugin_query_default $query
+   */
+  public function alter(&$view, &$query) {
+    // do nothing
```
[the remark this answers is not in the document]
mickadoo: I thought about that, but it might be that you have a view that implements any one of the other [view hooks](https://api.drupal.org/api/views/views.api.php/group/views_hooks/7.x-3.x) but doesn't use `alter`, so I thought it best to leave it open. Then the method can always be called, so you could do something later like 

```
foreach ($allViewClasses as $viewClass) {
  $viewClass->alter($view, $query);
}
```

But only those that actually make changes would alter the view  or query.

civihr_employee_portal/src/View/MyDetailsView.php
```diff
@@ -0,0 +1,86 @@
...
+    $revision = $this->getCurrentRevision($currentContactID);
+
+    // If no current contract then add an impossible join condition
+    $contractID = $revision ? $revision['jobcontract_id'] : 0;
+    $roleRevisionID = $revision ? $revision['role_revision_id'] : 0;
+    $detailsRevisionID = $revision ? $revision['details_revision_id'] : 0;
+
+    $revisionTable = 'hrjc_revision';
+    $contactTable = 'civicrm_contact';
+    $revisionAlias = sprintf('%s_%s', $revisionTable, $contactTable);
```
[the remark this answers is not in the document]
mickadoo: Yeah, you're right. I was experimenting with putting where clauses on other tables so I had more variables here but after removing the others these two survived in this form

civihr_employee_portal/src/View/MyDetailsView.php
```diff
@@ -0,0 +1,86 @@
...
+   * @param int $contactID
+   *
+   * @return int|null
+   */
+  private function getCurrentRevision($contactID) {
+    $contract = civicrm_api3('HRJobContract', 'getcurrentcontract', [
+      'contact_id' => $contactID,
+    ]);
+
+    // God knows why but this result is a stdClass and count is always "1"
```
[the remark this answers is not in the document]
mickadoo: I was sure it was always returning count 1, but can't reproduce now.

civihr_employee_portal/views/civihr_employee_portal.views.inc
mickadoo: Nice

PR #396, opened 2017-11-20
PCHR-2949: Avoid injection of Shoreditch assets in SSP

AkA84:
## Overview
The assets of the `org.civicrm.shoreditch` extensions were leaking into the SSP, causing visual issues and also adding resources that the browser had to fetch for no reason

## Before
You can see the shoreditch style affecting the style of the SSP in the admin toolbar and in the header, for example
leak-before

Network tab:
ssp-assets-before

## After
Now the SSP style is not affected, because the shoreditch style is not included in the ssp anymore
leak-after

Network tab:
ssp-assets-after

### Fixed .module file style
The entire `civihr_employee_portal.module` was fixed style-wise. This includes:
* Indentation
* Following proper phpdoc style for function documentations
* Using `[]` instead of `array()`

There were more fixes that could have been done (for example grouping related functions together), but at least it's a start

### Remove original inclusion of pre-shoreditch theme
As part of [this PR](https://github.com/compucorp/civihr-employee-portal/pull/124), the pre-shoreditch theme `org.civicrm.bootstrapcivicrm` was included in the SSP just to style the datepickers in the Reports page.

```php
function civihr_employee_portal_init() {
  drupal_add_css(getModuleUrl('org.civicrm.bootstrapcivicrm') . 'css/bootstrap-civicrm-style.css');
}
```

Fortunately we don't need to do that anymore, so the inclusion was removed

### Refactor hook_civicrm_coreResourceList()
This hook was introduced in [this PR](https://github.com/compucorp/civihr-employee-portal/pull/327/files) and it definitely need some improvement

1. The code was not optimal
2. It would run also when the browser would load a civicrm page
3. The entire dependency-swapping was unnecessary
4. The notify plugin needed to be added only for users without access to CiviCRM, otherwise the plugin is [already served](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Resources.php#L748)

*Before*
```php
function civihr_employee_portal_civicrm_coreResourceList(&$list, $region) {
  $jsToRemove[] = array_search('js/Common.js', $list);
  $jsToRemove[] = array_search('js/crm.ajax.js', $list);
  $jsToRemove[] = array_search('js/wysiwyg/crm.wysiwyg.js', $list);
  $jsToRemove[] = array_search('js/crm.drupal.js', $list);

  foreach ($jsToRemove as $filepath) {
    unset($list[$filepath]);
  }

  $list[] = "packages/jquery/plugins/jquery.notify.js";
  $list[] = "js/Common.js";
  $list[] = "js/crm.ajax.js";
  $list[] = "js/wysiwyg/crm.wysiwyg.js";
  $list[] = "js/crm.drupal.js";
}
```

*After*
```php
function civihr_employee_portal_civicrm_coreResourceList(&$list, $region) {
  if (!_isCiviCRM()) {
    $notifyPath = 'packages/jquery/plugins/jquery.notify.min.js';

    if (!array_search($notifyPath, $list)) {
      $list[] = $notifyPath;
    }
  }
}
```

### Remove Shoreditch assets
A simple `_remove_shoreditch_resources` function was introduced, that was then used in the `hook_css_alter()` and `hook_js_alter()` hooks, nothing fancy and i'm not sure if there is a more elegant way to tackle this

```php
function civihr_employee_portal_css_alter(&$css) {
  if (_isCiviCRM()) {
    // ...
  } else {
    _remove_shoreditch_resources($css);
  }
}

function civihr_employee_portal_js_alter(&$javascript) {
  if (_isCiviCRM()) {
    // ...
  } else {
    _remove_shoreditch_resources($javascript);
  }
}

function _remove_shoreditch_resources(&$resourcesList) {
  foreach ($resourcesList as $path => $resource) {
    if (strpos($path, 'org.civicrm.shoreditch')) {
      unset($resourcesList[$path]);
    }
  }
}
```

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -415,28 +415,6 @@ function _rebuild_view($viewName) {
 }
 
 /**
- * Get the module URL
- *
- * @param  string $moduleName <Module name>
- *
- * @return string             <Module url>
- */
-function getModuleUrl($moduleName) {
-  global $base_url;
-
-  $path = CRM_Core_Resources::singleton()->getUrl($moduleName);
-
-  // Need to do this to avoid some configurations issues
-  // Some environments return /sites/all/modules/civicrm
-  // And others return http://localhost:8900/sites/all/modules/civicrm
-  if (strstr($path, "$base_url/")) {
-    $path = explode("$base_url/", $path)[1];
-  }
-
-  return ltrim($path, '/');
-}
-
-/**
  * Implements hook_coreResourceList
  *
  * @param array $list
@@ -479,7 +457,6 @@ function civihr_employee_portal_init() {
 
     // Add our additional css libraries
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/css/custom.css");
-    drupal_add_css(getModuleUrl('org.civicrm.shoreditch') . 'css/bootstrap.css');
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/lib/sweetalert/sweetalert.css");
     drupal_add_css(drupal_get_path('module', 'civihr_employee_portal') . "/lib/tablesaw/tablesaw.css");
 
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -10,13 +10,9 @@ use Drupal\civihr_employee_portal\Forms\ContactForm;
  * Implements hook_css_alter().
  */
 function civihr_employee_portal_css_alter(&$css) {
-
-  // Remove defaults system messages css file.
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -77,6 +81,20 @@ function _remove_resources(&$resourcesList, $blacklist = []) {
   }
 }
 
+/**
+ * Removes any resource related to the shoreditch extension
+ *
+ * @param array &$resourcesList variable contains all the references to loaded
```
omarabuhussein: the param description should be in a new line like this : 

```php
/**
 * Removes any resource related to the shoreditch extension
 *
 * @param array &$resourcesList
 *   variable contains all the references to loaded
 *   js/css array from alter.
 */
```

PR #406, opened 2017-12-04
PCHR-3004: Webform Portability

mickadoo:
## Overview

Webforms that used CiviCRM components are not very portable. Some problems are the use of hardcoded custom group, custom field and location type IDs. This means that importing a webform on another machine can have lead to unexpected settings, such as a webform saving an email or phone with the wrong location type. The problem with custom field and group ID was noticed early in the Onboarding feature QA and a fix was made for it. This ticket will fix location type ID portability.

In addition when we make changes to a webform it tends to be difficult to release these changes. One way is by creating an upgrader function to load, update, and save the webform. However a single drush command to refresh all webforms to match their current state in the export files would be a lot easier. There is a drush command, but it will recreate new nodes each time it is run.

## Before

- Webforms did not include location type metadata when they were exported. Importing them could mean location type pointed to an incorrect option.
- Making changes to a webform was laborious.

## After

- Webforms includes location type metadata when they are exported. This metadata is used to match location types to the correct ID on the instance they are imported to.
- Making changes to a webform is easier with a single drush command. Running it will update existing nodes instead of creating new ones each time.

peek 2017-12-05 11-42

## Technical Details

The process for improving portability is to take a snapshot of the location type mapping on the developer's machine. This snapshot is included in the export file. An example would be 

```
    'customMapping' => array(
      'locationTypes' => array(
        1 => 'Home',
      ),
    ),
```

When importing the importer will check the mapping and lookup the current ID for the location type with that name. For example if on the new instance the "Home" location type has an ID of 3 it will replace all `"location_type_id": 1` with `"location_type_id": 3` before importing the webform.

## Notes 

- An warning from CiviCRM about a missing index in the contact image_URL is also fixed in this ticket.

---

- [x] Tests Pass


civihr_employee_portal/src/Webform/WebformTransferConvertor.php
```diff
@@ -0,0 +1,19 @@
...
+interface WebformTransferConvertor {
+
+  /**
+   * @param \stdClass $node
+   * @return void
+   */
+  public static function preExport(\stdClass $node);
+
+  /**
+   * @param \stdClass $node
```
tunbola: Please leave an empty line before @return tag for both functions in this class.

civihr_employee_portal/src/Webform/LocationTypeIDConvertor.php
```diff
@@ -0,0 +1,88 @@
...
+        $locationTypeID = $oldToNewMapping[$locationTypeID];
+      }
+    }
+  }
+
+  /**
+   * Gets all location type IDs for a webform.
+   * Results are returned by reference to allow easier modification.
+   *
+   * @param \stdClass $node
```
tunbola: Please leave an empty line before @return tag for all functions in this class.

civihr_employee_portal/src/Webform/LocationTypeIDConvertor.php
```diff
@@ -0,0 +1,88 @@
...
+   * @return array
+   */
+  private static function getLocationTypeIDsByRef(\stdClass $node) {
+    $locationTypeIDs = [];
+
+    $civicrmData = &$node->webform_civicrm;
```
tunbola: Do we need to pass most of the variables in this function by reference? Since we are not modifying the variables directly but just adding to the `$locationTypeIDs` array, I don't think this is necessary.
Or what do you think?

civihr_employee_portal/src/Webform/LocationTypeIDConvertor.php
```diff
@@ -0,0 +1,88 @@
...
+    $params = ['id' => ['IN' => $locationTypeIDs]];
+    $locationTypes = civicrm_api3('LocationType', 'get', $params)['values'];
+    $locationTypes = array_column($locationTypes, 'name', 'id');
+    $node->customMapping[self::MAPPING_KEY] = $locationTypes;
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public static function preImport(\stdClass $node) {
```
tunbola: Would be good to add a description for the preImport and preExport functions rather than they inheriting from the generic doc.

civihr_employee_portal/src/Webform/WebformTransferService.php
```diff
@@ -0,0 +1,29 @@
...
+ * Handles adding metadata for fields that rely on the database IDs and then
+ * using the metadata to use the correct IDs on whichever system the webform is
+ * exported to.
+ */
+class WebformTransferService {
+
+  /**
+   * @param \stdClass $node
+   */
+  public static function preExport(\stdClass $node) {
```
tunbola: Any reason why these methods are not actual class methods including the methods in the `WebformTransferConvertor` interface? Static methods are somewhat ugly. :(

civihr_employee_portal/src/Webform/WebformTransferService.php
```diff
@@ -0,0 +1,29 @@
...
+    WebformExportCustomFieldConvertor::preExport($node);
+    LocationTypeIDConvertor::preExport($node);
+  }
+
+  /**
+   * @param \stdClass $node
```
tunbola: Please add brief descriptions to the methods in this class.

civihr_employee_portal/src/Forms/OnboardingWebForm.php
```diff
@@ -190,4 +191,30 @@ private function userCreatedAfterOnboardingReleased() {
     return $user->created > $onboardingRelease;
   }
 
+  /**
+   * Unfortunately for us the contact profile page will expect an image URL
+   * using the civicrm/file?photo=foo.jpg style. Since our contact images are
+   * created using webform they won't match this so to avoid the warnings about
+   * 'Undefined index: photo' we append photo=0 here. I need a shower.
```
tunbola: Please leave a line before @see comment.
You can remove the I need a shower comment too  :)

civihr_employee_portal/src/Forms/OnboardingWebForm.php
```diff
@@ -190,4 +191,30 @@ private function userCreatedAfterOnboardingReleased() {
...
+   * @see CRM_Utils_File::getImageURL
+   *
+   * @param int $contactID
+   */
+  private function hackyFixForImageURL($contactID) {
+    $params = ['return' => "image_URL", 'id' => $contactID];
```
tunbola: Please use single quotes here.

civihr_employee_portal/src/Helpers/NodeHelper.php
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Finds a single node
+   *
+   * @param array $criteria
+   *   The criteria for searching for the node
+   *
+   * @return \stdClass|null
+   *   The node, if found. NULL if not found or more than one exist.
+   */
+  public static function findOneBy($criteria = []) {
```
tunbola: The method name is a bit off from what it does, It sounds like it should return the first node it finds based on the criteria or NULL of it can't find any rather than returning NULL when it finds more than one.
Maybe a `findIfOnlyOne` or `getIfOnlyOne` or `findIfOnlyOneExists` or something similar will be a better name.

civihr_employee_portal/civihr_employee_portal.drush.inc
```diff
@@ -70,27 +71,41 @@ function civihr_employee_portal_drush_command() {
...
+
+      node_export_save($importNode);
+      variable_set($file->name . '_webform_nid', $importNode->nid);
+      $count++;
+    }
   }
 
-  // Log to the command line with an OK status
-  drush_log('Node Webform refresh finished', 'ok');
+   drush_log(sprintf('%d nodes were updated', $count), 'ok');
```
tunbola: Wondering if it would it help if we knew the count of those that were updated and those that were newly created?
In that case we would need to set two count variables, one for each scenario.

civihr_employee_portal/civihr_employee_portal.drush.inc
```diff
@@ -100,6 +100,14 @@ function drush_civihr_employee_portal_refresh_node_export_files() {
       }
 
       node_export_save($importNode);
+
+      // Need to manually update webform_civicrm_form record
+      if ($importNode->type === 'webform' && !$importNode->is_new) {
```
tunbola: This change needs to happen only when the Imported node is not new?

civihr_employee_portal/tests/Webform/LocationTypeIDConvertorTest.php
```diff
@@ -0,0 +1,73 @@
...
+  protected function setUp() {
+    $result = civicrm_api3('LocationType', 'create', ['name' => 'Test']);
+    $this->locationType = array_shift($result['values']);
+  }
+
+  protected function tearDown() {
+    civicrm_api3('LocationType', 'delete', ['id' => $this->locationType['id']]);
+  }
+
+  public function testEmptyMapping() {
```
tunbola: This should probably be something like `testLocationTypeMappingIsNotEmpty` or something.
The current test name gives the impression that we would be asserting that that the mapping should be empty.

civihr_employee_portal/tests/Webform/LocationTypeIDConvertorTest.php
```diff
@@ -0,0 +1,73 @@
...
+    $node = new \stdClass();
+    $sampleLocationType = $this->locationType;
+    $correctID = $sampleLocationType['id'];
+    $wrongID = $correctID + 500;
+
+    $node->webform_civicrm = $this->getSampleCiviCRMWebform($correctID);
+
+    // Check that export will add the correct mapping
+    LocationTypeIDConvertor::preExport($node);
+    $mapping = &$node->customMapping[LocationTypeIDConvertor::MAPPING_KEY];
```
tunbola: Do we to pass by reference here?

tunbola:
In the after section:
s/Webforms location type metadata/Webforms includes location type metadata

civihr_employee_portal/civihr_employee_portal.drush.inc
mickadoo: I've tested it and when it is new it gets saved as expected. I think it might be a problem with webform_civicrm module where the same code is not run on update.

civihr_employee_portal/src/Helpers/NodeHelper.php
mickadoo: Good point, maybe I'll just change it to returning the first it finds, or null if it doesn't find any. I was trying to copy [Doctrine's method](http://www.doctrine-project.org/api/orm/2.5/source-class-Doctrine.ORM.EntityRepository.html#191) and they seem to return just the first or NULL.

civihr_employee_portal/src/Webform/LocationTypeIDConvertor.php
mickadoo: This isn't required for the first call to this, but when actually changing the values from `preImport` passing by reference saves repetition (since the same method can be used for just fetching the values) and avoids having to directly change the values like

`$node->civicrm_webform['data']['Contact'][0]['phone'][0]['location_type_id'] = $newID`

civihr_employee_portal/src/Webform/WebformTransferService.php
mickadoo: I would like to use non-static class methods, but since this is the employee portal I don't have any DI container available. It's also messy instantiating classes everywhere, especially if the dependencies change later. I'm not sure how to get around this, in CiviHR we can use the Symfony container, but just not here.

civihr_employee_portal/tests/Webform/LocationTypeIDConvertorTest.php
mickadoo: Again it's just avoiding long lines, 

```
$node->customMapping[LocationTypeIDConvertor::MAPPING_KEY][$wrongID] = $sampleLocationType['name'];
unset($node->customMapping[LocationTypeIDConvertor::MAPPING_KEY][$correctID];
```

civihr_employee_portal/src/Webform/WebformTransferConvertor.php
```diff
@@ -2,16 +2,28 @@
 
 namespace Drupal\civihr_employee_portal\Webform;
 
+/**
+ * Defines some methods that should be implements for the alteration of node
```
tunbola: s/be implements/be implemented

PR #410, opened 2017-12-13
PCHR-2599: Fix Tests

mickadoo:
## Overview

Although tests in CiviHR Employee Portal are running, tests in Tasks and Assignments are not because TASettings are fetched in an `init` function of this module.

## Before

T&A tests would not run because of an error related to fetching TASettings.

## After

TASettings are only fetched if T&A is enabled. Tests in T&A can run.

---

- [x] Tests Pass


mickadoo:
As this isn't really part of PCHR-2599 I created a separate ticket for this, closing this PR in favor of https://github.com/compucorp/civihr-employee-portal/pull/411

PR #411, opened 2017-12-13
PCHR-3053: Fix T&A Tests

mickadoo:
## Overview

Although tests in CiviHR Employee Portal are running, tests in Tasks and Assignments are not because TASettings are fetched in an `init` function of this module.

## Before

T&A tests would not run because of an error related to fetching TASettings.

## After

TASettings are only fetched if T&A is enabled. Tests in T&A can run.

## Technical Details

The tests in T&A were failing because `civihr_employee_portal_init` was being called when the T&A extension was not enabled. `civihr_employee_portal_init` calls `_load_ta_settings`, which does an API request to fetch `TASettings`. The `TASettings` API is provided by T&A so this call failed.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -5044,7 +5044,7 @@ function civihr_employee_portal_my_details() {
 function civihr_employee_portal_hrreport_landing_page() {
   $jsOptions = ['type' => 'file', 'scope' => 'footer'];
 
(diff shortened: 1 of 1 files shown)


mickadoo:
@davialexandre PCHR-3027 doesn't have a branch for civihr_employee_portal, so I thought it wouldn't make sense to create feature and epic branches in this repo just to make an unrelated fix.

I've updated the technical details section to provide more info on the cause.

PR #413, opened 2017-12-15
PCHR-3032: Fix dependencies names

AkA84:
Simply fixes names of modules and services, due to the changes done in [this PR](https://github.com/compucorp/civihr-tasks-assignments/pull/291)

### DIFF
--- a/civihr_employee_portal/js/ta-documents-app.js
+++ b/civihr_employee_portal/js/ta-documents-app.js
@@ -1,7 +1,7 @@
 /* globals angular */
 
 (function (angular) {
-  angular.module('taDocuments', ['civitasks.appDocuments', 'civitasks.directives'])
+  angular.module('taDocuments', ['tasks-assignments.documents', 'tasks-assignments.directives'])
     .config(function ($urlRouterProvider, $locationProvider) {
       $locationProvider.html5Mode(true); // This is required to remove # for the URL
       $urlRouterProvider.otherwise('/tasks-and-documents');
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/js/ta-documents-app.js
```diff
@@ -1,17 +1,17 @@
...
+  angular.module('taDocuments', ['tasks-assignments.documents', 'tasks-assignments.directives'])
     .config(function ($urlRouterProvider, $locationProvider) {
       $locationProvider.html5Mode(true); // This is required to remove # for the URL
       $urlRouterProvider.otherwise('/tasks-and-documents');
     })
     .controller('ModalController', ModalController);
 
   ModalController.$inject = ['$scope', '$rootScope', '$window', '$rootElement', '$log', '$uibModal',
-    'DocumentService', 'FileService', 'config', 'settings', 'DateFormat'];
+    'documentService', 'fileServiceTA', 'config', 'settings', 'DateFormat'];
```
igorpavlov: Why `fileServiceTA` and not `fileService`? Is there a specific file service for T&A?

AkA84:
Rebase, forgot to rename `DocumentService` into `documentService`

igorpavlov:
@AkA84 please let know how have you discovered that the name was forgotten to be renamed? Has a "search by text" (grep or similar) check been performed? Is there any chance other things could be forgotten to be renamed?

AkA84:
@igorpavlov QA reported a bug cause by the `DocumentService` variable not being defined. No, the entire app is just one file. Same as before, it's unlikely because it's just one file, i also tried to reproduce the bug reported by QA after my fix and I couldn't

igorpavlov:
@AkA84 I have also grepped and only one file was found. I think we should grep across the **whole site** when we rename a service, module etc. Just a note for the future.

PR #414, opened 2018-01-03
PCHR-3040: Report layout update

reneolivo:
## Overview
This PR updates the layout of the report page so it follows the mock detailed in the [Report Improvements Wiki](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/123699243/Reporting+User+Experience+Improvements).

image

## Before
anim

## After
anim

### Replacing the layout
To replace the layout DOM manipulation was needed since the report is generated by the [Pivot Table plugin](https://github.com/nicolaskruchten/pivottable) which doesn't support custom templates.

The template is first appended inside the pivot table container:

```js
HRReport.prototype.createReportSectionElement = function () {
  var html = '<div class="report-section">' +
    '<div class="row report-header-section">' +
      '<div class="report-filters col-sm-3"></div>' +
      '<div class="report-function col-sm-2">' +
        '<div class="report-function-select"></div>' +
        '<div class="report-function-group"></div>' +
      '</div>' +
      '<div class="report-field-columns col-sm-7"><table><tr></tr></table></div>' +
    '</div>' +
    '<div class="row report-content-section">' +
      '<div class="report-fields-selection col-sm-3"><table><tr></tr></table></div>' +
      '<div class="report-field-rows col-sm-2"><table><tr></tr></table></div>' +
      '<div class="report-area col-sm-7"></div>' +
    '</div>' +
  '</div>';

  this.pivotTableContainer.append(html);
};
```

And then each report section is moved to its new container:

```js
HRReport.prototype.updateCustomTemplate = function () {
  var hasReportSectionElement = this.pivotTableContainer.find('.report-section').length;

  if (!hasReportSectionElement) {
    this.createReportSectionElement();
    this.moveReportElements(); // <---- here each section is moved to its new container
    this.appendFilters();
    this.bindFilters();
    this.appendFieldsDraggingInstructions();
    this.bindDragAndDropEventListeners();
  }

  this.updateDropdown();
  this.updateFilterbox();
};
```

To move the elements, we detatch them from the DOM and then append them to the new container using this function:

```js
HRReport.prototype.moveReportElementFromTo = function (fromSelector, toSelector) {
  var fromElement = this.pivotTableContainer.find(fromSelector);
  var toElement = jQuery(toSelector);

  if (!fromElement.length) {
    return;
  }

  toElement.empty();
  fromElement.detach().appendTo(toElement);
};
```

### Fields selection scroll

To reduce the height of the report a vertical scroll was added to the fields selection:

```css

#reportPivotTable .report-fields-selection table {
  display: block; /* needs to be set as block or scrolls won't be displayed */
  width: 100%;
  max-height: 50vh;
  overflow-x: hidden;
  overflow-y: auto;
}
```

### Filters

The filters are generated outside of the pivot tablet element and are initialized using Angular. Angular is used to display the date pickers, set the initial date, format the date, and for showing/hiding the filter panel (which has been removed in the new layout).

Detaching and attaching angular elements somewhere else or cloning them just doesn't work (the date picker stops working for example). To be able to place the filters inside their new container, the Angular `$compile` function was wrapped into a `compileAngularElement` function:

```js
angularModule.run(['$compile', '$rootScope', function ($compile, $rootScope) {
  compileAngularElement = function (html) {
    var element = angular.element(html);
    var scope = $rootScope.$new();

    $compile(element)(scope);
    return element;
  };
}]);
``` 

The function is then used for compiling the filter's HTML like this:

```js
HRReport.prototype.appendFilters = function () {
  var element = this.originalFilterElement.clone();
  var filtersContainer = this.pivotTableContainer.find('.report-filters');

  compileAngularElement(element);
  filtersContainer.empty();
  element.appendTo(filtersContainer);
};
```

### Date Picker fixes

The date picker component's style, when inside the `#pivotReport` element, inherits a few rules that are not desired.

For example, an extra padding added by the SSP Theme [here](https://github.com/compucorp/civihr-employee-portal-theme/blob/staging/civihr_default_theme/scss/layout/_table.scss#L11). That padding is useful for the report, fields, etc. but not for the date picker.

A value of `7px 0` was applied instead. This came for the default value the date picker would have taken if not for the rule stated above.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
    <td></td>
    <td></td>
</table>

Also, the z-index of select elements was changed from 2 to 0 so they don't display over the date picker.

```css
#civihrReports .crm_custom-select > select {
  z-index: 0;
}
```

### Fields instructions and highlighting

Instructions for drag and dropping fields onto column and row containers was added via Javascript since the target containers are generated by the pivot table plugin.

The instructions are appended as follows:

```js
HRReport.prototype.appendFieldsDraggingInstructions = function () {
  var columnsMessage = 'Drag and drop a field here from the list on the left to add as a column heading / horizontal axis in the report.';
  var rowsMessage = 'Drag and drop a field here from the list on the left to add as a row heading / vertical axis in the report.';

  $('.report-field-columns .pvtAxisContainer').append('<p class="instructions">' + columnsMessage + '</p>');
  $('.report-field-rows .pvtAxisContainer').append('<p class="instructions">' + rowsMessage + '</p>');
};
```

Drag and drop even listeners were added for highlighting the droppable areas and also to remove the instructions once fields have been added to them:

```js
HRReport.prototype.bindDragAndDropEventListeners = function () {
  var draggableItems = this.pivotTableContainer.find('.report-fields-selection .pvtAxisContainer, .report-field-columns .pvtAxisContainer, .report-field-rows .pvtAxisContainer');
  var droppableContainers = this.pivotTableContainer.find('.report-field-columns .pvtAxisContainer, .report-field-rows .pvtAxisContainer');
  var highlightClass = 'highlight';

  draggableItems.on('sortstart', function () {
    droppableContainers.addClass(highlightClass);
  });

  draggableItems.on('sortstop', function (event) {
    var targetContainer = $(event.toElement).parents('.pvtAxisContainer');

    droppableContainers.removeClass(highlightClass);
    targetContainer.find('.instructions').slideUp();
  });
};
```

### jQuery reference

Most references to `jQuery` and `CRM.$` were renamed to `$` since they are synonymous. Only references to pivot table renderers were left using the jQuery variable:

```js
this.pivotTableContainer.pivotUI(this.data, {
  // ...
  renderers: $.extend(
    jQuery.pivotUtilities.renderers,
    jQuery.pivotUtilities.c3_renderers,
    jQuery.pivotUtilities.export_renderers
  ),
  // ...
});
``` 

## Comments

This PR doesn't cover the following:

* Mobile or Small Screen view.
* Auto selecting a report configuration.

These will be worked in other Tickets/PRs.

The changes were tested manually since the report page changes completely.

### DIFF
--- a/civihr_employee_portal/css/custom.css
+++ b/civihr_employee_portal/css/custom.css
@@ -300,7 +300,7 @@ span.appraisals-employee-legend {
 }
 
 #reportPivotTable .report-field-rows table {
-  min-height: 80px;
+  min-height: 50vh;
 }
 
 #reportPivotTable .pvtRendererArea {
--- a/civihr_employee_portal/css/custom.css
+++ b/civihr_employee_portal/css/custom.css
@@ -8,7 +8,8 @@ and open the template in the editor.
     Author     : gergelymeszaros
 */
 
-.view-absence-list .views-field, .view-absence-list th {
+.view-absence-list .views-field,
+.view-absence-list th {
   width: 15%;
 }
 
@@ -34,7 +35,8 @@ and open the template in the editor.
   }
 }
 
-.table-up, .table-down {
+.table-up,
+.table-down {
   color: #007;
   cursor: pointer;
 
@@ -180,7 +182,8 @@ span.appraisals-employee-legend {
   margin-right: 10px;
 }
 
-#reportPivotTable .report-filters, #reportPivotTable .report-function {
+#reportPivotTable .report-filters,
+#reportPivotTable .report-function {
   position: relative;
 }
 
@@ -206,7 +209,8 @@ span.appraisals-employee-legend {
   padding-top: 20px;
 }
 
-#civihrReports .report-block, .report-filters:before, .report-function,
+#civihrReports .report-block,
+#reportPivotTable .report-function,
 #reportPivotTable #report-filters,
 #reportPivotTable .report-fields-selection table,
 #reportPivotTable .report-field-rows table,
--- a/civihr_employee_portal/css/custom.css
+++ b/civihr_employee_portal/css/custom.css
@@ -162,7 +162,6 @@ span.appraisals-employee-legend {
 }
 
 #reportPivotTableConfiguration {
-  border-bottom: 1px solid #ddd;
   padding: 10px 10px;
   margin-top: 12px;
 }
@@ -179,6 +178,68 @@ span.appraisals-employee-legend {
 
 .view-id-hr_resource_types_list .add-button-icon {
   margin-right: 10px;
+}
+
+#reportPivotTable .report-filters, .report-function {
+  position: relative;
+}
+
+#reportPivotTable .report-filters:before {
+  border: 1px solid #c3cbd6;
+  content: '';
+  display: block;
+  height: 100%;
+  left: 15px;
+  position: absolute;
+  top: 0;
+  width: calc(100% - 30px);
+}
+
+#reportPivotTable .report-function {
+  padding-top: 20px;
+}
+
+#civihrReports .report-block, .report-filters:before, .report-function,
+#reportPivotTable .pvtAxisContainer {
+  border: 1px solid #c3cbd6;
+}
+
+#reportPivotTable .row {
+  display: table;
+  margin-bottom: 15px;
+}
+
+#reportPivotTable .row:last-child {
+  margin-bottom: 0;
+}
+
+#reportPivotTable .row > [class*="col-"] {
+  display: table-cell;
+  float: none;
+  vertical-align: top;
+}
+
+#reportPivotTable .report-field-columns {
+  position: relative;
+}
+
+#reportPivotTable .report-field-columns table {
+  bottom: 0;
+  min-height: 80px;
+  position: absolute;
+  width: calc(100% - 30px);
+}
+
+#reportPivotTable .report-field-rows {
+  padding: 0;
+}
+
+#reportPivotTable .report-field-rows table {
+  min-height: 80px;
+}
+
+#reportPivotTable .pvtRendererArea {
+  padding: 0;
 }
 
 #footer {
--- a/civihr_employee_portal/css/custom.css
+++ b/civihr_employee_portal/css/custom.css
@@ -272,8 +272,8 @@ span.appraisals-employee-legend {
   display: block;
   width: 100%;
   max-height: 50vh;
-  overflow-x: hidden;
   overflow-y: auto;
+  overflow-x: hidden;
 }
 
 #reportPivotTable .report-fields-selection table li {
@@ -286,10 +286,9 @@ span.appraisals-employee-legend {
   height: auto;
   line-height: 1.2em;
   padding: 8px 20px 8px 15px;
-  padding-right: 20px;
   position: relative;
   white-space: normal;
-  width: 100%;
+  width: inherit;
 }
 
 #reportPivotTable .report-fields-selection table li span.pvtTriangle,
@@ -299,16 +298,13 @@ span.appraisals-employee-legend {
   right: 10px;
 }
 
-#reportPivotTable .report-field-rows {
+#reportPivotTable .report-field-rows,
+#reportPivotTable .pvtRendererArea {
   padding: 0;
 }
 
 #reportPivotTable .report-field-rows table {
   min-height: 50vh;
-}
-
-#reportPivotTable .pvtRendererArea {
-  padding: 0;
 }
 
 #reportPivotTable [uib-datepicker-popup-wrap] table {
--- a/civihr_employee_portal/css/custom.css
+++ b/civihr_employee_portal/css/custom.css
@@ -223,6 +223,26 @@ span.appraisals-employee-legend {
   border: 1px solid #c3cbd6;
 }
 
+#reportPivotTable .pvtAxisContainer.highlight {
+  background: #fbfbfc;
+  border-width: 2px;
+}
+
+#reportPivotTable .pvtAxisContainer::after {
+  content: 'Drop here';
+  line-height: 33px;
+  opacity: 0;
+  transition: opacity 0.4s ease;
+}
+
+#reportPivotTable .pvtAxisContainer.highlight::after {
+  opacity: 1;
+}
+
+#reportPivotTable .pvtAxisContainer.highlight .drop-here-message {
+  content: 'Drop Here';
+}
+
 #reportPivotTable .row {
   display: table;
   margin-bottom: 15px;
--- a/civihr_employee_portal/css/custom.css
+++ b/civihr_employee_portal/css/custom.css
@@ -318,6 +318,14 @@ span.appraisals-employee-legend {
   padding: 0;
 }
 
+#reportPivotTable [uib-datepicker-popup-wrap] table {
+  margin: 0;
+}
+
+#reportPivotTable table.uib-daypicker tbody td {
+  padding: 7px 0;
+}
+
 #footer {
   height: 250px;
   color: #727e8a;
--- a/civihr_employee_portal/css/custom.css
+++ b/civihr_employee_portal/css/custom.css
@@ -227,23 +227,12 @@ span.appraisals-employee-legend {
   border: none;
 }
 
-#reportPivotTable .highlight {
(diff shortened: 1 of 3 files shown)


civihr_employee_portal/css/custom.css
```diff
@@ -230,6 +230,29 @@ span.appraisals-employee-legend {
...
+
+#reportPivotTable .report-fields-selection table li span.pvtAttr,
+#reportPivotTable .report-field-rows table li span.pvtAttr {
+  height: auto;
+  line-height: 1.2em;
+  padding: 8px 20px 8px 15px;
+  padding-right: 20px;
+  position: relative;
+  white-space: normal;
+  width: 100%;
```
bmnepali: Since you have `width` defined here:
```css
#reportPivotTable .report-fields-selection table li {
  margin-bottom: 5px;
  width: 100%;
}
```
Here you can inherit the width (please check):
`width: inherit;`

civihr_employee_portal/css/custom.css
```diff
@@ -180,10 +180,19 @@ span.appraisals-employee-legend {
   margin-right: 10px;
 }
 
-#reportPivotTable .report-filters, .report-function {
+#reportPivotTable .report-filters, #reportPivotTable .report-function {
```
bmnepali: Keep separate selectors in separate line.

civihr_employee_portal/css/custom.css
```diff
@@ -269,6 +275,14 @@ span.appraisals-employee-legend {
   width: calc(100% - 30px);
 }
 
+#reportPivotTable .report-fields-selection table {
+  display: block;
+  width: 100%;
+  max-height: 50vh;
+  overflow-x: hidden;
+  overflow-y: auto;
```
bmnepali: List properties in alphabetical order.

civihr_employee_portal/css/custom.css
```diff
@@ -318,6 +318,14 @@ span.appraisals-employee-legend {
   padding: 0;
 }
 
+#reportPivotTable [uib-datepicker-popup-wrap] table {
+  margin: 0;
+}
+
+#reportPivotTable table.uib-daypicker tbody td {
+  padding: 7px 0;
```
bmnepali: Does this padding matched the padding in mockups?

civihr_employee_portal/css/custom.css
```diff
@@ -193,18 +193,6 @@ span.appraisals-employee-legend {
...
-  content: '';
-  display: block;
-  height: 100%;
-  left: 15px;
-  position: absolute;
-  top: 0;
-  width: calc(100% - 30px);
-  z-index: 0;
-}
-
```
bmnepali: Why was it removed?

civihr_employee_portal/js/reports.js
```diff
@@ -202,16 +202,6 @@
...
-  /**
-   * Make the report field rows container at least the same height as the field
-   * selection.
-   */
-  HRReport.prototype.makeReportFieldRowsContainerFullHeight = function () {
-    var fullHeight = this.pivotTableContainer.find('.report-fields-selection table').height();
-
-    this.pivotTableContainer.find('.report-field-rows table').css('min-height', fullHeight);
-  };
-
```
bmnepali: Why is this  function removed? Don't we need it any more?

bmnepali:
Also, **Test** section is missing the the PR.

civihr_employee_portal/css/custom.css
reneolivo: This padding is not part of the mocks. It was added to override a padding added by SSP Theme [here](https://github.com/compucorp/civihr-employee-portal-theme/blob/staging/civihr_default_theme/scss/layout/_table.scss#L11). That padding is useful for the report, fields, etc. but not for the date picker.

The value of `7px 0` came for the default value the date picker would have taken if not for the rule stated above.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
    <td></td>
    <td></td>
</table>

civihr_employee_portal/js/reports.js
reneolivo: This was added and removed by me. At first the field rows' height was done dynamically, but after fixing the fields list I removed the JS code and added a CSS fixed max-height.

civihr_employee_portal/js/reports.js
```diff
@@ -18,6 +19,43 @@
...
     this.processData(this.data);
+    this.originalFilterElement = jQuery('#report-filters').detach();
+  };
+
+  /**
+   * Compiles the filters element into an angular component. This is done for
+   * elements created outside of the Angular cycle (ie: created using jQuery, etc.)
+   */
+  HRReport.prototype.appendFilters = function () {
+    var element = this.originalFilterElement.clone();
```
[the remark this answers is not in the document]
reneolivo: The original element is kept in memory (removed from the DOM), but the clone is the one that actually gets attached to the DOM since the clone can be removed at any time when the report is updated.

mi-kon:
@jamienovick we didn't have mockups for this change, can you please confirm if screenshots are okay for you from design perspective?

civihr_employee_portal/css/custom.css
reneolivo: For this particular one, when I try it, it sets the width to 150px, not 100%. The table is 100%, but the li is not a direct descendant of the table.

civihr_employee_portal/js/reports.js
```diff
@@ -379,6 +381,7 @@
    * @param string viewReportDataTableId
    */
   HRReport.prototype.refreshReportTableViewInstance = function (viewReportDataTableId) {
+    var AjaxViews = Drupal.views.ajaxView;
```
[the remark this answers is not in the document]
reneolivo: yes, because you can't use a property as a constructor. This is where it is being used:

```js
// Before:
Drupal.views.instances['views_dom_id:' + viewReportDataTableNewId] = new AjaxViews(Drupal.settings.views.ajaxViews['views_dom_id:' + viewReportDataTableNewId]);

// After:
Drupal.views.instances['views_dom_id:' + viewReportDataTableNewId] = new Drupal.views.ajaxView(Drupal.settings.views.ajaxViews['views_dom_id:' + viewReportDataTableNewId]);
```

civihr_employee_portal/templates/civihr-employee-portal-civihr-report-custom.tpl.php
```diff
@@ -55,35 +55,77 @@ class="panel-body-wrap panel-body-wrap--collapse"
...
+          </div>
+          <div class="report-columns col-md-8">
+            <table>
+              <tr>
+              </tr>
+            </table>
+          </div>
+        </div>
+        <div class="row">
+        <div class="fields-selection col-md-2">
```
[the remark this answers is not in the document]
reneolivo: this is from an outdated commit.

reneolivo:
@bmnepali no Backstop tests were run (as per the comments section in the PR). The reason for this is because the report details section is changing completely, and also these changes don't affect other sections but this one.

civihr_employee_portal/css/custom.css
bmnepali: The properties are not alphabetically arranged!  Not sure why you move `overflow-x: hidden;` to last.
It should be like following no?
```css
#reportPivotTable .report-fields-selection table {
    display: block;
    max-height: 50vh;
    width: 100%;
    overflow-x: hidden;
    overflow-y: auto;
}
```

bmnepali: I think it should be like following:
```css
#reportPivotTable .report-fields-selection table {
  display: block; 
  max-height: 50vh;
  overflow-x: hidden;
  overflow-y: auto;
  width: 100%;
}
```
Not sure why you moved `overflow-x: hidden;` to the last.

jamienovick:
@reneolivo @mi-kon I think the changes Rene has done look fine visually for now. Just one point that the title "report builder" is not left aligned with the other text on the page.

Perhaps the save report buttons could be shown a bit closer to the configuration dropdown.

Also could we only show the relevant button depending on the state of the dropdown. i.e. when I have no configuration I should have save as new, but when I have a configuration I have all the options.

We can spend more time on this once we migrate fully to the main reporting extension.

PR #416, opened 2018-01-15
PCHR-3012: Onboarding Form Image Upload Limit

mickadoo:
## Overview

The onboarding form user image upload size limit should be 10 MB. It is currently 2 MB

## Before

The onboarding form user image upload size limit was 2 MB

image

## After

The onboarding form user image upload size limit is 10 MB

image

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.install
+++ b/civihr_employee_portal/civihr_employee_portal.install
@@ -277,6 +277,15 @@ function civihr_employee_portal_update_7016() {
 }
 
 /**
+ * Refreshes the node export files since the onboarding form file upload limit
+ * was increased.
+ */
+function civihr_employee_portal_update_7017() {
+  civicrm_initialize();
+  drush_civihr_employee_portal_refresh_node_export_files();
+}
+
+/**
  * Function to determine whether menu link exists or not.
  *
  * @param string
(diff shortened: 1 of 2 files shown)


civihr_employee_portal/civihr_employee_portal.install
```diff
@@ -276,6 +276,15 @@ function civihr_employee_portal_update_7016() {
   drupal_write_record('webform_civicrm_forms', $civicrmWebform, 'nid');
 }
 
+/**
+ * Refreshes the node export files since the onboarding form file upload limit
+ * was increased.
+ */
+function civihr_employee_portal_update_7017() {
+  civicrm_initialize();
```
tunbola: Do we need to initialize civi?

mickadoo: Yep, I tried it without it and only Drupal is bootstrapped in these upgraders. It fails without this with an error about not finding `civicrm_api3`

tunbola: I asked because the `drush_civihr_employee_portal_refresh_node_export_files` doen't seem like it needed it at first glsnce. Since its needed, will approve then

PR #417, opened 2018-01-15
PCHR-3178: Remove "Report Builder" header

reneolivo:
## Overview
This PR removes the Report Builder header from report pages. 

The reason to remove this header is because it's part of the [Report Improvement requirements](https://compucorp.atlassian.net/wiki/spaces/PCHR/pages/123699243/Reporting+User+Experience+Improvements).

## Before
leave reports hr17 9000 3

## After
leave reports hr17 9000 2

## Technical Details
This markup was removed from `civihr-employee-portal-civihr-report-custom.tpl.php`:

```html
<div class="chr_search-result__header">
  <div class="chr_search-result__total">
    Report Builder
  </div>
</div>
```

### DIFF
--- a/civihr_employee_portal/templates/civihr-employee-portal-civihr-report-custom.tpl.php
+++ b/civihr_employee_portal/templates/civihr-employee-portal-civihr-report-custom.tpl.php
@@ -21,11 +21,6 @@
   <div class="report-content panel-pane pane-block chr_panel chr_panel--no-padding">
     <?php if (!empty($jsonUrl)): ?>
       <div class="report-block report-builder tab-pane">
-        <div class="chr_search-result__header">
-          <div class="chr_search-result__total">
-            Report Builder
-          </div>
-        </div>
         <div id="report-filters" ng-controller="FiltersController as filters">
           <?php print render($filters); ?>
         </div>


civihr_employee_portal/templates/civihr-employee-portal-civihr-report-custom.tpl.php
```diff
@@ -21,11 +21,6 @@
   <div class="report-content panel-pane pane-block chr_panel chr_panel--no-padding">
     <?php if (!empty($jsonUrl)): ?>
       <div class="report-block report-builder tab-pane">
-        <div class="chr_search-result__header">
-          <div class="chr_search-result__total">
```
bmnepali: Since you removed the markup, what about the css classes here? Are these classes specific to the header or used in other parts as well?

reneolivo: I checked, there are none.

reneolivo: @bmnepali these classes are used elsewhere in SSP so the class definitions can't be removed. Examples:

https://github.com/compucorp/civihr-employee-portal/search?utf8=%E2%9C%93&q=chr_search-result__header&type=

PR #418, opened 2018-01-15
PCHR-3177: Fix Report dropdown arrows

reneolivo:
## Overview
This PR fixes an issue where the dropdown arrow of select elements inside of the report page would not open their select element when clicked.

## Before
anim
Clicking on the dropdown arrow doesn't work.

## After
anim

## Technical Details
This issue was introduced by [this commit](https://github.com/compucorp/civihr-employee-portal/commit/ab00dce2) since it overrides the z-index of the select element which originally was 2. The purpose of this commit was to fix date pickers to display above the select elements.

The datepickers were not being displayed on top of the selects because a parent element was using `z-index: 1` ([here](https://github.com/compucorp/civihr-employee-portal/pull/418/files#diff-c9717edfd0ee9646951283cf7eadc4e4L205)).

These z-index rules were not needed and were introduced in order to fix z-index issues before the report layout change, but they are not needed anymore and were removed.

### DIFF
--- a/civihr_employee_portal/css/custom.css
+++ b/civihr_employee_portal/css/custom.css
@@ -132,7 +132,7 @@ span.appraisals-employee-legend {
   margin-right: 8px;
 }
 
-#civihrReports .crm_custom-select > select {
+#civihrReports .crm_custom-select {
   z-index: 0;
 }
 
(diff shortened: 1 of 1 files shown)


deb1990:
@reneolivo 
1. in the description, you have mentioned 
```
position: relative;
z-index: 0;
```
but I dont see `position: relative` anywhere, could you please check that.

2. Is this problem with datepicker, limited to report page? if yes, why? and if no, then we should make a global fix.

PR #419, opened 2018-01-16
PCHR-3175: Rename filter labels

reneolivo:
## Overview
This PR renames the labels for the report filters based on these mockups:

Leave and Absence report filters:
reporting user experience improvements p chr civihr confluence 2

People's report filters:
reporting user experience improvements p chr civihr confluence 1

## Before
Leave and Absence's report:
leave reports hr17 9000 6

People's report:
people reports hr17 9000 5

## After
Leave and Absence's report:
leave reports hr17 9000 4

People's report:
people reports hr17 9000 6

## Technical Details
The `civihr_employee_portal_form_views_exposed_form_alter` method was refactored. This method now redirects the flow to the specific form alteration functions depending on the form ID:

```php
function civihr_employee_portal_form_views_exposed_form_alter(&$form, &$form_state, $form_id) {
  switch ($form['#id']) {
    // People's report filters:
    case 'views-exposed-form-civihr-report-people-default':
      _civihr_employee_portal_form_view_exposed_form_alter_people_report($form, $form_state, $form_id);
      break;
    // Leave and Absence's report filters:
    case 'views-exposed-form-civihr-report-leave-and-absence-default':
      _civihr_employee_portal_form_view_exposed_form_alter_leave_report($form, $form_state, $form_id);
      break;
    // Staff Directory and HR Document's form filters:
    case 'views-exposed-form-civihr-staff-directory-page':
    case 'views-exposed-form-hr-documents-hr-resources':
      _civihr_employee_portal_form_view_exposed_form_alter_actions_wrapper($form, $form_state, $form_id);
      _civihr_employee_portal_form_view_exposed_form_alter_labels_as_placeholders($form, $form_state, $form_id);
      break;
  }

  $form['submit']['#value'] = 'Filter';
}
```
The switch pattern was used instead of running all alteration functions as suggested by @davialexandre [here](https://compucorp.atlassian.net/browse/PCHR-3175?focusedCommentId=65317&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-65317) because:

* For now, only the form ID is needed to determine which alterations to run and can't see this changing in the foreseeable future.
* It's more clear which alterations will run depending on the form ID - No need to check each individual function to see which one will run or not depending on the condition defined at the beginning of the function. 

For the Leave and Absence filters the alteration function is:
```php
function _civihr_employee_portal_form_view_exposed_form_alter_people_report(&$form, $form_state, $form_id) {
  $form['between_date_filter']['value'] = setInputDatepickerForm();
  $form['between_date_filter']['#prefix'] = '<header class="form-group"><label>Headcount on date:</label></header>';
}
```

And for the People filters is:
```php
function _civihr_employee_portal_form_view_exposed_form_alter_leave_report(&$form, $form_state, $form_id) {
  $dateFields = &$form['absence_date_filter'];
  $dateFields['min'] = setInputDatepickerForm('From:');
  $dateFields['max'] = setInputDatepickerForm('To:');
  $dateFields['#prefix'] = '<header class="form-group"><label>Leave dates:</label></header>';
}
```

In terms of refactoring, the Staff Directory's and HR Document filters was split into two functions:

an actions wrapper for the submit buttons:
```php
function _civihr_employee_portal_form_view_exposed_form_alter_actions_wrapper(&$form, $form_state, $form_id) {
  $form['#attributes']['class'][] = 'row';

  $form['submit']['#attributes'] = ['class' => ['chr_action']];
  $form['submit']['#prefix'] = '<div class="col-sm-12"><div class="chr_panel__footer"><div class="chr_actions-wrapper">';
  $form['submit']['#suffix'] = '</div></div></div>';
}
```

And a function that moves each label and converts them to field's placeholders:
```php
function _civihr_employee_portal_form_view_exposed_form_alter_labels_as_placeholders(&$form, $form_state, $form_id) {
  foreach ($form as $key => $value) {
    if (isset($form['#info']["filter-$key"])) {
      $form[$key]['#prefix'] = '<div class="col-sm-6">';
      $form[$key]['#suffix'] = '</div>';

      // Take the <label> value and use it as placeholder instead
      if ($value['#type'] == 'textfield') {
        $form[$key]['#attributes'] = ['placeholder' => $form['#info']["filter-$key"]['label']];
      }

      $form['#info']["filter-$key"]['label'] = '';
    }
  }
}
```

## Comments
This PR is related to: https://github.com/compucorp/civihr-employee-portal-theme/pull/297
since that PR gives a bit of marging to the filter's header introduced by this PR.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -5566,7 +5566,7 @@ function civihr_employee_portal_form_views_exposed_form_alter(&$form, &$form_sta
  * @param array $form_state
  * @param string $form_id
  */
-function _civihr_employee_portal_form_view_exposed_form_alter_people_report(&$form, $form_state, $form_id) {
+function _civihr_employee_portal_form_view_exposed_form_alter_people_report(&$form, &$form_state, $form_id) {
   $form['between_date_filter']['value'] = setInputDatepickerForm();
   $form['between_date_filter']['#prefix'] = '<header class="form-group"><label>Headcount on date:</label></header>';
 }
@@ -5578,7 +5578,7 @@ function _civihr_employee_portal_form_view_exposed_form_alter_people_report(&$fo
  * @param array $form_state
  * @param string $form_id
  */
-function _civihr_employee_portal_form_view_exposed_form_alter_leave_report(&$form, $form_state, $form_id) {
+function _civihr_employee_portal_form_view_exposed_form_alter_leave_report(&$form, &$form_state, $form_id) {
   $dateFields = &$form['absence_date_filter'];
   $dateFields['min'] = setInputDatepickerForm('From:');
   $dateFields['max'] = setInputDatepickerForm('To:');
@@ -5592,7 +5592,7 @@ function _civihr_employee_portal_form_view_exposed_form_alter_leave_report(&$for
  * @param array $form_state
  * @param string $form_id
  */
-function _civihr_employee_portal_form_view_exposed_form_alter_actions_wrapper(&$form, $form_state, $form_id) {
+function _civihr_employee_portal_form_view_exposed_form_alter_actions_wrapper(&$form, &$form_state, $form_id) {
   $form['#attributes']['class'][] = 'row';
 
   $form['submit']['#attributes'] = ['class' => ['chr_action']];
@@ -5608,7 +5608,7 @@ function _civihr_employee_portal_form_view_exposed_form_alter_actions_wrapper(&$
  * @param array $form_state
  * @param string $form_id
  */
-function _civihr_employee_portal_form_view_exposed_form_alter_labels_as_placeholders(&$form, $form_state, $form_id) {
+function _civihr_employee_portal_form_view_exposed_form_alter_labels_as_placeholders(&$form, &$form_state, $form_id) {
   foreach ($form as $key => $value) {
     if (isset($form['#info']["filter-$key"])) {
       $form[$key]['#prefix'] = '<div class="col-sm-6">';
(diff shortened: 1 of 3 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5579,9 +5579,10 @@ function _civihr_employee_portal_form_view_exposed_form_alter_people_report(&$fo
  * @param string $form_id
  */
 function _civihr_employee_portal_form_view_exposed_form_alter_leave_report(&$form, $form_state, $form_id) {
-  $key = 'absence_date_filter';
-  $form[$key]['min'] = setInputDatepickerForm($form[$key]['min']['#title']);
```
davialexandre: So, before this change, the field label was set according to this `#title` property, right? Why did you have to manually change this by passing a different title to `setInputDatepickerForm()` instead of changing the value this `#title` property?

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5537,50 +5537,59 @@ function civihr_employee_portal_hrreport_custom_printtable($reportName, $dateFil
...
 
-  if (isset($form['#id']) && in_array($form['#id'], $exposed_filters_forms)) {
-    $form['#attributes']['class'][] = 'row';
+  $form['submit']['#value'] = 'Filter';
+}
 
-    $form['submit']['#attributes'] = ['class' => ['chr_action']];
-    $form['submit']['#prefix'] = '<div class="col-sm-12"><div class="chr_panel__footer"><div class="chr_actions-wrapper">';
-    $form['submit']['#suffix'] = '</div></div></div>';
+function _civihr_employee_portal_form_view_exposed_form_alter_people_report(&$form, $form_state, $form_id) {
```
davialexandre: Since `$form_state` is passed by reference to the original function, so it's better to do the same for these functions you've extracted in case we need to change the form state in any of them

civihr_employee_portal/civihr_employee_portal.module
davialexandre: We talked about this and, since this report solution will soon be removed, the changes in this PR are ok

PR #422, opened 2018-01-23
PCHR-3207: SSP Report fixes

reneolivo:
## Overview
This PR fixes the following report page issues:

* SVG charts breaking the new report layout. 
* Make the report look better in small screens, though it doesn't rearrange the elements as there are no wireframes for it.
* Fixes an issue where the data table scrolls would not appear  until scrolling with the mouse wheel or by chance.

### Desktop
leave reports hr17 9000 13

### Small Screens
leave reports hr17 9000 16

### Desktop
leave reports hr17 9000 11

### Small Screens
leave reports hr17 9000 12
**Note:** Fields selection can't have a `overflow: scroll` without considerable JS changes since the draggable elements don't work well with overflowed elements.

### Before
anim
**Notes:** The scrollbar only appears after scrolling with the mouse wheel. This is not a regression issue, it has been like that since the implementation of the report.

### After
anim

### Moving CSS rules into SSP Theme

Some of the CSS rules for the pivot table where moved from the static [civihr_employee_portal/css/custom.css](https://github.com/compucorp/civihr-employee-portal/pull/422/files#diff-c9717edfd0ee9646951283cf7eadc4e4L229) file into [the SSP Theme](https://github.com/compucorp/civihr-employee-portal-theme/pull/302/files) because the former file should be removed at some point and all CSS rules should reside in the SSP Theme. This is a step towards that goal.

### Column breakpoints

The columns breakpoints changed from SM to MD since the line charts were breaking using the medium width.

### Adapting the SVG chart to parent container

The following method was added so the SVG charts can fit inside their parent container and adapt to its width:

```js
HRReport.prototype.adaptSvgProportionsToContainer = function () {
  this.pivotTableContainer.find('svg')
    .removeAttr('width') // No width or height = auto
    .removeAttr('height')
    .each(function () {
      // this is done so jQuery can properly set the viewBox attribute of the SVG element.
      // 0 0 are the origin coordinates
      // 800 800 were choosen because they give the best ratio to display the charts
      $(this)[0].setAttribute('viewBox', '0 0 800 800');
    });
};
```

### Fixing scrollbar issue

The problem with the scrollbar not showing up is that the `Ps.initialize(el)` only works when the DOM element is visible. To fix this, `Ps.initialize` is only executed when clicking on the "view-data" tab and a timeout is added to wait for the tabs component to properly display the data:

```js
HRReport.prototype.initScrollbarFallbackOnTabChange = function () {
  var el = document.querySelector('.chr_custom-scrollbar');

  $('[data-tab="view-data"]').click(function () {
    setTimeout(function () {
      Ps.initialize(el);
    }, 0);
  });
};
```

## Comments

All the CSS changes for this PR are included in this SSP Theme PR:
https://github.com/compucorp/civihr-employee-portal-theme/pull/302

### DIFF
--- a/civihr_employee_portal/js/reports.js
+++ b/civihr_employee_portal/js/reports.js
@@ -7,9 +7,7 @@
   /**
    * Define HRReport object.
    */
-  function HRReport () {
-    this.initScrollbarFallback();
-  }
+  function HRReport () {}
 
   /**
    * Initialization function.
@@ -19,6 +17,7 @@
   HRReport.prototype.init = function (options) {
     $.extend(this, options);
     this.initAngular();
+    this.initScrollbarFallbackOnTabChange();
     this.processData(this.data);
     this.originalFilterElement = $('#report-filters').detach();
   };
@@ -408,12 +407,14 @@
    * Init the scrollbar fallback
    *
    */
-  HRReport.prototype.initScrollbarFallback = function () {
+  HRReport.prototype.initScrollbarFallbackOnTabChange = function () {
     var el = document.querySelector('.chr_custom-scrollbar');
 
-    if (el) {
-      Ps.initialize(el);
-    }
+    $('[data-tab="view-data"]').click(function () {
+      setTimeout(function () {
+        Ps.initialize(el);
+      }, 0);
+    });
   };
 
   /**
@@ -473,7 +474,6 @@
       success: function (data) {
         that.tableContainer.html(data);
         that.refreshReportTableViewInstance(tableDomId);
-        that.initScrollbarFallback();
       },
       type: 'GET'
     });
--- a/civihr_employee_portal/js/reports.js
+++ b/civihr_employee_portal/js/reports.js
@@ -939,15 +939,13 @@
      * Bind tab switching events for the report tabs.
      */
     bindTabsEvents: function () {
-      var self = this;
-
-      $('.report-tabs a').bind('click', function (e) {
+      $('.report-tabs a').on('click', function (e) {
         $('.report-tabs li').removeClass('active');
-        $(this).parent().addClass('active');
+        $(e.target).parent().addClass('active');
         $('.report-block').addClass('hidden');
-        $('.report-block.' + $(this).data('tab')).removeClass('hidden');
-        self.initScrollbarFallbackOnTabChange();
-      });
+        $('.report-block.' + $(e.target).data('tab')).removeClass('hidden');
+        this.initScrollbarFallbackOnTabChange();
+      }.bind(this));
     },
     /**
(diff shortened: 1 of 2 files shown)


deb1990:
pchr-3207__ssp_report_fixes_by_reneolivo_ _pull_request__422_ _compucorp_civihr-employee-portal

Padding top should be consistent with other sides.

civihr_employee_portal/js/reports.js
```diff
@@ -323,6 +323,20 @@
...
+   * Updates the SVG element proportions so it adapts them to its
+   * parent container. This helps the SVG charts to display their information
+   * without breaking.
+   */
+  HRReport.prototype.adaptSvgProportionsToContainer = function () {
+    this.pivotTableContainer.find('svg')
+      .removeAttr('width')
+      .removeAttr('height')
+      .each(function () {
+        $(this)[0].setAttribute('viewBox', '0 0 800 800');
```
deb1990: Why such a arbitrary value of `'0 0 800 800'`?

civihr_employee_portal/js/reports.js
```diff
@@ -393,12 +407,14 @@
...
    */
-  HRReport.prototype.initScrollbarFallback = function () {
+  HRReport.prototype.initScrollbarFallbackOnTabChange = function () {
     var el = document.querySelector('.chr_custom-scrollbar');
 
-    if (el) {
-      Ps.initialize(el);
-    }
+    $('[data-tab="view-data"]').click(function () {
+      setTimeout(function () {
```
deb1990: Why do we need a timeout?

deb1990: `timeout is added to wait for the tabs component to properly display the data:`
can we use `shown.bs.tab` event to wait for the tab to be shown properly, and eliminate timeout?

reneolivo: These tabs are not using Bootstrap's tabs, but [custom code](https://github.com/compucorp/civihr-employee-portal/blob/staging/civihr_employee_portal/js/reports.js#L680). Simply changing the tabs to Bootstrap doesn't work since the URL switch functionality is lost and the tab switch events are not triggered.

For now I have moved the scroll initialization to the Behaviour, in this way we avoid the setTimeout.

civihr_employee_portal/js/reports.js
```diff
@@ -935,11 +939,14 @@
      * Bind tab switching events for the report tabs.
      */
     bindTabsEvents: function () {
+      var self = this;
```
deb1990: Please never use `self`, instead use `.call`/`.bind`/`.apply`.

PR #425, opened 2018-01-23
PCHR-3051: Make Some Onboarding Form Fields Non Required

mickadoo:
## Overview

After some client feedback we need to make fewer of the onboarding form fields required. We will add a radio button to hide the fields on a section and make them non-required. This should speed progress through the form.

## Before

The following fields were required:

- Nationality
- Preferred name
- Account holder
- Sort code
- Account number
- Bank name
- NI/SSN
- Marital status
- Relationship with employee (for all emergency contacts)
- Mobile number (for all dependants)
- Relationship with employee (for all dependants)

And there was no option to skip any step

image

## After

The following sections will have radio button to allow you to collapse all fields under it when "Add Later" is selected:

- Payroll
- Emergency Contacts
- Dependants

The following fields are no longer required:

- Nationality
- Preferred name
- Account holder
- Sort code
- Account number
- Bank name
- NI/SSN
- Marital status
- Relationship with employee (for all emergency contacts)
- Mobile number (for all dependants)
- Relationship with employee (for all dependants)

There is an  radio button option to "Add Later" which hides the input fields.

peek 2018-01-25 11-17

## Notes

- There is a warning in the GIF shown, this is already reported and a separate ticket was created to fix it

---

- [ ] Tests Pass


tunbola:
@mickadoo, Code looks good but it would be good to make some changes to the PR description/title
A better title would be Make Some Onboarding Form Fields Non Required
It will also be nice to list in the PR details the Form sections that allows you to skip the fields, I was under the impression that all the sections allowed it until i took a look at the code and paid more attention to the attached Gif.

tunbola:
@mickadoo, Code looks good but since the form field for skipping the steps has changed, It will be good to update the Gif to reflect the change.

PR #426, opened 2018-01-24
PCHR-3209: Add Link to Skip Onboarding

mickadoo:
## Overview

Based on client feedback some people felt that there should be an option to skip the onboarding form. We [previously](https://github.com/compucorp/civihr-employee-portal/pull/387) added a description for users who had been created before the onboarding release. This description will be extended to all users, but the content will depend on whether they are users created before the onboarding release.

## Before

- There was no button to skip the onboarding form
- Only users created before the onboarding form release were shown a message explaining the form

#### Users created before onboarding:

image

#### Users created after onboarding

image

## After

- There is a button to skip the onboarding form
- All users will have a description of the onboarding form
- The description will differ depending on whether they were created before the onboarding form release
- The link will bring users with "Access CiviCRM" permission to the tasks dashboard, and those without this permission to the SSP dashboard.

#### Users created before onboarding:

image

#### Users created after onboarding

image

## Notes
- The styles for the `Remind Me` Button has been fixed in https://github.com/compucorp/civihr-employee-portal-theme/pull/309. There the updated screenshots can be found.
- The same code to find the landing page would be used in two places, so I moved it to a new class, the `LinkProvider`

---

- [ ] Tests Pass
Some tests are not running because they depend on custom groups being present. I'll fix these in a separate ticket since they're not related to this.

### DIFF
--- a/civihr_employee_portal/src/Forms/OnboardingWebForm.php
+++ b/civihr_employee_portal/src/Forms/OnboardingWebForm.php
@@ -5,6 +5,7 @@ namespace Drupal\civihr_employee_portal\Forms;
 use Drupal\civihr_employee_portal\Helpers\WebformHelper;
 use Drupal\civihr_employee_portal\Service\ContactService;
 use Drupal\civihr_employee_portal\Service\TaskCreationService;
+use Drupal\civihr_employee_portal\Helpers\LinkProvider;
 
 class OnboardingWebForm {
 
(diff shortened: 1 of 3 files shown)


mi-kon:
@jon what do you think normal and hover/clicked colours for button and button text should be?

JonScreat:
@mi-kon I would suggest a white button, keeping the text the same as it is at the moment. On hover perhaps  #757b8d with white text.

civihr_employee_portal/src/Forms/OnboardingWebForm.php
```diff
@@ -234,11 +234,10 @@ private function getHelpText() {
    * @return string
    */
   private function getSkipButtonMarkup() {
-    $classes = 'btn btn-default-outline';
-    $style = 'float:none;';
+    $classes = 'btn btn-default chr_onboarding-wizard_remind-me-later';
     $format = '<button type="button" style="%s" class="%s">%s</button>';
     $buttonText = ts('Skip and Remind Me Later');
-    $buttonMarkup = sprintf($format, $style, $classes, $buttonText);
+    $buttonMarkup = sprintf($format, null, $classes, $buttonText);
```
mickadoo: Instead of changing `$style` to NULL you can just remove it from the `$format` so it would become

```
     $format = '<button type="button" class="%s">%s</button>';
```

And then remove the argument completely from `sprintf`

```
$buttonMarkup = sprintf($format, $classes, $buttonText);
```

PR #435, opened 2018-02-02
PCHR-3071: Rename Some Roles Without Changing the Machine Names

tunbola:
## Overview
This PR renames some roles without changing their machine names:
- civihr_admin was renamed to HR Admin
- civihr_staff was renamed to  Staff
-  civihr_manager was renamed to Manager
- civihr_local_admin was renamed to Regional HR Admin

## Before
The roles were not renamed.

## After
The role names were changed.

## Technical Details
- For new sites, the role names were changed via the UI and the `civihr_default_permissions` feature was re-exported.
- For existing site, an upgrader was added to rename the roles, Also there was a need to rename some permissions created by the role_delegation module .
These permissions are always renamed when the role names are changed via the UI (as can be seen in the exported feature) via the `form_user_admin_role_submit` Hook, See snippet below:

```php
function role_delegation_form_user_admin_role_submit($form, $form_state) {
  $op = $form_state['values']['op'];
  if ($op != t('Save role')) {
    return;
  }
  $oldrole = $form_state['build_info']['args'][0]->name;
  $newrole = $form_state['values']['name'];
  if ($oldrole == $newrole) {
    return;
  }
  db_update('role_permission')
  ->condition('permission', "assign $oldrole role")
  ->fields(array('permission' => "assign $newrole role"))
  ->execute();
}
```

Since renaming the roles via the upgrader would not rename these permissions, it is done in the upgrader after the role is renamed.

- [X] Tests Pass


civihr_employee_portal/features/civihr_default_permissions/civihr_default_permissions.features.user_permission.inc
```diff
@@ -36,9 +36,9 @@ function civihr_default_permissions_user_default_permissions() {
   $permissions['access CiviCRM'] = array(
     'name' => 'access CiviCRM',
     'roles' => array(
+      'HR Admin' => 'HR Admin',
+      'Regional HR Admin' => 'Regional HR Admin',
```
davialexandre: I thought this array was something like "machine_name" => "name". Why would someone need an array where the keys and the values are the same??????

civihr_employee_portal/features/civihr_default_permissions/civihr_default_permissions.features.user_permission.inc
```diff
@@ -23,11 +23,11 @@ function civihr_default_permissions_user_default_permissions() {
   $permissions['access AJAX API'] = array(
     'name' => 'access AJAX API',
     'roles' => array(
```
davialexandre: Just to confirm, you updated this by downloading the feature from the Drupal admin and then extracting the files, right?

civihr_employee_portal/civihr_employee_portal.install
```diff
@@ -317,6 +317,37 @@ function civihr_employee_portal_update_7020() {
...
+    if (array_key_exists($roleName, $oldToNewNames)) {
+      $roleObject = user_role_load($id);
+      $roleObject->name = $oldToNewNames[$roleName];
+      user_role_save($roleObject);
+
+      //The role delegation module has dynamic permissions that changes
+      //when a role name is updated in UI via hook form_user_admin_role_submit
+      //We need to update it manually here.
+
+      db_update('role_permission')
```
davialexandre: Wouldn't reverting the civihr_default_permissions feature have the same effect?

tunbola: It did have the same effect but I discovered that the previous permissions were left in the database causing redundant data.

civihr_employee_portal/features/civihr_default_permissions/civihr_default_permissions.features.user_permission.inc
tunbola: Thought same too, doesn't seem to make much sense, Was when I re-exported the feature that I saw it that way too

PR #437, opened 2018-02-08
PCHR-3269: Fix Image Query Params

mickadoo:
## Overview

During the course of onboarding testing we discovered that when using the image uploaded from the webform for a contact image CiviCRM would show a warning about [an undefined "photo" index.](https://github.com/civicrm/civicrm-core/blob/17ae6ca46f1d544d40e5bfb77ab6727975ba7e17/CRM/Utils/File.php#L913).

To fix this I made [this commit](https://github.com/compucorp/civihr-employee-portal/pull/406/commits/92cc1dffd0f20040f1a58c2cf8ded77896eafbe1) to add `photo=0` to images when the onboarding form was submitted. This fixed the warning and both regular images and images saved from the webform were displayed without warning.

Unfortunately I did not consider this case:

1. A contact was created and somebody uploaded an image for them
1. This contact logs in and goes through the onboarding form
1. They do not upload their own image, so the existing image URL is used
1. This image URL is not stored by the webform, it's a regular `civicrm/contact/imagefile?photo=<imagename>` because an extra `photo=0` is still appended to the end it breaks the URL

## Before

* If a user with an existing contact image does the onboarding form an errant `photo=0` will be appended to their image URL, breaking the URL
* If a user redoes the onboarding form an extra `photo=0` will be appended to their image URL

Examples

|Image URL Before Onboarding|Image URL After Onboarding|
|--|--|
|`/civicrm/contact/imagefile?photo=cat.jpeg`|`/civicrm/contact/imagefile?photo=cat.jpeg&photo=0`|
|`/sites/default/files/webform/500_3.jpg?photo=0`|`/sites/default/files/webform/500_3.jpg?photo=0&photo=0`|

## After

* Users can complete the onboarding form as many times as they want, and their image URL will not be broken.
* Any existing image URLs with duplicate query parameters will be fixed

## Notes

It's possible that in production some administrators created contacts, uploaded their images, and then created accounts for these people. If these people logged in afterwards and complete onboarding _without_ changing their image, their image URLs would be broken, so I added an upgrader to fix these cases.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.install
+++ b/civihr_employee_portal/civihr_employee_portal.install
@@ -330,7 +330,7 @@ function civihr_employee_portal_update_7021() {
 /**
  * Fixes contact images that have multiple 'photo' query params.
  */
-function civihr_employee_portal_update_7021() {
+function civihr_employee_portal_update_7022() {
   civicrm_initialize();
   $params = ['return' => ['image_URL']];
   $contacts = civicrm_api3('Contact', 'get', $params)['values'];
--- a/civihr_employee_portal/civihr_employee_portal.install
+++ b/civihr_employee_portal/civihr_employee_portal.install
@@ -2,6 +2,7 @@
 
 use Drupal\civihr_employee_portal\Helpers\WebformHelper;
 use Drupal\civihr_employee_portal\Forms\OnboardingWebForm;
+use Drupal\civihr_employee_portal\Helpers\UrlHelper;
 
 /**
  * Create Report Pages and Report Age Settings menu links.
@@ -327,9 +328,28 @@ function civihr_employee_portal_update_7021() {
 }
 
 /**
+ * Fixes contact images that have multiple 'photo' query params.
+ */
+function civihr_employee_portal_update_7021() {
+  civicrm_initialize();
+  $params = ['return' => ['image_URL']];
+  $contacts = civicrm_api3('Contact', 'get', $params)['values'];
+
+  foreach ($contacts as $contact) {
+    $imageUrl = $contact['image_URL'];
+    $dedupedImageUrl = UrlHelper::dedupeUrlQueryParams($imageUrl);
+
+    if ($imageUrl !== $dedupedImageUrl) {
+      $params = ['image_URL' => $dedupedImageUrl, 'id' => $contact['id']];
+      civicrm_api3('Contact', 'create', $params);
+    }
+  }
+}
+
+/**
  * Function to determine whether menu link exists or not.
  *
- * @param string
+ * @param string $path
  *   Path of the menu link.
  *
  * @return bool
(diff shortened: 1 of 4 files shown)


civihr_employee_portal/tests/Helpers/UrlHelperTest.php
```diff
@@ -0,0 +1,49 @@
...
+      [
+        'http://civihr.local/civicrm/contact/imagefile?photo=pp_2e6adadea8d41b0ac065e888f2e61dc5.jpeg&photo=0',
+        'http://civihr.local/civicrm/contact/imagefile?photo=pp_2e6adadea8d41b0ac065e888f2e61dc5.jpeg',
+      ],
+      [
+        'http://www.civihr.net?foo=bar&bar=car&foo=bar2&bar=car2#someanchor',
+        'http://www.civihr.net?foo=bar&bar=car#someanchor',
+      ],
+    ];
+  }
```
davialexandre: What about adding tests for string in this format: `foo[]=1&foo[]=2&bar[a]=3` ?

civihr_employee_portal/civihr_employee_portal.install
```diff
@@ -317,10 +318,29 @@ function civihr_employee_portal_update_7020() {
...
+ * Fixes contact images that have multiple 'photo' query params.
+ */
+function civihr_employee_portal_update_7021() {
+  civicrm_initialize();
+  $params = ['return' => ['image_URL']];
+  $contacts = civicrm_api3('Contact', 'get', $params)['values'];
+
+  foreach ($contacts as $contact) {
+    $imageUrl = $contact['image_URL'];
+    $dedupedImageUrl = UrlHelper::dedupeUrlQueryParams($imageUrl);
```
davialexandre: This will not only photo duplicates but any duplicate param. Is this really want we want? Wouldn't it be a good idea if we could specify which param we want to be fixed?

civihr_employee_portal/tests/Helpers/UrlHelperTest.php
mickadoo: Good idea, it also made me realize I hadn't prepared for this so I updated the `UrlHelper` too

civihr_employee_portal/civihr_employee_portal.install
mickadoo: There isn't really a standard around this area. I thought using arrays like `foo[]=bar&foo[]=foo` was the norm, but it seems to depend on which language you're using. For PHP anyway, if you use `foo=bar&foo=foo` it will use the last value for `foo`. This is what led to the bug this ticket is fixing.

Since the environment is PHP I think removing all duplicate params would be practical, but if you think it's best to only target photo I can update the upgrader and add a new method to the `UrlHelper` like `dedupeSpecificUrlQueryParam`

davialexandre: Since now the helper can handle arrays, I think it's safe to keep this as is now.

PR #447, opened 2018-02-22
PCHR-3356: Adds an upgrader to revert default civihr user permissions

mickadoo:
## Overview

Because of the changes in #446 we need a new upgrader to revert the default permissions.

## Before

There was no new updgrader to refresh permissions.

## After

A new upgrader was added which refreshes only the `user_permission` for `civihr_default_permissions`.

---

- [x] Tests Pass


mickadoo:
Closing and merging changes in this to #446

PR #450, opened 2018-02-26
PCHR-3354: Rename permission

mickadoo:
## Overview

After #449 was merged @davialexandre asked me to update the permission names and add the new permission in CiviHR too (since it will be used in CiviHR Admin). This PR updates the permission name and label to match those in https://github.com/civicrm/civihr/pull/2498

## Before

- The 'access root menu items and configurations' had a 'civihr -' prefix. 
- The label used `-` instead of `:`

## After

- The 'civihr -' prefix was removed from 'civihr - access root menu items and configurations'. 
- The label uses `:` instead of `-`

---

- [x] Tests Pass


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -706,8 +706,8 @@ function civihr_employee_portal_permission() {
       'title' => t('Manage HR Reports configuration'),
       'description' => t('Availability to manage HR Reports configuration'),
     ],
-    'civihr - access root menu items and configurations' => array(
-      'title' => t('CiviHR - Access root menu items and configurations'),
+    'access root menu items and configurations' => array(
+      'title' => t('CiviHR: Access root menu items and configurations'),
       'description' => t('Access advanced configuration options that could potentially impair site functionality')
```
davialexandre: since you're adding the permission in the civicrm extension, you don't need to add it here as well

PR #451, opened 2018-02-26
PCHR-3362: Rename Default option to bulk edit.

tunbola:
## Overview
This PR renames the default option for the Bulk operations dropdown from `Choose an operation` to `Bulk edit`

## Before
 The default option for the Bulk operations dropdown is `Choose an operation`

users list _ staging17 2018-02-26 14-39-57

## After
 The default option for the Bulk operations dropdown is `Bulk edit`

users list _ staging17 2018-02-26 14-42-08

Changes does not affect tests.
- [] Tests Pass

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -4715,6 +4715,22 @@ function civihr_employee_portal_form_user_profile_form_alter(&$form, &$form_stat
   $form['#submit'][] = 'civihr_employee_portal_form_user_profile_form_submit';
 }
 
+
+/**
+ * Implementation of hook_views_bulk_operations_form_alter.
+ *
+ * Performs alterations on the VBO form before it is rendered.
+ *
+ * @param array $form
+ * @param array $form_state
+ * @param object $vbo
+ */
+function civihr_employee_portal_views_bulk_operations_form_alter(&$form, &$form_state, $vbo) {
+  if ($form['#id'] == 'views-form-users-list-page') {
+    $form['select']['operation']['#options'][0] = '- Bulk edit -';
+  }
+}
+
 /**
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -4715,6 +4715,22 @@ function civihr_employee_portal_form_user_profile_form_alter(&$form, &$form_stat
...
+ *
+ * @param array $form
+ * @param array $form_state
+ * @param object $vbo
+ */
+function civihr_employee_portal_views_bulk_operations_form_alter(&$form, &$form_state, $vbo) {
+  if ($form['#id'] == 'views-form-users-list-page') {
+    $form['select']['operation']['#options'][0] = '- Bulk edit -';
+  }
+}
```
davialexandre: Do we really need to use a hook to change the text? Can't we change it in the view definition?

tunbola: I tried changing it there, It didn't work: https://github.com/compucorp/civihr-employee-portal/blob/3bd2300e8315f11bc2c1ab0b89e785d664cbabe3/civihr_employee_portal/features/civihr_employee_portal_features/civihr_employee_portal_features.views_default.inc#L2132
I had to dig around before I found the hook specific for VBO for altering the fields.

davialexandre: did you revert the view after changing that?

tunbola: I did and the message was, the default state already matches current and it aborted.
I think the issue is that the option label is hardcoded in the `function views_bulk_operations_form($form, &$form_state, $vbo) ` in the VBO module and the hook is needed to override this.

PR #454, opened 2018-03-02
PCHR-3273: Fix Staff Directory Responsiveness

deb1990:
## Overview
In SSP Staff Directory, the views table was broken in the mobile view. This PR fixes that.

## Before
civihr local_8914_staff-directory iphone 7 1

## After
civihr local_8914_staff-directory iphone 7

## Technical Details
1. In `civihr_employee_portal/civihr_employee_portal.module`
Added `civihr_staff_directory` to the `$viewsList` array in `_is_scrollable_table_view_with_results` function, to use the `views-view-table-scrollable-with-results.tpl.php` which is meant to solve the scroll issue.

2. In `civihr_employee_portal.views.inc`
Removed `($view->name == 'civihr_staff_directory' && $display_id == 'page')` condition as `views-view-table-scrollable-with-results.tpl.php` already holds the header count.

3. From the staff directory view, removed the header from drupal views menu as `views-view-table-scrollable-with-results.tpl.php` already holds the header count.

4. In `views-view-table-scrollable-with-results.tpl.php` previously we were only showing the number of records visible i.e. if we had 12 records, and only 10 can be shown at once, and rest after pagination. Previously 10 was shown on header. Now it is fixed to show the total number.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -6095,7 +6095,10 @@ function _is_hrreports_data_table_view($viewName) {
  * @return boolean
  */
 function _is_scrollable_table_view_with_results($viewName) {
-  $viewsList = ['users_list'];
+  $viewsList = [
+    'civihr_staff_directory',
+    'users_list'
+  ];
 
   return in_array($viewName, $viewsList);
 }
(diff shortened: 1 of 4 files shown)


tunbola:
@deb1990 , In PR description s/Not it is fixed to show the total number./Now it is fixed to show the total number.

PR #458, opened 2018-03-26
PCHR-2982: Onboarding Emergency Contact Create Only

mickadoo:
## Overview

There was a problem in how "Emergency Contacts" and "Dependants" were displayed in the onboarding form. To avoid problems with the display order we will change these to be "Create Only" entries, creating a new emergency contact each time the form is submitted.

## Before

If a user had emergency contacts already created for them they would be displayed in the order the appear in the database. The actual order in the onboarding form is:

- 1-2: Emergency Contacts
- 3 -7: Dependants

So sometimes dependants would appear in the emergency contact tab, and vice versa.

image

Some work was done to extend the improvements to webform portability started in #397. The fieldset keys were set to the format expected by `webform_civicrm`. See technical details for more.

## After

Both "Emergency Contact" and "Dependant" use "Create Only" mode. This means no pre-existing values will be populated in the form and a new entry will be created every time.

peek 2018-03-26 18-44

You can save "Create Mode" for onboarding form custom groups.

Create mode is successfully imported even if custom group ID changes.

## Notes

There is a warning in the GIF about undefined indices. This is coming from `webform_civicrm` after Kacper's PR to add "Create Only" mode and I have submitted [a patch](https://github.com/colemanw/webform_civicrm/pull/125) for it.

#### Importing Fieldset Keys

Say we have a link to "Emergency Contacts" in a webform, and this custom group has an ID of 99. If the exported fieldset key is `civicrm_1_contact_1_cg99_fieldset"` but the production system we're importing the webform has a different custom group ID some of the keys will be broken. In #379 I added some handling for different fields. In this PR the handling is extended to fix custom group ID in fieldset keys and create mode.

---

- [x] Tests Pass


civihr_employee_portal/tests/Webform/CustomComponentKeyHelperTest.php
```diff
@@ -39,10 +39,48 @@ public function testGettingFieldID($key, $expectedGroup, $expectedField) {
...
+    $expected = !is_null($expected); // if expected not null it is custom field
+    $this->assertEquals($expected, KeyHelper::isCustomFieldsetKey($key));
+  }
+
+  /**
+   * @dataProvider rebuildCustomFieldsetKeyProvider
+   *
+   * @param $original
+   * @param $newGroup
+   * @param $expected
```
tunbola: Please add param type to the parameters for this function.

civihr_employee_portal/tests/Webform/CustomComponentKeyHelperTest.php
```diff
@@ -39,10 +39,48 @@ public function testGettingFieldID($key, $expectedGroup, $expectedField) {
...
+   */
+  public function testGettingGroupIdWillReturnExpectedGroup($key, $expected) {
+    $this->assertEquals($expected, KeyHelper::getCustomFieldsetGroupId($key));
+  }
+
+  /**
+   * @dataProvider fieldsetKeyProvider
+   *
+   * @param string $key
+   * @param bool $expected
```
tunbola: Param type should still be string as above. The conversion to boolean happens inside the function.

civihr_employee_portal/src/Webform/WebformExportCustomFieldConvertor.php
```diff
@@ -194,10 +182,86 @@ private static function reverseNameMapping($entity, $originalMapping) {
...
+   *   The original form key
+   * @param array $groupMapping
+   *   Mapping of original custom group IDs to current ones on system
+   * @param $fieldMapping
+   *   Mapping of original custom field IDs to current ones on system
+   *
+   * @return string|null
+   *   The updated field key, or null if the custom group or field is unregonize
+   */
+  private static function getUpdatedCustomFieldKey(
```
tunbola: Missing function description

civihr_employee_portal/src/Webform/WebformExportCustomFieldConvertor.php
```diff
@@ -194,10 +182,86 @@ private static function reverseNameMapping($entity, $originalMapping) {
...
+      return KeyHelper::rebuildCustomFieldKey($newGroupID, $newFieldID, $formKey);
+    }
+
+    return NULL;
+  }
+
+  /**
+   * @param $formKey
+   * @param $groupMapping
+   * @return string
```
tunbola: Missing param types.
return string|null

PR #459, opened 2018-04-06
PCHR-3508: Alter Meta viewport tag

igorpavlov:
## Overview

This PR fixes viewport auto-zooming on inputs focusing on iOS devices. It still allows users to pinch and zoom with 2 fingers.

## Technical Details

Please see https://api.drupal.org/api/drupal/modules%21system%21system.api.php/function/hook_html_head_alter/7.x

```php
// hook_html_head_alter
function civihr_employee_portal_html_head_alter(&$head_elements) {
  // ...
  _set_maximum_scale_to_viewport_meta_tag($head_elements);

function _set_maximum_scale_to_viewport_meta_tag(&$head_elements) {
  // bruteforcing is the only secure way to get the Meta Viewport tag in this case
  foreach ($head_elements as $key => $tag) {
    if (isset($tag['#attributes']['name']) && $tag['#attributes']['name'] === 'viewport') {
      $head_elements[$key]['#attributes']['content'] .= ', maximum-scale=1';
```

---

✅Manual Tests - passed

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -6192,6 +6192,34 @@ function _get_task_filter_by_date($date, $normalize = false) {
 }
 
 /**
+ * Sets the maximum viewport scale to 1 to prevent auto zooming on input focus
+ * @see https://stackoverflow.com/questions/11064237/prevent-iphone-from-zooming-form
+ *
+ * @param object $head_elements
+ */
+function _set_maximum_scale_to_viewport_meta_tag(&$head_elements) {
+  foreach ($head_elements as $key => $tag) {
+    if (isset($tag['#attributes']['name']) && $tag['#attributes']['name'] === 'viewport') {
+      $head_elements[$key]['#attributes']['content'] .= ', maximum-scale=1';
+    }
+  }
+}
+
+/**
+ * Sets theme favicon
+ *
+ * @param object $head_elements
+ */
+function _set_favicon(&$head_elements) {
+  global $base_url;
+
+  $default_favicon_element = 'drupal_add_html_head_link:shortcut icon:' . $base_url . '/misc/favicon.ico';
+  $icoPath = drupal_get_path('module', 'civihr_employee_portal') . "/images/favicon.ico";
+
+  $head_elements[$default_favicon_element]['#attributes']['href'] = "{$base_url}/{$icoPath}";
+}
+
+/**
  * This method handle user redirection for special use cases
  *
  * The redirection is made by manually setting the Location header,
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6191,6 +6191,34 @@ function _get_task_filter_by_date($date, $normalize = false) {
...
+/**
+ * Sets the maximum viewport scale to 1 to prevent auto zooming on input focus
+ * @see https://stackoverflow.com/questions/11064237/prevent-iphone-from-zooming-form
+ *
+ * @param object $head_elements
+ */
+function _set_maximum_scale_to_viewport_meta_tag(&$head_elements) {
+  foreach ($head_elements as $key => $tag) {
+    if (isset($tag['#attributes']['name']) && $tag['#attributes']['name'] === 'viewport') {
+      $head_elements[$key]['#attributes']['content'] .= ', maximum-scale=1';
```
deb1990: Its better to check if maximum-scale is already present. I know its not now, but incase we add it from somewhere else or we update the theme later, there is a chance of having this twice.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6191,6 +6191,34 @@ function _get_task_filter_by_date($date, $normalize = false) {
...
 
+/**
+ * Sets the maximum viewport scale to 1 to prevent auto zooming on input focus
+ * @see https://stackoverflow.com/questions/11064237/prevent-iphone-from-zooming-form
+ *
+ * @param object $head_elements
+ */
+function _set_maximum_scale_to_viewport_meta_tag(&$head_elements) {
+  foreach ($head_elements as $key => $tag) {
+    if (isset($tag['#attributes']['name']) && $tag['#attributes']['name'] === 'viewport') {
```
deb1990: What if we just do `$tag['#attributes']['name'] === 'viewport'`

deb1990:
@igorpavlov I see that we can still pinch to zoom. is that a requirement? because in mobile version we generally disable the same.

civihr_employee_portal/civihr_employee_portal.module
igorpavlov: @deb1990 this will throw a warning if the key does not exist. This was also recommended by a community https://stackoverflow.com/questions/49656547/how-to-override-already-added-meta-tags-in-drupal-7/49656741#49656741 and official pages https://api.drupal.org/api/drupal/modules%21system%21system.api.php/function/hook_html_head_alter/7.x

igorpavlov:
@deb1990 you cannot disable the pinch on iOS anymore.

igorpavlov:
The PR moved to https://github.com/compucorp/civihr-employee-portal-theme/pull/333

civihr_employee_portal/civihr_employee_portal.module
igorpavlov: Done in the new PR

PR #462, opened 2018-04-23
PCHR-3573: Fix Staff Directory View (round 2)

mickadoo:
## Overview

Contacts with only a single inactive role or only a single inactive relationship were not displayed in the staff directory view

## Before

- Contacts with only inactive roles did not show in the staff directory view
- Contacts with only inactive relationships did not show in the staff directory view

image

## After

- All contacts with active contracts are shown regardless of role or relationship status

image

## Technical Details

The issue with filtering by role end was originally dealt with in PCHR-1890 by [removing](https://github.com/compucorp/civihr-employee-portal/commit/cd15baf1ed35d1530e08f65baf3b64bdd95acb54) the filter.

[This commit re-added a role filter](https://github.com/compucorp/civihr-employee-portal/commit/7f1b3d142c731198943e4c285db02ee3ad2e0150) and [this commit added a relationship filter](https://github.com/compucorp/civihr-employee-portal/commit/9d72c2669d7b0a161b84df17fc353d231a23f14c), which caused contacts with only inactive roles / relationships not to be displayed in the staff directory. 

It might seem that filtering out inactive roles and relationships seems like a good idea, but our staff directory is not for showing contacts with _only_ active roles and relationships. Active role and relationship data should be added to the view if it is available, but if it isn't the the contact should still be visible.

To fix this I reverted both commits mentioned above. These commits were part of a ticket to remove extra commas. We already had a function for this (added in [2015](https://github.com/compucorp/civihr-employee-portal/commit/57870a78029574241915c2427f96f3f68db20127#diff-16cb90fad42c8d4f2a7159c43f614960R358)) but since some of the fields were changed it wasn't working properly. This PR also fixes that function.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/views/civihr_employee_portal.views.inc
+++ b/civihr_employee_portal/views/civihr_employee_portal.views.inc
@@ -996,24 +996,25 @@ function _civihr_employee_portal_decode_vacancy_description_html(&$view, &$outpu
 }
 
 function removeAdditionalCommas($view) {
-  $fieldsToClean = array();
+  $replacements = [];
 
   foreach ($view->style_plugin->rendered_fields as $renderedField) {
-    $managerFieldValues = !empty($renderedField['last_name_1']) ? $renderedField['last_name_1'] : NULL;
-    $departmentFieldValues = !empty($renderedField['department']) ? $renderedField['department'] : NULL;
+    $fieldsToClean = [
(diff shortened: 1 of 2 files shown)


tunbola:
s/All contacts with active contacts/All contacts with active contracts

PR #474, opened 2018-05-31
FAR-265: Fix Date Filter Issue On Leave Reports and People Reports

tunbola:
## Overview
Recently, We have been having some performance issues with the Leave and Absence Reports on sites with particularly large leave data. On closer investigation, it was discovered that the Leave filters for this report does not work and it fetches the data set all at once and does not allow to fetch the data in chunks. 
The date filter for the people report works but the whole data is fetched on initial page load.

## Before
- The date filters on the Leave and Absence report does not work.
- The date filter for the people report works but the whole data is fetched on initial page load.

## After
- The date filters on the Leave and Absence reports was fixed and also on first load, the pivot table fetches result for the current year so as not to fetch the result set at once. The whole result set can still be fetched by clearing out the date filters.

- The date filter was not working because the filter parameter was added on the wrong field in the Leave and Absence report view. This was fixed by adding the correct `absence_date` field.

- The date filter parameter was added to the people report and defaulted to the current date so that the initial data fetched is for contracts active in the current date only.

datefilterafter

## Technical details

On initial report load, the whole result set was being fetched initially, but a new function was added that allows to fetch result set for current year for the leave and absence report. For the people report the report is fetched for contacts having active contracts at the current date.

```php

  HRReport.prototype.processFilterValues = function (filterValues) {
    if (filterValues === '?' && this.reportName === 'leave_and_absence') {
      return this.getDefaultFilterQueryForLeaveReport()
    }

    if (filterValues === '?' && this.reportName === 'people') {
      return this.getDefaultFilterQueryForPeopleReport()
    }

    return filterValues;
  }
```

The `initCurrentAbsencePeriodFilterDates` function was previously used to default the filter date values to the current period for the leave and absence report but it was removed since its no longer needed.

A new handler was added for the date filter for the leave and absence report since Drupal does the date comparison using timestamp by default. This handler allows to compare dates in `Y-m-d` format.
```php
class views_between_operation_dates_filter_handler extends views_handler_filter_date {
  function op_between($field) {
    // Use the substitutions to ensure a consistent timestamp.
    $query_substitutions = views_views_query_substitutions($this->view);
    $a = intval(strtotime($this->value['min'], $query_substitutions['***CURRENT_TIME***']));
    $b = intval(strtotime($this->value['max'], $query_substitutions['***CURRENT_TIME***']));
    $a = date('Y-m-d', $a);
    $b = date('Y-m-d', $b);
    // This is safe because we are manually scrubbing the values.
    // It is necessary to do it this way because $a and $b are formulas when using an offset.
    $operator = strtoupper($this->operator);
    $this->query->add_where_expression($this->options['group'], "$field $operator '$a' AND '$b'");
  }
}
```
The changes does not affect any Tests.
- [ ] Tests Pass

### DIFF
--- a/civihr_employee_portal/js/reports.js
+++ b/civihr_employee_portal/js/reports.js
@@ -407,6 +407,25 @@
   };
 
   /**
+   * Gets the default query parameters for fetching results for the
+   * Leave and absence report. It basically defaults to parameters
+   * for fetching results for the last 30 days starting from current
+   * date.
+   *
+   * @returns {String}
+   */
+  HRReport.prototype.getDefaultFilterQueryForLeaveReport = function () {
+    var toDate = moment().format("YYYY-MM-DD");
+    var fromDate = moment().subtract(1, 'months').format("YYYY-MM-DD");
+    var defaultFilterValues = [
+      { name: "absence_date_filter[min]", value: fromDate },
+      { name: "absence_date_filter[max]", value: toDate }
+    ];
+
+    return '?' + $.param(defaultFilterValues);
+  }
+
+  /**
    * Refresh JSON data and Pivot Tables using provided filter values
    *
    * @param string filterValues
@@ -416,6 +435,12 @@
     if (!this.jsonUrl) {
       return;
     }
+
+    //On load of page the form values are not appended to query string.
+    if(filterValues === '?' && that.reportName === 'leave_and_absence') {
+      filterValues = that.getDefaultFilterQueryForLeaveReport()
+    }
+
     $.ajax({
       url: this.jsonUrl + filterValues,
       error: function () {
@@ -454,6 +479,12 @@
     if (!this.tableUrl) {
       return;
     }
+
+    //On load of page the form values are not appended to query string.
+    if(filterValues === '?' && that.reportName === 'leave_and_absence') {
+      filterValues = that.getDefaultFilterQueryForLeaveReport()
+    }
+
     var tableDomId = this.getReportTableDomID();
     $.ajax({
       url: that.tableUrl + filterValues,
@@ -851,7 +882,9 @@
               }
 
               if (REPORT_NAME === 'leave_and_absence') {
-                initCurrentAbsencePeriodFilterDates().then(resolve, reject);
+                formFiltersStore.values.fromDate = moment().subtract(1, 'months').toDate();
+                formFiltersStore.values.toDate = moment().toDate();
+                resolve();
               } else {
                 formFiltersStore.values.date = new Date();
                 resolve();
--- a/civihr_employee_portal/js/reports.js
+++ b/civihr_employee_portal/js/reports.js
@@ -415,11 +415,11 @@
    * @returns {String}
    */
   HRReport.prototype.getDefaultFilterQueryForLeaveReport = function () {
-    var toDate = moment().format("YYYY-MM-DD");
-    var fromDate = moment().subtract(1, 'months').format("YYYY-MM-DD");
+    var toDate = moment().dayOfYear(1).add(1, 'year').subtract(1, 'day').format('YYYY-MM-DD');
+    var fromDate = moment().dayOfYear(1).format('YYYY-MM-DD');
     var defaultFilterValues = [
-      { name: "absence_date_filter[min]", value: fromDate },
-      { name: "absence_date_filter[max]", value: toDate }
+      { name: 'absence_date_filter[min]', value: fromDate },
+      { name: 'absence_date_filter[max]', value: toDate }
     ];
 
     return '?' + $.param(defaultFilterValues);
@@ -875,8 +875,8 @@
               }
 
               if (REPORT_NAME === 'leave_and_absence') {
-                formFiltersStore.values.fromDate = moment().subtract(1, 'months').toDate();
-                formFiltersStore.values.toDate = moment().toDate();
+                formFiltersStore.values.fromDate = moment().dayOfYear(1).toDate();
+                formFiltersStore.values.toDate = moment().dayOfYear(1).add(1, 'year').subtract(1, 'day').toDate();
                 resolve();
               } else {
                 formFiltersStore.values.date = new Date();
--- a/civihr_employee_portal/js/reports.js
+++ b/civihr_employee_portal/js/reports.js
@@ -409,8 +409,7 @@
   /**
    * Gets the default query parameters for fetching results for the
    * Leave and absence report. It basically defaults to parameters
-   * for fetching results for the last 30 days starting from current
-   * date.
+   * for fetching results for the current year
    *
    * @return {String}
    */
@@ -425,11 +424,26 @@
     return '?' + $.param(defaultFilterValues);
   }
 
+  /**
+   * Gets the default query parameters for fetching results for the
+   * People report. It basically defaults the date parameter to the
+   * current date.
+   *
+   * @return {String}
+   */
+  HRReport.prototype.getDefaultFilterQueryForPeopleReport = function () {
+    var currentDate = moment().format('YYYY-MM-DD');
+    var defaultFilterValues = [
+      { name: 'between_date_filter[value]', value: currentDate }
+    ];
+
+    return '?' + $.param(defaultFilterValues);
+  }
 
   /**
    * Processes the filter values and replaces with default values
-   * for the leave reports when the filter values does not have a
-   * query parameter appended.
+   * for the people and leave reports when the filter values does
+   * not have a query parameter appended.
    * On load of page the form values are not appended to query string
    * hence filterValues has a value of '?'
    *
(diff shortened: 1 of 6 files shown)


mickadoo:
I think maybe the ticket reference in this commit is wrong

> PCHR-3071: Fix wrong date filter field in leave and absence report view.

civihr_employee_portal/js/reports.js
```diff
@@ -416,6 +435,12 @@
     if (!this.jsonUrl) {
       return;
     }
+
+    //On load of page the form values are not appended to query string.
+    if(filterValues === '?' && that.reportName === 'leave_and_absence') {
```
mickadoo: Not sure about JS coding standard but should there be a space after `if`?

civihr_employee_portal/js/reports.js
```diff
@@ -454,6 +479,12 @@
     if (!this.tableUrl) {
       return;
     }
+
+    //On load of page the form values are not appended to query string.
+    if(filterValues === '?' && that.reportName === 'leave_and_absence') {
```
mickadoo: Why is this being done twice, seems like code repetition

civihr_employee_portal/js/reports.js
```diff
@@ -851,7 +882,9 @@
               }
 
               if (REPORT_NAME === 'leave_and_absence') {
-                initCurrentAbsencePeriodFilterDates().then(resolve, reject);
+                formFiltersStore.values.fromDate = moment().subtract(1, 'months').toDate();
```
mickadoo: is `initCurrentAbsencePeriodFilterDates` still used anywhere?

civihr_employee_portal/views/includes/views_between_operation_dates_filter_handler.inc
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * @file
+ * A filter that allows the between operator to compare dates in 'Y-m-d' format
+ * rather than the default timestamp comparison since absence dates are stored
+ * in 'Y-m-d' format in the db.
+ */
+
+class views_between_operation_dates_filter_handler extends views_handler_filter_date {
```
mickadoo: It's annoying that Drupal doesn't have default handlers for this kind of thing :-(

civihr_employee_portal/views/includes/views_between_operation_dates_filter_handler.inc
```diff
@@ -0,0 +1,24 @@
...
+  function op_between($field) {
+    // Use the substitutions to ensure a consistent timestamp.
+    $query_substitutions = views_views_query_substitutions($this->view);
+    $a = intval(strtotime($this->value['min'], $query_substitutions['***CURRENT_TIME***']));
+    $b = intval(strtotime($this->value['max'], $query_substitutions['***CURRENT_TIME***']));
+
+    $a = date('Y-m-d', $a);
```
mickadoo: Do you think it would be clearer to use more descriptive variable names?

civihr_employee_portal/views/includes/views_between_operation_dates_filter_handler.inc
```diff
@@ -0,0 +1,24 @@
...
+    $b = date('Y-m-d', $b);
+
+    // This is safe because we are manually scrubbing the values.
```
mickadoo: Can you explain this comment a bit more please?

civihr_employee_portal/views/includes/views_between_operation_dates_filter_handler.inc
tunbola: Well, I copied this over from the base class and only added two extra lines of code.
what do you think?

civihr_employee_portal/views/includes/views_between_operation_dates_filter_handler.inc
tunbola: Copied the comment from the base class.

civihr_employee_portal/js/reports.js
tunbola: Nope, I guess I'll remove it.

civihr_employee_portal/js/reports.js
tunbola: Looking at the other conditions in the file, looks like there should be a space.

mi-kon:
@tunbola can it default to the current absence period, rather than last 30 days?

civihr_employee_portal/views/includes/views_between_operation_dates_filter_handler.inc
mickadoo: Alright, I guess the problem is with the base class then. It might still be nice to change it to `$to` and `$from` or something but I'll leave it to you

civihr_employee_portal/views/includes/views_between_operation_dates_filter_handler.inc
mickadoo: I'm not sure it makes sense to leave the comment as it is then, even if it is copy and paste. If it doesn't make sense can you remove it, or maybe it does make sense and I just don't understand it.

tunbola:
@mi-kon , Defaulted the filter to current year as discussed

igorpavlov:
@tunbola there are no technical explanation on what has been done, could you please include, like here: https://github.com/civicrm/civihr/pull/2643

You don't need to include the whole code, just explanations and snippets.

civihr_employee_portal/js/reports.js
```diff
@@ -406,6 +406,43 @@
...
+   * Gets the default query parameters for fetching results for the
+   * Leave and absence report. It basically defaults to parameters
+   * for fetching results for the last 30 days starting from current
+   * date.
+   *
+   * @returns {String}
+   */
+  HRReport.prototype.getDefaultFilterQueryForLeaveReport = function () {
+    var toDate = moment().dayOfYear(1).add(1, 'year').subtract(1, 'day').format('YYYY-MM-DD');
+    var fromDate = moment().dayOfYear(1).format('YYYY-MM-DD');
```
igorpavlov: Please swap `toDate` with `fromDate`, it would be easier to read.

civihr_employee_portal/js/reports.js
```diff
@@ -406,6 +406,43 @@
...
+    var defaultFilterValues = [
+      { name: 'absence_date_filter[min]', value: fromDate },
+      { name: 'absence_date_filter[max]', value: toDate }
+    ];
+
+    return '?' + $.param(defaultFilterValues);
```
igorpavlov: Any chance this is going to be populated to URL already with `?querystring=smth` ?

civihr_employee_portal/js/reports.js
```diff
@@ -406,6 +406,43 @@
     this.applyFilters();
   };
 
+  /**
+   * Gets the default query parameters for fetching results for the
+   * Leave and absence report. It basically defaults to parameters
+   * for fetching results for the last 30 days starting from current
+   * date.
+   *
+   * @returns {String}
```
igorpavlov: In JS we use `@return` (though both are OK, but we use just "return")

civihr_employee_portal/js/reports.js
```diff
@@ -406,6 +406,43 @@
...
+    return '?' + $.param(defaultFilterValues);
+  }
+
+
+  /**
+   * Processes the filter values and replaces with default values
+   * for the leave reports when the filter values does not have a
+   * query parameter appended.
+   *
+   * @returns string {String}
```
igorpavlov:
```
* @param  {String} filterValues
* @return {String}
```

civihr_employee_portal/js/reports.js
```diff
@@ -406,6 +406,43 @@
...
+   */
+  HRReport.prototype.processFilterValues = function (filterValues) {
+    var that = this;
```
igorpavlov: Do we really need `that` here? The scope does not change inside the func.

civihr_employee_portal/js/reports.js
```diff
@@ -406,6 +406,43 @@
...
+    //On load of page the form values are not appended to query string.
```
igorpavlov: Please add space after `//`

civihr_employee_portal/js/reports.js
tunbola: Nope, there's a check before this function is called, so all good.

civihr_employee_portal/js/reports.js
```diff
@@ -431,13 +431,14 @@
    * for the leave reports when the filter values does not have a
    * query parameter appended.
    *
-   * @returns string {String}
+   * @param  {String} filterValues
+   *
```
igorpavlov: No need in this empty line space.

civihr_employee_portal/js/reports.js
```diff
@@ -431,13 +431,14 @@
...
+   * @return {String}
    */
   HRReport.prototype.processFilterValues = function (filterValues) {
-    var that = this;
-    //On load of page the form values are not appended to query string.
-    if (filterValues === '?' && that.reportName === 'leave_and_absence') {
-      return that.getDefaultFilterQueryForLeaveReport()
+    // On load of page the form values are not appended to query string.
```
igorpavlov: Now because we reduced the code, the comment can actually go to the function description.

PR #478, opened 2018-06-06
PCHR-3706: Add default Resource Types

deb1990:
## Overview
As part of this PR the following default HR Resource Types are added
- Policy
- Handbook
- Forms
- Training Manual

## Screenshot
screen shot 2018-06-06 at 1 26 34 pm

## Technical Details
1. `HR Resource Type` was previously added via feature, and both `civihr_employee_portal_features` and `civihr_employee_portal` has similar weight so the hooks of feature was run after the module. For that reason it was not possible to add the terms before defining the vocabulary. So the vocabulary is removed from feature, and created from module file.

2. A helper class `TaxonomyHelper` is defined, which shares the code for adding the `HR Resource Type` vocabulary and the default terms between `civihr_employee_portal.install` and `civihr_employee_portal.module` file.

3. In the `civihr_employee_portal_update_7029` upgrader the `TaxonomyHelper::createHRRecourceTypeVocabulary()` function is not called because for older builds the vocabulary is already present in the DB.


civihr_employee_portal/src/Helpers/TaxonomyHelper.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+namespace Drupal\civihr_employee_portal\Helpers;
+
+use stdClass;
```
davialexandre: this is not necessary

civihr_employee_portal/src/Helpers/TaxonomyHelper.php
```diff
@@ -0,0 +1,59 @@
...
+
+class TaxonomyHelper {
+
+  /**
+   * Create HR Resource Type Vocabulary
+   */
+  public static function createHRRecourceTypeVocabulary() {
```
davialexandre: s/Recource/Resource/

deb1990: Oops :(

civihr_employee_portal/src/Helpers/TaxonomyHelper.php
deb1990: Without this I got error in the build process, something like `stdClass' not found in ...`

davialexandre: This happens because of the namespace declaration on line 3. To fix the error, change line 49 to `$term = new \stdClass()`;

PR #487, opened 2018-06-12
PCHR-3570: Assign Default Roles For Permission to View Footer Status

backstageel:
## Overview
Adding default Permissions to view system status no footer, and using upgrader to revert the default 
permissions
This is part of this pull request: https://github.com/compucorp/civihr/pull/2674

## Before
All users could view the status on footer. 

## After
Only root admins can view the status of on the footer


mickadoo:
The commit message "Footer changes" isn't very descriptive, can you change it to something that reflects what it does specifically? For example "Assign default roles for permission to view footer status"

If you need to expand on that it's fine, I usually try to keep the first message short, leave a blank line, and then expand on it if needed. There is something called the [50/72 rule](https://medium.com/@preslavrachev/what-s-with-the-50-72-rule-8a906f61f09c) but since we have the issue number in the title it's hard to fit inside 50 so I usually try to keep to 72 for the first line. Have a look [here](https://github.com/compucorp/civihr-employee-portal/commit/a8de80ca8ec0921430401bd1e8bb48cddb5cba47) for an example

PR #491, opened 2018-06-15
PCHR-3837: onboarding wizard improvements

powdevel:
## Requirement 11

Move the following fields to the Personal Details section:
- Gender
- Marital Status
- NI /SSN 
- "I am applying for NI / SSN" checkbox

image

## Requirement 12

Rename section heading from "Address" to "Home Address"
This address should map to personal address in the HR Admin side. 

image

## Requirement 13

HR Admin, Root and Local HR Admin users should not have to fill the onboarding wizard at any point. They can update their data from the HR Admin side. 
Add a button to the bottom of the /customize-onboarding-wizard page a link "Test out the onboarding wizard as an administrator" - this should link to the onboarding wizard as an administrator. 

image

## Requirement 14
The labels Home Phone Number and Home Email should be changed to display as Personal Phone and Personal Email. They will map to the personal phone and personal email on the HR Admin side.

image

### Tecnical note
To be able to update the "onboarding wizard" I had to manually edit grid (group id) value in the exported code of the "onboarding wizard", to avoid trying to insert duplicated PK on table webform_conditional when applying the update. This was happening:
image
After the change, tests were done. Data was correct. No issues were detected.


civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
```diff
@@ -22,6 +23,7 @@ public function build() {
     $form[self::INTRODUCTION_TEXT_KEY] = $this->getIntroElement();
     $form[self::CAROUSEL_OPTIONS_KEY] = $this->getCarouselElement();
     $form['actions']['cancel'] = $this->getCancelButton();
+    $form['actions']['save-and-test'] = $this->saveAndTestOutThisForm();;
```
davialexandre: Please remove the trailing `;`

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
```diff
@@ -45,6 +47,12 @@ public function onSubmit($form, &$formState) {
 
     $this->saveLogo($formState);
     $this->updateCarouselContent($formState);
+
+    // sends user to the first step of the onboarding wizard
+    // which is just node edit page with custom theming
+    if ($this->clickedSaveAndTest($formState)) {
+      drupal_goto('/user/'. user_uid_optional_load()->uid .'/edit', ['query' => ['testing'=>1]] );
```
davialexandre: Please add spaces between all the operators and operands here.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5024,8 +5024,15 @@ function _civihr_employee_portal_should_do_onboarding($account) {
 
   $canViewOnboardingForm = node_access('view', $onboardingWebform, $account);
   $submissions = WebformHelper::getUserSubmissionsByTitle($account, $formName);
+  $user_wants_to_test = isset($_GET['testing']);
+  // user who can edit it's contact doesn't need this form
+  if (!$user_wants_to_test && user_access('edit my contact')) {
+    return false;
+  }
```
davialexandre: I can't see this as a requirement either in the ticket or in the specification. Where did it come from?

If it is really necessary, what do you think of moving it right to the beginning of this function? This way, we could avoid all the unnecessary calls to load the webform since the user won't have access to it anyway

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5024,8 +5024,15 @@ function _civihr_employee_portal_should_do_onboarding($account) {
 
   $canViewOnboardingForm = node_access('view', $onboardingWebform, $account);
   $submissions = WebformHelper::getUserSubmissionsByTitle($account, $formName);
+  $user_wants_to_test = isset($_GET['testing']);
+  // user who can edit it's contact doesn't need this form
```
davialexandre: s/it's/its/

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5024,8 +5024,11 @@ function _civihr_employee_portal_should_do_onboarding($account) {
 
   $canViewOnboardingForm = node_access('view', $onboardingWebform, $account);
   $submissions = WebformHelper::getUserSubmissionsByTitle($account, $formName);
+  $user_wants_to_test = isset($_GET['testing']);
```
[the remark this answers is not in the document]
powdevel: I choose "testing" because the context is defined by the URL, from my point of view. "test_onboarding_wizard" seems repetitive to me, but I'm confident on your criteria so I will change it to the latter option.

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
powdevel: Right, thanks, I will start using PHP-CS for that kind of formatting.

civihr_employee_portal/civihr_employee_portal.module
powdevel: Is related to this: "HR Admin, Root and Local HR Admin users should not have to fill the onboarding wizard at any point. They can update their data from the HR Admin side." The ones who can "edit my contact" are not required to do the "Onboarding Wizard".

"If it is really necessary, what do you think of moving it right to the beginning of this function?" Totally agree, thanks.

civihr_employee_portal/civihr_employee_portal.module
powdevel: Right :+1:

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
powdevel: Sure

civihr_employee_portal/civihr_employee_portal.module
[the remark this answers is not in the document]
powdevel: I will change variables to camelCase to follow the way that the function was written, although snake_case provides a better readability in my opinion.

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
davialexandre: @powdevel Buildkit already includes this tool called [civilint](https://docs.civicrm.org/dev/en/latest/tools/civilint/), which is just PHP-CS configured for the code standards used in CiviCRM. You might want to have a look at it

civihr_employee_portal/civihr_employee_portal.module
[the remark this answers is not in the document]
davialexandre: @powdevel now with your explanation, "testing" does sounds better and less redundant. No need to change that

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5015,6 +5015,12 @@ function civihr_employee_portal_user_login(&$edit, $account) {
  * @return bool
  */
 function _civihr_employee_portal_should_do_onboarding($account) {
+  $userWantsToTest = isset($_GET['test_onboarding_wizard']);
```
mickadoo: Will the format for this query parameter just be `?test_onboarding_wizard` or will it be `?test_onboarding_wizard=1`? If it's the latter you might want to check the actual value here, not just whether it's set.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5025,7 +5031,9 @@ function _civihr_employee_portal_should_do_onboarding($account) {
   $canViewOnboardingForm = node_access('view', $onboardingWebform, $account);
   $submissions = WebformHelper::getUserSubmissionsByTitle($account, $formName);
 
-  return $canViewOnboardingForm && empty($submissions);
+  return
```
mickadoo: I think this conditional is getting too long, can you separate it into something else like 

```
$canCustomizeWizard = user_access('customize welcome wizard');
$isAdminTest = $userWantsToTest && $canCustomizeWizard;
```

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
```diff
@@ -45,6 +47,12 @@ public function onSubmit($form, &$formState) {
 
     $this->saveLogo($formState);
     $this->updateCarouselContent($formState);
+
+    // sends user to the first step of the onboarding wizard
+    // which is just node edit page with custom theming
+    if ($this->clickedSaveAndTest($formState)) {
+      drupal_goto('/user/' . user_uid_optional_load()->uid . '/edit', ['query' => [ 'test_onboarding_wizard' => 1 ]]);
```
mickadoo: This line is very long, can you separate it into different variables, like `$userEditUrl` and `$queryParams`?

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
```diff
@@ -177,6 +185,18 @@ private function clickedCancel($formState) {
...
 
+  /**
+   * On form submission checks if the button pressed was "SaveAndTest"
+   *
+   * @param array $formState
+   *
+   * @return bool
+   */
+  private function clickedSaveAndTest($formState) {
+    $clickedButton = \CRM_Utils_Array::value('clicked_button', $formState);
```
mickadoo: I think we usually have an empty line before `return`

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
```diff
@@ -190,4 +210,17 @@ private function adjustSaveButton($form) {
 
     return $form;
   }
+
+  /**
+   * Returns the structure for the button which saves and go to test
+   *
+   * @return array
+   */
+  private function saveAndTestOutThisForm() {
```
mickadoo: I'm not sure this is a good name for this function, it doesn't "save and test" this form, it returns the button for this, so maybe something like 'getSaveAndTestButton'

civihr_employee_portal/civihr_employee_portal.module
powdevel: Without 1 as value to the parameter, Drupal will not set the parameter when making the link, that is the reason for 1. In fact is it sufficient to have the parameter "test_onboarding_wizard" set because 1 doesn't have any meaning, what has meaning is the existence of the parameter. For example there is no "test_onboarding_wizard=0" to not test.

civihr_employee_portal/civihr_employee_portal.module
powdevel: Right, thanks!

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
powdevel: Sure, no problem. Thanks.

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
powdevel: Totally agree.

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
powdevel: Ok, I will add that line.

civihr_employee_portal/src/Forms/OnboardingWizardCustomizationForm.php
powdevel: Thanks, good advice. I will check that tool.

civihr_employee_portal/civihr_employee_portal.module
mickadoo: I think that it's more common to have query parameters in the form `?a=b&c=d` instead of `?c&d`, what do you think? If that form is more common it should be an easy check to change from `isset` to checking `isset` as well as the value

powdevel: It is an easy change. I will change it now.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5015,6 +5015,12 @@ function civihr_employee_portal_user_login(&$edit, $account) {
  * @return bool
  */
 function _civihr_employee_portal_should_do_onboarding($account) {
+  $userWantsToTest = isset($_GET['testing']) && $_GET['testing']==1;
```
mickadoo: Not sure about the formatting here, think there should be spaces around `==`

powdevel: You are right, I will correct it.

PR #497, opened 2018-06-20
PCHR-3861: Fill data layer container on module init

AkA84:
The ssp module will now populate Google Analytic's `dataLayer` with the user's roles (minus the "authenticated user", which is a given).

The user roles, stored in the `userRoles` property, will be then used as custom dimension to create GA reports

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -552,12 +552,12 @@ function civihr_employee_portal_init() {
 function _civihr_employee_portal_set_up_google_analytics_data_layer() {
   global $user;
 
-  datalayer_add(array(
+  datalayer_add([
     // The list of roles of the current user
     'userRoles' => join(';', array_filter($user->roles, function ($role) {
       return $role != 'authenticated user';
     })),
-  ));
+  ]);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -494,6 +494,7 @@ function civihr_employee_portal_init() {
   // calling this inside hook_init so we could access
   // some drupal functions such as user_access
   _user_redirection();
+  _civihr_employee_portal_set_up_google_analytics_data_layer();
```
davialexandre: @AkA84 hook_init() is also called from the command line (for example, when you're running drush commands for the site). In that context, a few things used by this function might not be available (a logged in user, for example). Could you please try running a few drush commands from a site pointing to this branch and see if you will get any errors? I suggest you try things like, `drush uli`, `drush pm-list`, `drush cc`, `drush fr`

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -545,6 +546,20 @@ function civihr_employee_portal_init() {
   }
 }
 
+/**
+ * Sets up the content of GA's data layer container
+ */
+function _civihr_employee_portal_set_up_google_analytics_data_layer() {
+  global $user;
+
+  datalayer_add(array(
```
davialexandre: Please use the short array syntax here

civihr_employee_portal/civihr_employee_portal.module
AkA84: that's a good point, thanks. I tried several `drush` commands on my local and got no error, seems to be working fine

PR #505, opened 2018-06-28
PCHR-3925: Check if datalayer module exists before using it

AkA84:
## Overview
The SSP module tried to use the `datalayer_add` regardless of the presence and/or status of the "datalayer" module, resulting in the SSP breaking if the module was disabled or non existing

## Technical Details
The PR simply adds a check to determine if the "datalayer" module exists, before using its function

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -552,6 +552,10 @@ function civihr_employee_portal_init() {
 function _civihr_employee_portal_set_up_google_analytics_data_layer() {
   global $user;
 
+  if (!function_exists('datalayer_add')) {
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -552,6 +552,10 @@ function civihr_employee_portal_init() {
 function _civihr_employee_portal_set_up_google_analytics_data_layer() {
   global $user;
 
+  if (!function_exists('datalayer_add')) {
```
davialexandre: Usually, what we do is to check if the module that declares the function exists, not the function itself. It's very unlikely that another module will have a function with this same name, but not impossible. Please change this to `if (!module_exists('data_layer'))`;

PR #510, opened 2018-07-10
PCHR-3932: Fix Regional HR Admin Permissions

deb1990:
## Overview
In `/node/add/hr-documents` page, the Revision section is visible for Regional HR Admin. But it should not. This PR fixes that bug.

## Before
b

## After
a

## Technical Details
The permissions of **Regional HR Admin** and **HR Admin** were not same and hence this bug. Both the roles should have the exact same permissions. So changed all permissions of Regional HR Admin to exactly mimic HR Admin.

## Comments
**For Devs:**  To test this, open permissions page in drupal, HR Admin and Regional HR Admin should have the same permissions.


deb1990:
The PR was wrong, as I misunderstood the requirements. PR has been reverted from PCHR-3744-ssp-redesign-epic branch by rebasing and force pushing.

PR #516, opened 2018-07-15
PCHR-3964: Modifications for Personal Details Webform and Block

powdevel:
## Overview
Options for gender were manually selected on the webform configuration, which results an error in webform exportation for select options in different environments, specifically CiviHR testing site and sites built with buildkit tool.

For NI / SSN field are being showed in webforms and included in My Details block, but there is an issue with the way that these fields are created on CiviHR test site. For now we don’t display the field either in a view or in a webform.

## Before
* As webform export code for My Personal Details using custom IDs for gender options, when saving the webform "My Personal Details" the gender disappears from My Details page because it is saving an unknown option for the current environment.

* Displaying NI / SSN fields on My Personal Details webform and view

## After
* As webform export code for My Personal Details using live options, taken from current environment when saving the webform "My Personal Details" the chosen gender is saved correctly and shows in My Details page.

* Not displaying NI / SSN fields on My Personal Details webform and view
image

## Technical Details
NI / SSN fields were removed because there was an issue with the generation of the names and IDs related to those fields, which was producing that views can't find those fields. This issue is due to how these fields are created in the Test CiviHR site in comparison with buildkit.


civihr_employee_portal/features/node_export_files/mydetails_personal_details.export
```diff
@@ -220,16 +220,19 @@ array(
...
-              'aslist' => 1,
+              'multiple' => 0,
+              'title_display' => 'before',
+              'description_above' => 0,
               'private' => 0,
-              'multiple' => NULL,
+              'wrapper_classes' => '',
+              'css_classes' => '',
+              'aslist' => 1,
+              'civicrm_live_options' => '1',
```
davialexandre: I would expect this to be the only change in this file, but I can actually see a lot more updates for other options. @powdevel can you confirm that this is not changing anything we don't want to?

davialexandre:
@powdevel please also explain in the PR description what is the issue with the NI/SSN field so that people can understand why it is being removed

civihr_employee_portal/features/node_export_files/mydetails_personal_details.export
powdevel: @davialexandre This PR also removes NI/SSN fields. See this link: https://compucorp.atlassian.net/browse/PCHR-3967?focusedCommentId=73007&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-73007

davialexandre: @powdevel I know that, but this does not answer my question. I'm talking about the changes in the 0a17d3eaa9bad3b9e1e75912f684ef12f89b4606 which are only about "setting the live option when exporting the webform". If that was the only thing done in that commit, then why we have all of these other changes it in?

powdevel: @davialexandre that is what the export returned. I think the changes you are looking are related to field that was changed, but as webform has many sections, the form, the notifications, the civicrm, etc one simple change will update more than one section.

PR #517, opened 2018-07-16
PCHR-3969: Move City Above Postal Code in My Home Address Webform

powdevel:
## Overview
In Home Address webform from My Details page Postal code is showing above city. It should be the opposite, City must be above Postal Code.

## Before
image

## After
image


deb1990:
Merging on behalf of @powdevel as it is urgent for the release.

PR #522, opened 2018-07-23
PCHR-4010: Update path for reqangular.min.js

deb1990:
## Overview
As part of this PR the path for `reqangular.min.js` is updated for the Reports page.

## Before
2018-07-23 at 3 10 pm

2018-07-23 at 3 10 pm

## After
2018-07-23 at 3 06 pm

2018-07-23 at 3 07 pm

## Technical Details
In PCHR-3942 (https://github.com/compucorp/civihr/pull/2772) the path for `reqangular.min.js` was changed, but the same was not updated in `civihr_employee_portal.module`, which broke the Reports pages.
The path has been fixed.

---

- [x] Manual Tests Pass


igorpavlov:
I have a feeling this is a regression bug. Could you please mention the ticket the regression has introduced and add an appropriate label?

PR #526, opened 2018-07-26
PCHR-4037: fix views import of custom fields at the creation of views

powdevel:
## Overview
Depending on the creation of CiviCRM custom fields and CiviCRM custom tables their names may vary on different installations. Examples:

- From `custom_field_25` to `custom_field_10001`
- From `custom_table_38` (exported table name from another site) to `custom_table_51` (existing table on current site)
 
When these fields/tables names change it may produce bad references when importing views that use these names, which generates empty queries due to missing fields or tables, which in turn causes views not showing the expected results.

## When it happens
It happens when a Custom Group/Field is created using the CiviCRM API. The table and field names will get the next ID available at the moment the API call is made and there's no way to guarantee that this next ID will be the same in all sites. This is why different sites have different IDs.

## Solution

This PR proposes a work around to fix this issue by altering the structure of the view that is going to be imported to match existing names of fields and tables based on the names of the missing fields and tables.

Implementing `hook_views_default_views` the view structure can be altered previous import. 

- Every table not found in current installation is searched against the CiviCRM views intergration array defined in `settings.php`, to find a similar table name, if found, the view structure values are changed accordingly.

- Every field not found in current installation is searched in the table where the field belongs, to find a similar field name, if found, the view structure values are changed accordingly.

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -15,6 +15,7 @@ use Drupal\civihr_employee_portal\Page\HRDetailsPage;
 use Drupal\civihr_employee_portal\Webform\WebformTransferService;
 use Drupal\civihr_employee_portal\Helpers\LinkProvider;
 use Drupal\civihr_employee_portal\Helpers\TaxonomyHelper;
+use Drupal\civihr_employee_portal\View\CustomFieldsMappingFixer;
 
 /**
  * Implements hook_install().
@@ -854,6 +855,7 @@ function civihr_employee_portal_views_default_views() {
 
     if (isset($view)) {
       $views[$view->name] = $view;
+      CustomFieldsMappingFixer::fixMapping($view);
     }
   }
 
--- a/civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
+++ b/civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
@@ -1,0 +1,164 @@
+<?php
+
+namespace Drupal\civihr_employee_portal\View;
+
+/**
+ * To match unknown table names and field names when importing views
+ * with custom fields and custom groups from CiviCRM.
+ * Replaces unknown fields names and unknown tables like:
+ * - exported [name_of_the_field]_[ID1] with existing [name_of_the_field]_[ID2]
+ * - exported [name_of_the_table]_[ID3] with existing [name_of_the_table]_[ID4]
+ * ... where ID1, ID2, ID3 & ID4 are different numbers created by different
+ * mechanisms thus producing different IDs.
+ */
+class CustomFieldsMappingFixer {
+
+  /**
+   * To fix mapping on Views import of CiviCRM custom fields
+   *
+   * @param view $view
+   *     the view that is being imported / recreated
+   */
+  public static function fixMapping($view) {
+    foreach ($view->display as $display) {
+      foreach ($display->display_options as &$displayOption) {
+        if (!is_array($displayOption)) {
+          continue;
+        }
+
+        self::processDisplayOptions($displayOption);
+      }
+    }
+  }
+
+  /**
+   * Process all display options for an exported view only when it is related
+   * to a field and the field has a number at the end of its name, which is
+   * the pattern for custom fields from CiviCRM
+   *
+   * @var array
+   *     Display option a exported view structure
+   */
+  private static function processDisplayOptions(&$displayOption) {
+    foreach ($displayOption as &$option) {
+      // only continue if this option is about a field
+      if (!isset($option['field'])) {
+        continue;
+      }
+
+      $optionIdParts = explode('_', $option['field']);
+      $finalIdPart = array_pop($optionIdParts);
+      if (count($optionIdParts) === 0 || !intval($finalIdPart)) {
+        continue;
+      }
+
+      self::processOption($option);
+    }
+  }
+
+  /**
+   * To process a display option for a field in view
+   *
+   * @var array $option
+   *     structure that contains an option for a field in view
+   */
+  private static function processOption(&$option) {
+    // checking if this table is part of the civicrm views integration
+    $tableName = self::getCiviCRMtableName($option['table']);
+    $prefix = $tableName ? self::getPrefix($tableName) : false;
+    if ( $prefix ) {
+      $option['table'] = $tableName;
+      $columns = self::getColumnsFromTable($prefix . $tableName);
+      // if the table where the current field belongs exists does not have it
+      // then we can search for similar fields in the same table and make replacements
+      if (!isset($columns[$option['field']])) {
+        $fieldNameBase = self::removeLastNumberOfString($option['field']);
+        $pos = array_search($fieldNameBase, $columns);
+        if ($pos !== FALSE) {
+          $option['field'] = array_search($columns[$pos], $columns);
+        }
+      }
+    }
+  }
+
+  /**
+   * Removes the number in end of a string
+   * In this context that number is always separated by an underscore
+   *
+   * @param  string $string
+   *     the string which could have underscores or not
+   *
+   * @return string
+   *     the string without the number at the end,
+   *     if no number found at the end returns the same string
+   */
+  private static function removeLastNumberOfString($string) {
+    $pieces = explode('_', $string);
+    $lastIndex = count($pieces) - 1;
+    return $lastIndex > 0 && intval(array_pop($pieces)) ? implode('_', $pieces) : $string;
+  }
+
+  /**
+   * Return an existing CiviCRM table based on the base name of the table,
+   * where "base name of the table" is the name of the table without the number
+   * at the end
+   *
+   * @param string $tableName
+   *     this is the table name exported by views
+   *
+   * @return string
+   *     the name of the existing table if found or an empty string
+   */
+  private static function getCiviCRMtableName($tableName) {
+    global $databases;
+    static $tableNamesMap = [];
+    // creates a map with table names without IDs to existing table names with IDs
+    if (!$tableNamesMap) {
+      $tablesWithIds = array_keys($databases['default']['default']['prefix']);
+      foreach ($tablesWithIds as $tableWithId) {
+        $tableNamesMap[self::removeLastNumberOfString($tableWithId)] = $tableWithId;
+      }
+    }
+
+    $exportedTableBaseName = self::removeLastNumberOfString($tableName);
+    $tableFound = isset($tableNamesMap[$exportedTableBaseName]);
+    return $tableFound ? $tableNamesMap[$exportedTableBaseName] : '';
+  }
+
+  /**
+   * To obtain the prefix for a table in the current installation
+   *
+   * @param string $tableName
+   *     the name of the table
+   *
+   * @return string
+   *     the prefix for the table
+   */
+  private static function getPrefix($tableName) {
+    global $databases;
+    return $databases['default']['default']['prefix'][$tableName];
+  }
+
+  /**
+   * Gets the columns from a table keyed by the name of the column without
+   * the sequential at the end
+   *
+   * @param string $table
+   *     the name of the table
+   *
+   * @return array
+   *     the names of the columns of the table
+   */
+  private static function getColumnsFromTable($table) {
+    static $columns = [];
+    if (!isset( $columns[$table])) {
+      $columns[$table] = [];
+      $res = db_query('SHOW COLUMNS FROM ' . $table)->fetchAll();
+      foreach ($res as $field) {
+        $columns[$table][$field->Field] = self::removeLastNumberOfString($field->Field);
+      }
+    }
+
+    return $columns[$table];
+  }
+}


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+      }
+      foreach ($display_option as &$option) {
+        // only continue if this option is about a field
+        if (!isset($option['field'])) {
+          continue;
+        }
+        $option_id_parts = explode('_', $option['field']);
+        $final_id_part = array_pop($option_id_parts);
+        // it must contain underscores and a number at the end to be checked
+        if (!$option_id_parts > 0 || !intval($final_id_part)) {
```
davialexandre: `$option_id_parts` is an array. I don't think it makes sense to compare if it is greater than a number. Also, why are you negating the array? I think that what you've meant here is `count($option_id_parts) === 0)` or `empty($option_id_parts)`.

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+          continue;
+        }
+        // checking if this table is part of the civicrm views integration
+        if ( isset($databases['default']['default']['prefix'][$option['table']]) ) {
+          $prefix_table = $databases['default']['default']['prefix'][$option['table']];
+          $table = $prefix_table.$option['table'];
```
davialexandre: Please always add spaces between operators

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+          $columns = civihr_employee_portal_get_columns_from_table($table);
+          // if the table where the current field must exists does not have the field
+          // it could be fixed...
+          if (!isset($columns[$option['field']])) {
+            $field_name_base = civihr_employee_portal_field_name_without_sequential($option['field']);
+            // searching field by base name in columns
+            $pos = array_search($field_name_base, $columns);
+            // if found then replace
+            if ($pos) {
```
davialexandre: Please check the php documentation about `array_search`. I don't think it's safe to use the value returned by it in this way

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+        }
+      }
+    }
+  }
+}
+
+/**
+ * Removes the number at the end of the name of a field
+ *
+ * @param  string $field
```
davialexandre: It looks like there is one too many space between `@param` and the type

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+ *     name of the field
+ *
+ * @return string
+ *     the name of the field without the number at the end
+ */
+function civihr_employee_portal_field_name_without_sequential($field) {
+  $field_parts = explode('_', $field);
+  if ($field_parts > 1) {
```
davialexandre: `$field_parts` is an array. This comparison doesn't make sense

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+    array_pop($field_parts);
+  }
+  return implode('_', $field_parts);
+}
+
+/**
+ * Gets the columns from a table keyed by the name of the column without
+ * the sequential at the end
+ *
+ * @param  string $table
```
davialexandre: Too many spaces here too

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+ *     the name of the table
+ *
+ * @return array
+ *     the names of the columns of the table
+ */
+function civihr_employee_portal_get_columns_from_table($table) {
+  static $columns = [];
+  if (!isset( $columns[$table])) {
+    $columns[$table] = [];
+    $res = db_query('SHOW COLUMNS FROM '.$table)->fetchAll();
```
davialexandre: Spaces between operators

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+        if (!$option_id_parts > 0 || !intval($final_id_part)) {
+          continue;
+        }
+        // checking if this table is part of the civicrm views integration
+        if ( isset($databases['default']['default']['prefix'][$option['table']]) ) {
+          $prefix_table = $databases['default']['default']['prefix'][$option['table']];
+          $table = $prefix_table.$option['table'];
+          $columns = civihr_employee_portal_get_columns_from_table($table);
+          // if the table where the current field must exists does not have the field
+          // it could be fixed...
```
davialexandre: The way this is written, it gives the impression that it cannot be fixed. Also, avoid using ellipses when writing documentation. They indicate the omission of words, an unfinished thought or a pause. None of these make sense here

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+          if (!isset($columns[$option['field']])) {
+            $field_name_base = civihr_employee_portal_field_name_without_sequential($option['field']);
+            // searching field by base name in columns
+            $pos = array_search($field_name_base, $columns);
+            // if found then replace
```
davialexandre: this comment is unnecessary

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+        // checking if this table is part of the civicrm views integration
+        if ( isset($databases['default']['default']['prefix'][$option['table']]) ) {
+          $prefix_table = $databases['default']['default']['prefix'][$option['table']];
+          $table = $prefix_table.$option['table'];
+          $columns = civihr_employee_portal_get_columns_from_table($table);
+          // if the table where the current field must exists does not have the field
+          // it could be fixed...
+          if (!isset($columns[$option['field']])) {
+            $field_name_base = civihr_employee_portal_field_name_without_sequential($option['field']);
+            // searching field by base name in columns
```
davialexandre: this comment is unnecessary

civihr_employee_portal/civihr_employee_portal.module
davialexandre: The table names are also prefixed with ids. Did you try do check if these IDs can also differ from site to site? I think they can, depending on how the custom groups are created

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -854,12 +854,100 @@ function civihr_employee_portal_views_default_views() {
...
+            // if found then replace
+            if ($pos) {
+              $option['field'] = array_search($columns[$pos], $columns);
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
davialexandre: This entire function is really difficult to read given it has so many levels of nested `foreach` loops and ifs. It can definitely be split into smaller functions. Please try to do that

powdevel: @davialexandre I move everything to a class and separate code with multiple functions.

civihr_employee_portal/civihr_employee_portal.module
powdevel: @davialexandre  Good point. Tables are now checked too with a similar mechanism as fields, so while tables keep their base name no matter if their number at the end differs, the import will be fine. I tested it manually changing table names on an exported view.

powdevel:
@davialexandre I made corrections following every comment. Code was moved to a separate class CustomFieldsMappingFixer and the main function was splitted.

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * To process a display option for a field in view
+   * @var array $option
+   *     structure that contains an option for a field in view
+   */
+  static private function processOption(&$option) {
+    // checking if this table is part of the civicrm views integration
+    $prefix = self::setCiviCRMtableAndGetPrefix($option['table']);
+    if ( $prefix ) {
+      $columns = self::getColumnsFromTable($prefix.$option['table']);
```
davialexandre: Missing spaces between operators

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+      // if the table where the current field belongs exists does not have it
+      // then we can search for similar fields in the same table and make replacements
+      if (!isset($columns[$option['field']])) {
+        $field_name_base = self::removeLastNumberOfString($option['field']);
```
davialexandre: Since now this is inside of a class, please use camelCase for variable names

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+   *     Display option a exported view structure
+   */
+  static private function processDisplayOptions(&$displayOption) {
+    foreach ($displayOption as &$option) {
+      // only continue if this option is about a field
+      if (!isset($option['field'])) {
+        continue;
+      }
+      $option_id_parts = explode('_', $option['field']);
+      $final_id_part = array_pop($option_id_parts);
```
davialexandre: Please use camelCase here too

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+class CustomFieldsMappingFixer {
+
+  /**
+   * To fix mapping on Views import of CiviCRM custom fields
+   *
+   * @param view $view
+   *     the view that is being imported / recreated
+   */
+  static public function fixMapping($view) {
+    // iterating displays in view
```
davialexandre: The code itself is clear enough here and this comment isn't adding any useful information. Please remove it

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+    foreach ($view->display as $display) {
+      // iterating options on display: fields, filters, arguments
```
davialexandre: This comment is not useful as well

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+   * Process all display options for an exported view only when it is related
+   * to a field and the field has a number at the end of its name, which is
+   * the pattern for custom fields from CiviCRM
+   *
+   * @var array
+   *     Display option a exported view structure
+   */
+  static private function processDisplayOptions(&$displayOption) {
+    foreach ($displayOption as &$option) {
+      // only continue if this option is about a field
```
davialexandre: This comment is also not useful

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+      if (!isset($option['field'])) {
+        continue;
+      }
```
davialexandre: Please add an empty line between logical blocks of code to improve readability. This applies to all the other methods on this class

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+      $option_id_parts = explode('_', $option['field']);
+      $final_id_part = array_pop($option_id_parts);
+      // it must contain underscores and a number at the end to be checked
```
davialexandre: Please also remove this comment

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+      if (count($option_id_parts) === 0 || !intval($final_id_part)) {
+        continue;
+      }
+      self::processOption($option);
+    }
+  }
+
+  /**
+   * To process a display option for a field in view
+   * @var array $option
```
davialexandre: Please add a empty line between the method description and `@var`

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+ */
+class CustomFieldsMappingFixer {
+
+  /**
+   * To fix mapping on Views import of CiviCRM custom fields
+   *
+   * @param view $view
+   *     the view that is being imported / recreated
+   */
+  static public function fixMapping($view) {
```
davialexandre: Please always add the visibility modifier first. I.e. `public static`. This applies to all the methods on this class

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,146 @@
...
+   * Return an existing CiviCRM table based on the base name of the table,
+   * which is the name of the table without the number at the end
+   *
+   * @param string $exportedTableWithId
+   *     this is the table name exported by views
+   *
+   * @return string
+   *     the name of the existing table if found or an empty string
+   */
+  private static function setCiviCRMtableAndGetPrefix(&$tableName) {
```
davialexandre: Almost every time we have a method with an "and" in its name, it means it does doing two things. A method should only do one thing (that is, have only one responsibility). Please check if we have split this. Perhaps one that sets the tables map and other that gets the prefix from this table

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,166 @@
...
+      // then we can search for similar fields in the same table and make replacements
+      if (!isset($columns[$option['field']])) {
+        $fieldNameBase = self::removeLastNumberOfString($option['field']);
+        $pos = array_search($fieldNameBase, $columns);
+        if ($pos !== FALSE) {
+          $option['field'] = array_search($columns[$pos], $columns);
+        }
+
+      }
+
```
davialexandre: These empty lines after each closing `}` are a bit weird and are not improving the readability. What I meant with adding empty lines between logical blocks of code is something more or less like this:

```php
// checking if this table is part of the civicrm views integration
$tableName = self::getCiviCRMtableName($option['table']);
$prefix = $tableName ? self::getPrefix($tableName) : false;

if ( $prefix ) {
  $option['table'] = $tableName;
  
  $columns = self::getColumnsFromTable($prefix . $tableName);
  
  // if the table where the current field belongs exists does not have it
  // then we can search for similar fields in the same table and make replacements
  if (!isset($columns[$option['field']])) {
    $fieldNameBase = self::removeLastNumberOfString($option['field']);
    
    $pos = array_search($fieldNameBase, $columns);
    if ($pos !== FALSE) {
      $option['field'] = array_search($columns[$pos], $columns);
    }
  }
}
```

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,166 @@
...
+    $pieces = explode('_', $string);
+    $lastIndex = count($pieces) - 1;
+    return $lastIndex > 0 && intval(array_pop($pieces)) ? implode('_', $pieces) : $string;
+  }
+
+  /**
+   * Return an existing CiviCRM table based on the base name of the table,
+   * which is the name of the table without the number at the end
+   *
+   * @param string $exportedTableWithId
```
davialexandre: the param name here doesn't match the method param name

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,166 @@
...
+   * @return string
+   *     the name of the existing table if found or an empty string
+   */
+  private static function getCiviCRMtableName($tableName) {
+    global $databases;
+    static $tableNamesMap = [];
+    // creates a map with table names without IDs to existing table names with IDs
+    if (!$tableNamesMap) {
+      $tablesWithIds = array_keys($databases['default']['default']['prefix']);
+      foreach ($tablesWithIds as &$tableWithId) {
```
davialexandre: Does `$tableWithId` really need to be a reference?

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,166 @@
...
+        $tableNamesMap[self::removeLastNumberOfString($tableWithId)] = $tableWithId;
+      }
+    }
+
+    $exportedTableBaseName = self::removeLastNumberOfString($tableName);
+    $tableFound = isset($tableNamesMap[$exportedTableBaseName]);
+    $tableName = $tableNamesMap[$exportedTableBaseName];
+    return !$tableFound ? '' : $tableName;
```
davialexandre: Do we really need the `!`? I think we can just `return $tableFound ? $tableName : ''`

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,166 @@
...
+    if (!$tableNamesMap) {
+      $tablesWithIds = array_keys($databases['default']['default']['prefix']);
+      foreach ($tablesWithIds as &$tableWithId) {
+        $tableNamesMap[self::removeLastNumberOfString($tableWithId)] = $tableWithId;
+      }
+    }
+
+    $exportedTableBaseName = self::removeLastNumberOfString($tableName);
+    $tableFound = isset($tableNamesMap[$exportedTableBaseName]);
+    $tableName = $tableNamesMap[$exportedTableBaseName];
```
davialexandre: If the table is not found, this line will throw a notice. I think it is better to change this to something like:

```php
if (isset($tableNamesMap[$exportedTableBaseName])) {
  return $tableNamesMap[$exportedTableBaseName];
}

return '';
```

powdevel:
@davialexandre No problem Davi, thanks as always for your feedback.

PR #527, opened 2018-07-26
PCHR-3941: Detach require.min.js from reqangular

deb1990:
## Overview
In https://github.com/compucorp/civihr/pull/2775, the `require.min.js` has been detached from `reqangular.min.js`. In this PR the 'require.min.js' is attached to the DOM

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -6691,13 +6691,9 @@ function _civihr_employee_portal_create_default_reports_configuration() {
 function civihr_employee_portal_get_markup_for_extension($resource = '', $scriptExtensionKey, $markupModuleName) {
   $script = CRM_Core_Config::singleton()->debug ? "{$resource}.js" : "{$resource}.min.js";
 
-  // Adds RequireJS library only once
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6689,7 +6691,12 @@ function _civihr_employee_portal_create_default_reports_configuration() {
 function civihr_employee_portal_get_markup_for_extension($resource = '', $scriptExtensionKey, $markupModuleName) {
   $script = CRM_Core_Config::singleton()->debug ? "{$resource}.js" : "{$resource}.min.js";
 
-  // Adds RequireJS + AngularJS dependencies only once
+  // Adds RequireJS library only once
```
AkA84: i think you can write it like this, so it's more compact
```js
// Adds RequireJS + AngularJS dependencies only once
if (!strpos(drupal_get_js('footer'), 'require.min.js')) {
  civicrm_resources_load('org.civicrm.reqangular', ['require.min.js']);
  civicrm_resources_load('org.civicrm.reqangular', ['reqangular.min.js']);
}
```
i think it's safe to assume that if `require.min.js` is not present, neither it's reqangular?

PR #528, opened 2018-07-27
PCHR-4034: Settings to Send Notifications on Contact Details Changes

powdevel:
## Overview
Add settings to let HR Admin to configure notifications when a contact changes its details.
Settings added to the Onboarding Wizard config form.

## Before
peek 2018-07-26 23-43

## After
peek 2018-07-26 23-41


powdevel:
@mickadoo Corrections made as per your comments. The snippet generation was simplified, but not taken to another file, which is exactly what I tried not to do, however if you think it is still important then tell me to fix that right away.

PR #530, opened 2018-08-03
PCHR-3928: Notifications for Onboarding Wizard

powdevel:
## Overview
Sends notifications about user data changes when a user submits the Onboarding Wizard. Changes on data are detected comparing the view results before submission is saved and after submission is inserted. If changes are detected an email is sent to the admin if the admin has previously chosen to be notified (setting inside Onboarding Wizard configuration form).

## New Classes and Updated Classes

2 Classes were created:
- `Drupal\civihr_employee_portal\View\ViewsResultDiff` obtains differences between view results.
- `Drupal\civihr_employee_portal\Mail\ContactDetailsChangesMail` sends notifications to the admin with the changes of user data.

`Drupal\civihr_employee_portal\Mail\AbstractDrupalSystemMail` class was refactored and classes that extend this class, `AccountActivatedSystemMail` and `ContactDetailsChangesMail` were updated accordingly.

## Additions
- **7 Views** were created on `./views/views_export/notifications` to obtain data about changes due to Onboarding submissions. These views also define the HTML output on notification emails, more precisely labels and values of the changed fields and the title of the section as the view title.

- **6 templates** inside `./templates/smarty/notifications` were created to wrap the results of the differences and build an entire notification email.

- **3 functions** were created on `Drupal\civihr_employee_portal\Helpers\HelperClass`, 2 of them are related to avoid using IDs for values like _Location Types_ or _Phone Types_, because these IDs could vary between installations.

## Implemented Hooks
- `civihr_employee_portal_mail` _hook_mail_
- `civihr_employee_portal_preprocess_views_view_fields` _hook_preprocess_views_view_fields_
- `civihr_employee_portal_webform_submission_presave` _hook_submission_presave_

## Before

No notifications were sent on data changes related to contact due to Onboarding submissions.

## After
The following are examples notifications taken from Chrome PC in a Gmail account:

image

image

### DIFF
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -6803,3 +6803,16 @@ function civihr_employee_portal_send_email_of_changes($informationVia, $changes)
     drupal_mail('civihr_employee_portal', $params['key'], $mailto, language_default(), $params);
   }
 }
+
+/**
+ * Implements theme_preprocess_views_view_fields
+ *
+ * @param  array $vars
+ *     The variables on the template
+ */
+function civihr_employee_portal_preprocess_views_view_fields(&$vars) {
+  if (!isset($vars['view']->fieldsMarkup)) {
+    $vars['view']->fieldsMarkup = [];
+  }
+  $vars['view']->fieldsMarkup[] = $vars['fields'];
+}
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -5108,6 +5108,70 @@ function _civihr_employee_portal_should_do_onboarding($account) {
 }
 
 /**
+ * Implements hook_webform_submission_presave()
+ *
+ * @param stdClass $node
+ * @param stdClass $submission
+ */
+function civihr_employee_portal_webform_submission_presave($node, &$submission) {
+
+  if ($node->title === OnboardingWebForm::NAME) {
+    // check if updates are active before storing data for diffs
+    if (variable_get('civihr_onboarding_send_updates', FALSE)) {
+
+      $locationTypeIdPersonal = HelperClass::getLocationTypeIdByName('Personal');
+      $locationTypeIdWork = HelperClass::getLocationTypeIdByName('Work');
+      $PhoneTypeIdPhone = HelperClass::getPhoneTypeByName('Phone');
+
+      $dependantsContacts = views_get_view_result('mydetails_notifications__emergency_contacts', 'emergency_contact_ids', 'Yes', '_');
+      $newestDependantContactId = (int) HelperClass::getFirstValueFromFirstObjectInArray($dependantsContacts);
+
+      $emergencyContacts = views_get_view_result('mydetails_notifications__emergency_contacts', 'emergency_contact_ids', 'No', '_');
+      $newestEmergencyContactId = (int) HelperClass::getFirstValueFromFirstObjectInArray($emergencyContacts);
+
+      // diff configuration for Onboarding form
+      $viewsResultDiff = new ViewsResultDiff([
+        [ 'view' => 'mydetails_notifications__my_personal_details'],
+        [ 'view' => 'mydetails_notifications__my_address',
+          'args' => [$locationTypeIdPersonal]],
+        [ 'group' => [
+            [ 'view' => 'mydetails_notifications__my_contact_info_other'],
+            [ 'view' => 'mydetails_notifications__my_contact_info_emails',
+              'args' => [$locationTypeIdPersonal, 'Personal']
+            ],
+            [ 'view' => 'mydetails_notifications__my_contact_info_emails',
+              'args' => [$locationTypeIdWork, 'Work']
+            ],
+            [ 'view' => 'mydetails_notifications__my_contact_info_phones',
+              'exposedFilters' => ['location_type'=> $locationTypeIdPersonal, 'phone_type' => $PhoneTypeIdPhone],
+            ],
+            [ 'view' => 'mydetails_notifications__my_contact_info_phones',
+              'exposedFilters' => ['location_type'=> $locationTypeIdWork, 'phone_type' => $PhoneTypeIdPhone],
+            ],
+          ],
+        ],
+        [ 'view' => 'mydetails_notifications__my_payroll'],
+        [ 'view' => 'mydetails_notifications__emergency_contacts',
+          'args' => ['No', 'Emergency Contacts'],
+          'fieldAsKeyForRows' => 'id',
+          'exposedFilters' => ['id' => $newestDependantContactId],
+        ],
+        [ 'view' => 'mydetails_notifications__emergency_contacts',
+          'args' => ['Yes', 'Dependants'],
+          'fieldAsKeyForRows' => 'id',
+          'exposedFilters' => ['id' => $newestEmergencyContactId],
+        ],
+      ]);
+
+      // Saving data just before saving the new results of the form and
+      // storing ViewsResultDiff instance on submission object to access it
+      // once the submission gets saved
+      $submission->ViewsResultDiff = $viewsResultDiff->storeOldData();
+    }
+  }
+}
+
+/**
  * Implements hook_webform_submission_insert().
  * Some post processing needs to be done on the form, such as cache clearing
  *
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -5182,6 +5182,13 @@ function civihr_employee_portal_webform_submission_insert($node, $submission) {
   if ($node->title === OnboardingWebForm::NAME) {
     $form = new OnboardingWebForm();
     $form->onSubmit($node, $submission);
+    // check if updates are active before comparing data or send updates
+    if (variable_get('civihr_onboarding_send_updates', FALSE)) {
+      // comparing new data and getting changes
+      $changes = $submission->ViewsResultDiff->storeNewData()->compareData()->changes;
+      // sending notifications
+      civihr_employee_portal_send_email_of_changes('The Onboarding Wizard', $changes);
+    }
   }
 }
 
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -6780,3 +6780,26 @@ function civihr_employee_portal_page_build(&$page) {
       ctools_modal_add_js();
   }
 }
+
+/**
+ * Send notification emails about changes to the email entered in
+ * the Onboarding Form Configuration
+ *
+ * @var string $informationVia
+ *    A string which indicates where the information is coming from. Ex: Onboarding Form
+ * @var array $changes
+ *    Keyed array with sections and field names and field values
+ */
+function civihr_employee_portal_send_email_of_changes($informationVia, $changes) {
+  $mailto = variable_get('civihr_onboarding_email_to_send_updates');
+  // if there are changes and an email to send the updates
+  if ($mailto && $changes) {
+    $params = [];
+    $params['informationVia'] = $informationVia;
+    $params['changes'] = $changes;
+    $params['subject'] = 'CiviHR Self Service Data Submission';
+    $params['contactData'] = views_get_view_result('notifications_contact_data')[0];
+    $params['key'] = 'details_update_mail_notification';
+    drupal_mail('civihr_employee_portal', $params['key'], $mailto, language_default(), $params);
+  }
+}
--- a/civihr_employee_portal/civihr_employee_portal.module
+++ b/civihr_employee_portal/civihr_employee_portal.module
@@ -17,6 +17,7 @@ use Drupal\civihr_employee_portal\Helpers\LinkProvider;
 use Drupal\civihr_employee_portal\Helpers\TaxonomyHelper;
 use Drupal\civihr_employee_portal\View\ViewsResultDiff;
 use Drupal\civihr_employee_portal\Mail\ContactDetailsChangesMail;
+use Drupal\civihr_employee_portal\View\CustomFieldsMappingFixer;
 
 /**
  * Implements hook_install().
@@ -833,6 +834,7 @@ function civihr_employee_portal_views_default_views() {
 
     if (isset($view)) {
       $views[$view->name] = $view;
+      CustomFieldsMappingFixer::fixMapping($view);
     }
   }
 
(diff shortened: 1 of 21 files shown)


mickadoo:
@powdevel This PR is targeting `PCHR-3747-Notifications-for-Onboarding-Wizard` but the [PCHR-3928 ticket](https://compucorp.atlassian.net/browse/PCHR-3928) is a technical task of parent [PCHR-3748](https://compucorp.atlassian.net/browse/PCHR-3748). Can you explain the structure here?

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,164 @@
...
+   *     the string which could have underscores or not
+   *
+   * @return string
+   *     the string without the number at the end,
+   *     if no number found at the end returns the same string
+   */
+  private static function removeLastNumberOfString($string) {
+    $pieces = explode('_', $string);
+    $lastIndex = count($pieces) - 1;
+    return $lastIndex > 0 && intval(array_pop($pieces)) ? implode('_', $pieces) : $string;
```
mickadoo: space before `return`, also this long line of logic would be a lot more readable with separate variables like

```
$pieces = explode('_', $string);
$lastIndex = count($pieces) - 1;
$hasEnoughParts === ($lastIndex > 0);
if (!$hasEnoughParts) {
  return $string;
}
$lastPart = array_pop($pieces);
$lastPartIsInt === is_int($lastPart);

return  $lastPartIsInt ? implode('_', $pieces) : $string;
```

Although I think if I was writing this I'd probably avoid expensive `explode` and `implode` calls and just use some function that used `strrpos` and `substr`

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,164 @@
...
+   * where "base name of the table" is the name of the table without the number
+   * at the end
+   *
+   * @param string $tableName
+   *     this is the table name exported by views
+   *
+   * @return string
+   *     the name of the existing table if found or an empty string
+   */
+  private static function getCiviCRMtableName($tableName) {
```
mickadoo: I don't know what this function does, it takes and argument `$tableName` and returns `CiviCRMtableName`. Have you checked if there's an easier way to get the table names than using the global `$databases` variable, have you looked at `CRM_Core_DAO_AllCoreTables`?

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,164 @@
...
+    global $databases;
+    static $tableNamesMap = [];
```
mickadoo: I don't think we should use static function variables, this seems like a good situation to use just a regular class property and fetch it just once. Does this class need to be static or can we make it non-static?

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,164 @@
...
+    if (!$tableNamesMap) {
+      $tablesWithIds = array_keys($databases['default']['default']['prefix']);
+      foreach ($tablesWithIds as $tableWithId) {
+        $tableNamesMap[self::removeLastNumberOfString($tableWithId)] = $tableWithId;
+      }
+    }
+
+    $exportedTableBaseName = self::removeLastNumberOfString($tableName);
+    $tableFound = isset($tableNamesMap[$exportedTableBaseName]);
+    return $tableFound ? $tableNamesMap[$exportedTableBaseName] : '';
```
mickadoo: space before return

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  /**
+   * To obtain the prefix for a table in the current installation
+   *
+   * @param string $tableName
+   *     the name of the table
```
mickadoo: 2 spaces before description

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @return string
+   *     the prefix for the table
+   */
+  private static function getPrefix($tableName) {
+    global $databases;
+    return $databases['default']['default']['prefix'][$tableName];
```
mickadoo: Please avoid using global variables if you can, it's better to use a class that allows access to this in case something changes with the underlying implementation later.

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  /**
+   * Gets the columns from a table keyed by the name of the column without
+   * the sequential at the end
+   *
+   * @param string $table
+   *     the name of the table
```
mickadoo: spaces before description

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @return array
+   *     the names of the columns of the table
+   */
+  private static function getColumnsFromTable($table) {
+    static $columns = [];
+    if (!isset( $columns[$table])) {
```
mickadoo: extra space before `$columns`

civihr_employee_portal/src/View/ViewsResultDiff.php
```diff
@@ -0,0 +1,284 @@
...
+   *
+   * In the array where "group" key is defined:
+   * - The value will be an array of "views" settings, as defined above
+   *
+   * @var array
+   */
+  public $viewsSettings = [];
+
+  /**
+   * usually will hold data after a form submission but previous to save or insert
```
mickadoo: This has the same description as `$oldData` - is this correct?

civihr_employee_portal/src/View/ViewsResultDiff.php
```diff
@@ -0,0 +1,284 @@
...
+   * @param  array $oldData
+   *     previous result from view
+   * @param  array $newData
+   *     a recent result from view
+   * @param  array $changes
+   *     the difference between $oldData and $newData
+   *
+   * @return stdClass $this
+   */
+  private function compareDataRecursive(&$oldData, &$newData, &$changes) {
```
mickadoo: Why are $oldData and $newData passed by reference if their values are just checked in this function?

civihr_employee_portal/src/View/ViewsResultDiff.php
```diff
@@ -0,0 +1,284 @@
...
+    return $this;
+  }
+
+  /**
+   * To retrieve the current data of the views configured to diff
+   * this function will be called by storeNewData or storeOldData
+   * to be able to store the data produced by the views for later
+   * comparison
+   *
+   * @return array
```
mickadoo: missing `$param`

civihr_employee_portal/src/View/ViewsResultDiff.php
```diff
@@ -0,0 +1,284 @@
...
+    $data = [];
+    $settings = !$settings ? $this->viewsSettings : $settings;
+
+    foreach ($settings as $setting) {
+      // if it is a group then apply recursion
+      if (isset($setting['group'])) {
+        $data[] = ['group' => $this->getCurrentData($setting['group'])];
+      } else if (isset($setting['view'])) {
+        // if it is a view it must be previewed to get the results
+        $view = views_get_view($setting['view']);
```
mickadoo: Is there a way we can create separate classes for comparing and fetching the data? Ideally the diff generator would just receive 2 sets of data and generate the diff without any drupal function calls.

civihr_employee_portal/src/View/ViewsResultDiff.php
```diff
@@ -0,0 +1,284 @@
...
+
+        // Setting Exposed Filters
+        // backing up $_GET
+        $_GETbkp = $_GET;
```
mickadoo: This seems really messy, is there no other way to get view data without doing this?

civihr_employee_portal/src/View/ViewsResultDiff.php
```diff
@@ -0,0 +1,284 @@
...
+        // add exposed filter keys to $_GET
+        $_GET = isset($setting['exposedFilters']) ? $setting['exposedFilters'] + $_GET : $_GET ;
+        $view->preview($setting['displayId'], $setting['args']);
+        // reverting $_GET to backed state once the view is executed
+        $_GET = $_GETbkp;
+
+        // views which are empty will not have fieldsMarkup
+        $view->fieldsMarkup = isset($view->fieldsMarkup) ? $view->fieldsMarkup : [];
+        $currentViewData = [];
+        $fieldKeyForRows = isset($setting['fieldAsKeyForRows']) ? $setting['fieldAsKeyForRows'] : false;
```
mickadoo: boolean case

civihr_employee_portal/src/View/ViewsResultDiff.php
```diff
@@ -0,0 +1,284 @@
...
+    return $data;
+  }
+
+  /**
+   * Stores data on property newData
+   *
+   * @return ViewsResultDiff
+   */
+  public function storeNewData() {
+    $this->newData = $this->getCurrentData();
```
mickadoo: the whole distinction between `storeNewData` and `storeOldData` is confusing, this whole class is doing a lot of stuff, I'd suggest breaking it down into separate services like
```
ViewsDiffGenerator::getDiff($firstDataSet, $secondDataSet);
ViewsDataFetcher::getCurrentData($all, $params, $that, $are, $required);
```

Also the whole view setting structure is so complicated that it doesn't make any sense when I'm looking at it in the constructor. Maybe you could consider creating an object for this, in general this class is just hard to follow - it would be better to break it into more specific classes. I think diff generation is an ideal candidate for unit testing but I don't see any tests in this PR at all. With smaller, more specific classes we can write clear tests for them that don't require Drupal function calls.

civihr_employee_portal/templates/smarty/notifications/html.tpl
```diff
@@ -0,0 +1,151 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
```
mickadoo: Where did you get the code for this?

civihr_employee_portal/views/views_export/notifications/mydetails_notifications__emergency_contacts.inc
```diff
@@ -0,0 +1,202 @@
+<?php
+
+$view = new view();
```
mickadoo: I wasn't clear just from reading the PR description, why are we creating all these new views? Is it just to generate the email content? If that's the case then why not just use smarty templates if we already have all the data required to generate them?

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -5136,13 +5181,16 @@ function _civihr_employee_portal_should_do_onboarding($account) {
  * @param \stdClass $submission
  */
 function civihr_employee_portal_webform_submission_insert($node, $submission) {
-  $nodeTitle = property_exists($node, 'title') ? $node->title : NULL;
-  $onboardingTitle = OnboardingWebForm::NAME;
-  $isOnboardingForm = $nodeTitle === $onboardingTitle;
-
-  if ($isOnboardingForm) {
+  if ($node->title === OnboardingWebForm::NAME) {
```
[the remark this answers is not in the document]
powdevel: @mickadoo in my experience, and please tell me if I'm wrong, I haven't seen a node without a title property. I think this check it not needed.

civihr_employee_portal/src/View/ViewsResultDiff.php
powdevel: @mickadoo not currently. I tried using arguments for this but wasn't enough to get the results needed, probably because wrong integration of CiviCRM and views. So I have to use exposed filters, and as far as I know, this is the only way of achieving it, and for sure the simplest.

civihr_employee_portal/views/views_export/notifications/mydetails_notifications__emergency_contacts.inc
powdevel: @mickadoo new views serves for obtaining results before and after, with that information differences are calculated. Those differences are sent in the notification email. Hence new views are created with 2 purposes: for generating differences and generating the notifications.

civihr_employee_portal/templates/smarty/notifications/html.tpl
powdevel: @mickadoo from another smarty template that is used to send notifications.

civihr_employee_portal/src/View/ViewsResultDiff.php
powdevel: `ViewsDiffResultDiff` is a class that generates differences between views. I will create a class to fetch the data, and inject the results in `ViewsDiffResultDiff` to avoid external dependencies.

civihr_employee_portal/src/View/ViewsResultDiff.php
powdevel: @mickadoo Sure. It is doable.

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
powdevel: @mickadoo I will search for code that returns this information and will use it in case I found it.

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
powdevel: @mickadoo this class is static because it does not need to store different properties in multiple instances (don't access instance data). The code for this class was approved here https://github.com/compucorp/civihr-employee-portal/pull/526 but I have to include it because it was causing conflicts with the base branch.

civihr_employee_portal/src/View/CustomFieldsMappingFixer.php
```diff
@@ -0,0 +1,164 @@
...
+   * In this context that number is always separated by an underscore
+   *
+   * @param  string $string
+   *     the string which could have underscores or not
+   *
+   * @return string
+   *     the string without the number at the end,
+   *     if no number found at the end returns the same string
+   */
+  private static function removeLastNumberOfString($string) {
```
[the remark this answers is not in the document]
powdevel: @mickadoo then the name should be "removeLastNumberOfStringIfItIsAfterAnUnderscore" ?

mickadoo:
Closing for now as unclear if this will be used

PR #531, opened 2018-08-03
PCHR-4004: Removing HR Vacancies From SSP

backstageel:
## Overview
This PR removes all rereferences to the HR-Vacancies extensions, since it is being removed.

## Before
There was a "Vacancies" menu in SSP

## After
The "Vacancies" menu will be removed after releasing this PR and running the drush command to disable the view


civihr_employee_portal/civihr_employee_portal.install
```diff
@@ -454,6 +454,15 @@ function civihr_employee_portal_update_7035() {
...
 
+/**
+ * Disables the HR Vacancies View
+ */
+function civihr_employee_portal_update_7036() {
+  $status = variable_get('views_defaults', array());
+  $status['hr-vacancies'] = TRUE;
+  variable_set('views_defaults', $status);
+}
+
```
tunbola: I guess we probably need to remove this also: https://github.com/compucorp/civihr-employee-portal/blob/65650c5f4adc715ffc49336204131ad1136c7c68/civihr_employee_portal/features/civihr_employee_portal_features/civihr_employee_portal_features.info#L61

PR #533, opened 2018-08-07
PCHR-4073: Fix SSP Filter on Staff Directory

ajesamson:
## Overview
Filtering staff records in SSP have some issue with job roles that are not active. This PR fixes the issue by applying additional filter condition to job role start date. Date filtering is now applied to filter requests containing department and job title.

## Before
Staff with job roles starting in the future or past end dates show up when running filter.

## After
Staff with job roles starting in the future or past end dates no longer show up in filter result.

## Technical Details
In [Staff Details View](https://github.com/compucorp/civihr-employee-portal/blob/staging/civihr_employee_portal/src/View/StaffDirectoryView.php#L15) the query was altered to cater for `job title` and `role start date` field in addition to `department` and `role end date` fields.
```
$titleField = 'hrjc_role_hrjc_revision.role_title';
$roleStartField = 'hrjc_role_hrjc_revision.role_start_date';
```

Additional query grouping was added.
```
$andGroup = $query->set_where_group();
$query->add_where($andGroup, $roleStartField, date('Y-m-d'), '<=');
```

### DIFF
--- a/civihr_employee_portal/src/View/StaffDirectoryView.php
+++ b/civihr_employee_portal/src/View/StaffDirectoryView.php
@@ -16,13 +16,17 @@ class StaffDirectoryView extends AbstractView {
     $queryParts = [];
     $this->getWhereFields($query->where, $queryParts);
     $departmentField = 'hrjc_role_hrjc_revision.role_department';
+    $titleField = 'hrjc_role_hrjc_revision.role_title';
     $roleEndField = 'hrjc_role_hrjc_revision.role_end_date';
+    $roleStartField = 'hrjc_role_hrjc_revision.role_start_date';
 
-    // if department is set we must limit to only current roles
-    if (!empty($queryParts[$departmentField])) {
-      $group = $query->set_where_group('OR'); // add a new WHERE group
-      $query->add_where($group, $roleEndField, date('Y-m-d'), '>=');
-      $query->add_where($group, $roleEndField, NULL, 'IS');
+    // if department or title is set we must limit to only current roles
+    if (!empty($queryParts[$departmentField]) || !empty($queryParts[$titleField])) {
+      $roleEndGroup = $query->set_where_group('OR'); // add a new WHERE group
+      $query->add_where($roleEndGroup, $roleEndField, date('Y-m-d'), '>=');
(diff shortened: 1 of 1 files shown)


civihr_employee_portal/src/View/StaffDirectoryView.php
```diff
@@ -16,13 +16,19 @@ public function alter($view, $query) {
...
     $roleEndField = 'hrjc_role_hrjc_revision.role_end_date';
+    $roleStartField = 'hrjc_role_hrjc_revision.role_start_date';
 
-    // if department is set we must limit to only current roles
-    if (!empty($queryParts[$departmentField])) {
-      $group = $query->set_where_group('OR'); // add a new WHERE group
-      $query->add_where($group, $roleEndField, date('Y-m-d'), '>=');
-      $query->add_where($group, $roleEndField, NULL, 'IS');
+    // if department or title is set we must limit to only current roles
+    if (!empty($queryParts[$departmentField]) ||
```
mickadoo: Doesn't this fit in the 80 character limit on one line?

civihr_employee_portal/src/View/StaffDirectoryView.php
```diff
@@ -16,13 +16,19 @@ public function alter($view, $query) {
...
+      !empty($queryParts[$titleField])
+    ) {
+      $orGroup = $query->set_where_group('OR'); // add a new WHERE group
```
mickadoo: I think this is `$orGroup` because it's `endDate is in future OR is null`, but you have `$andGroup` that doesn't contain AND conditions. Do you think it would be clearer to rename these `roleEndGroup` and `roleStartGroup` or `roleEndCondition` or something like that?

ajesamson: The `add_where` function requires a group which will determine what section the query will be added (AND/OR).  It makes sense to rename as suggested 👌

civihr_employee_portal/src/View/StaffDirectoryView.php
ajesamson: Will check and update as appropriate 👓

PR #535, opened 2018-08-09
PCHR-4104: JS Error on SSP - Dashboard and Tasks

ajesamson:
## Overview
This PR fixes JS error in SSP arising as a result order in which required libraries as loaded.

## Before
Despite specifying the order in which required files are to be loaded, required priorities are sometimes mixed up.

## After
Required order of resources are strictly followed when loading files.

## Technical Details
Instead of iterating through available resources and search for the required file uri, the required files were first looped through. This ensures the order of required files are maintained when loading their URI.
```
foreach ($req_files as $req_file) {
  $index = array_search($req_file, $fileNames);
  if ($index && preg_match('/\.js$/', $req_file)) {
    $js_list[] = $resourceFilenameUriValues[$index]['uri'];
  }
  elseif ($index && preg_match('/\.css$/', $req_file)) {
    $css_list[] = $resourceFilenameUriValues[$index]['uri'];
  }
}
```

Existing implementations were modified to prevent repetition of task.

### DIFF
--- a/civicrm_resources/civicrm_resources.module
+++ b/civicrm_resources/civicrm_resources.module
@@ -23,23 +23,29 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
   $js_list = [];
   $css_list = [];
 
+  $resourceFilenameUri = array_map(function ($file) {
+    return ['filename' => $file->filename, 'uri' => $file->uri];
+  }, $files);
+  $resourceFilenameUriValues = array_values($resourceFilenameUri);
+  $fileNames = array_column($resourceFilenameUri, 'filename');
+  foreach ($req_files as $req_file) {
+    $index = array_search($req_file, $fileNames);
+    if ($index && preg_match('/\.js$/', $req_file)) {
+      $js_list[] = $resourceFilenameUriValues[$index]['uri'];
+    }
+    elseif ($index && preg_match('/\.css$/', $req_file)) {
+      $css_list[] = $resourceFilenameUriValues[$index]['uri'];
+    }
+  }
+
   foreach ($files as $key => $file) {
     if (!empty($req_files)) {
       if (in_array('*.js', $req_files) && preg_match('/\.js$/', $file->filename)) {
         $js_list[] = $file->uri;
       }
-      else {
-        if (in_array($file->filename, $req_files) && preg_match('/\.js$/', $file->filename)) {
-          $js_list[] = $file->uri;
-        }
-      }
+
       if (in_array('*.css', $req_files) && preg_match('/\.css$/', $file->filename)) {
         $css_list[] = $file->uri;
-      }
-      else {
-        if (in_array($file->filename, $req_files) && preg_match('/\.css$/', $file->filename)) {
-          $css_list[] = $file->uri;
-        }
       }
     }
     else {


civicrm_resources/civicrm_resources.module
```diff
@@ -23,24 +23,30 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
   $js_list = [];
   $css_list = [];
 
+  $resourceFilenameUri = array_map(function ($file) {
+    return ['filename' => $file->filename, 'uri' => $file->uri];
+  }, $files);
+  $resourceFilenameUriValues = array_values($resourceFilenameUri);
+  $fileNames = array_column($resourceFilenameUri, 'filename');
```
davialexandre: We'll have a problem here if have two files with the same name, but in different folders

civicrm_resources/civicrm_resources.module
```diff
@@ -23,24 +23,30 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
...
+  foreach ($req_files as $req_file) {
+    $index = array_search($req_file, $fileNames);
+    if ($index && preg_match('/\.js$/', $req_file)) {
```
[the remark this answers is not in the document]
ajesamson: Yes, that is correct. It means the required file is not found in the loaded resources.

civicrm_resources/civicrm_resources.module
```diff
@@ -23,24 +23,30 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
...
+      $js_list[] = $resourceFilenameUriValues[$index]['uri'];
+    }
+    elseif ($index && preg_match('/\.css$/', $req_file)) {
```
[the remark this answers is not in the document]
ajesamson: It only meant that the requested file cannot be found in the loaded resources.

civicrm_resources/civicrm_resources.module
ajesamson: Yes that is true but since `array_search` is used to [find index](https://github.com/compucorp/civihr-employee-portal/pull/535/files#diff-ec67a98d0a58dd334cfd09811516a417R32) of the file, it will always return the first corresponding key if successful. So the second will always be ignored.

ajesamson:
@davialexandre The logic is correct. I will look into implementation.

civicrm_resources/civicrm_resources.module
```diff
@@ -59,6 +61,46 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
   }
 }
 
+/**
+ * Filters files into css or js
+ *
+ * @param array $files
+ * @return array
+ */
+function filter_files($files) {
```
davialexandre: Functions are shared in the global namespace and a generic name like this can easily cause conflicts. Please prefix them with the module name: `civicrm_resources_filter_files`. Also, this seems like a function that is not supposed to be used by other modules. In this case, I recommend you also prefix it with an underscore to indicate it is "private"

civicrm_resources/civicrm_resources.module
```diff
@@ -59,6 +61,46 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
...
+  $filteredJsFile = array_filter($files, function ($file) {
+    return preg_match('/\.js$/', $file->filename);
+  });
+  $filteredCssFile = array_filter($files, function ($file) {
+    return preg_match('/\.css$/', $file->filename);
+  });
+  return [$filteredJsFile, $filteredCssFile];
```
davialexandre: Please add some spaces here between the function calls to improve the readbility

civicrm_resources/civicrm_resources.module
```diff
@@ -59,6 +61,46 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
   }
 }
 
+/**
+ * Filters files into css or js
+ *
+ * @param array $files
+ * @return array
```
davialexandre: Please add one empty line between `@param` and `@return` to conform to the drupal/civi code standards

civicrm_resources/civicrm_resources.module
```diff
@@ -59,6 +61,46 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
...
+  return [$filteredJsFile, $filteredCssFile];
+}
+
+/**
+ * Retrieves uri for file list
+ *
+ * @param array $files
+ * @return array
+ */
+function get_uris($files) {
```
davialexandre: Please also add prefixes to this function

civicrm_resources/civicrm_resources.module
```diff
@@ -59,6 +61,46 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
...
+    return preg_match('/\.css$/', $file->filename);
+  });
+  return [$filteredJsFile, $filteredCssFile];
+}
+
+/**
+ * Retrieves uri for file list
+ *
+ * @param array $files
+ * @return array
```
davialexandre: Please also add an empty line here

civicrm_resources/civicrm_resources.module
```diff
@@ -59,6 +61,46 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
...
+ */
+function get_uris($files) {
+  $resourceUri = array_map(function ($file) {
+    return  $file->uri;
+  }, $files);
+  return array_values($resourceUri);
```
davialexandre: Please add an empty line above the `return` statement, to improve readability

civicrm_resources/civicrm_resources.module
```diff
@@ -59,6 +61,46 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
...
+}
+
+/**
+ * Retrieves uri for a specific file
+ *
+ * @param int $index
+ * @param array $resourceList
+ * @return string
```
davialexandre: Please add empty line here as well

civicrm_resources/civicrm_resources.module
```diff
@@ -59,6 +61,46 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
...
+ */
+function get_file_uri($index, $resourceList) {
```
davialexandre: Please add prefixes here too

civicrm_resources/civicrm_resources.module
```diff
@@ -19,38 +19,40 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
...
+  if (in_array('*.js', $req_files)) {
+    $js_list = get_uris($all_js);
+  }
+
+  if (in_array('*.css', $req_files)) {
+    $css_list = get_uris($all_css);
+  }
+
+  if (!in_array('*.css', $req_files) || !in_array('*.js', $req_files)) {
+    $resourceFilenameUri = array_map(function ($file) {
```
davialexandre: The other variables in this function are following the `snake_case` standard, so please make sure of following that for all the variable you've add

davialexandre: I wonder if it wouldn't simply be easier to change `civicrm_resources_file_scan_directory` to return an array of arrays instead of objects. This way, it wouldn't be necessary to use `array_map` here to create this array

civicrm_resources/civicrm_resources.module
```diff
@@ -19,38 +19,40 @@ function civicrm_resources_add_resources($extention_path, $req_files = []) {
...
+      return ['filename' => $file->filename, 'uri' => $file->uri];
+    }, $files);
+    $fileNames = array_column($resourceFilenameUri, 'filename');
```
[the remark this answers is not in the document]
ajesamson: Using `array_search` will returns the first matched key by default. So others will be ignored.

civicrm_resources/civicrm_resources.module
ajesamson: Using array will be much better since the `array_column` in the following requires an array, making the `array_map` unnecessary.

PR #539, opened 2018-08-20
PCHR-4121: Fix SSP resources inclusion

igorpavlov:
## Overview

This PR fixes an issue which breaks all SSP pages if node modules are installed in CiviHR repository. This issue appeared as a result of https://github.com/compucorp/civihr-employee-portal/pull/535. This PR also increases the performance of initial SSP pages load.

## Before

Pages do not load normally and resource load JS errors are thrown in the console.

image

Performance (T&A page as an example):
image

## After

Pages load normally, without resource load JS errors in the console:

image

Performance (T&A page as an example) - ~40% faster:
image

## Technical Details

The issue appeared because of the implementation of the concept of finding full paths to resource files by specifying only extension and resource file names. It mixed JS and CSS files in one collection, then performed an search for an index in the JS only collection and then applied that index in the mixed collection. Because these collections have different sizes due to different content, some JS files were not in sync in the collections and thus wrong JS resources were injected.

```php
function civicrm_resources_add_resources($extention_path, $req_files = []) {
  global $base_url;

  $files = civicrm_resources_file_scan_directory($extention_path, '/\.(css|js)$/i'); // mix CSS and JS in one list
  
  // ...
  if (!in_array('*.css', $req_files) || !in_array('*.js', $req_files)) {
    $file_names = array_column($files, 'filename');

    foreach ($req_files as $req_file) {
      $index = array_search($req_file, $file_names); // index is gotten from the mixed resources list
      if (($index !== FALSE) && preg_match('/\.js$/', $req_file)) {
        $js_list[] = _civicrm_resources_get_file_uri($index, $all_js); // and applied to a specific type resources list
        continue;

      if (($index !== FALSE) && preg_match('/\.css$/', $req_file)) {
        $css_list[] = _civicrm_resources_get_file_uri($index, $all_css); // same
        continue;
```

It works on Staging, because `node_modules` folder is not presented there, so only JS files from the *Regangular* are listed. However, once you install node modules, the script scans them as well and because some node modules contain CSS, the issue described above happens.

Not only that, the function that recursively searched for ALL JS/CSS files was called every time a resource needs to be injected, which affected the performance dramatically, especially in case of `node_modules` folder presence.

It was decided to simply specify resource paths relatively to their extension. This not only enhances the reliability that the correct resource file will be loaded, but also completely removes the need in recursive file listing, which improves the performance of pages load.

```php
function civicrm_resources_add_resources($extension_path, $resources_uris = []) {
  foreach ($resources_uris as $resources_uri) {
    _civicrm_resources_add_resource($extension_path, $resources_uri);

// ...

function _civicrm_resources_add_resource($extension_path, $resource_uri) {
  $resource_type = pathinfo($resource_uri)['extension'];
  $relative_to_drupal_extension_path = str_replace(DRUPAL_ROOT, '', $extension_path);
  $relative_to_drupal_resource_path = $relative_to_drupal_extension_path . '/' . $resource_uri;

   if ($resource_type === 'js') {
    drupal_add_js($relative_to_drupal_resource_path,  array('scope' => 'footer'));
  } else if ($resource_type === 'css') {
    drupal_add_css($relative_to_drupal_resource_path);
  } else {
    throw 'The resource type "' . $resource_type . '" is not supported';
```

### Usage

```php
civicrm_resources_load('org.civicrm.reqangular', [
  'js/src/common/vendor/require.min.js', // path relative to regangular extension path
  'js/dist/reqangular.min.js'
]);
```

-------------------------

✅Manual Tests - passed
⏹PHP Unit Tests - N/A


civicrm_resources/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+Usage
+-----
+
+The module invokes a custom drupal cache bin to store path information of extensions, so as not to call civicrm_initialize() on every api call.
+To load JS/CSS resources from "Job Contract (CiviHR)", call the API function as follows:
+
+```js
+civicrm_resources_load('org.civicrm.hrjobcontract', [
+  'css/dist/hrjc.css',
```
ajesamson: Can we use `*.css` or `*.js` here ❓

ajesamson: Tested locally and ok

igorpavlov: Yes, you can.

davialexandre: @ajesamson we'll not be supporting `*.css` and `*.js` anymore

civicrm_resources/civicrm_resources.module
```diff
@@ -6,205 +6,43 @@
...
- *     to 0.
- * @param $depth
- *   Current depth of recursion. This parameter is only used internally and
- *   should not be passed in.
- *
- * @return
- *   An associative array (keyed on the chosen key) of objects with 'uri',
- *   'filename', and 'name' members corresponding to the matching files.
+ * @param string $extension_path
+ *   Machine path of civicrm extension.
```
davialexandre: This is not really a path. In CiviCRM this is called `extension key`. Could you please update the param to use this terminology?

civicrm_resources/civicrm_resources.module
```diff
@@ -6,205 +6,43 @@
...
 }
 
 /**
- * API function to be used across drupal site to load resources.
+ * Loads resources from the given extension
  *
- * @param string $extention_full_name
- *   Machine name of civicrm extention.
+ * @param string $extension_full_name
+ *   Machine name of civicrm extension.
```
davialexandre: Please update the terminology here as well

civicrm_resources/civicrm_resources.module
```diff
@@ -213,3 +51,25 @@ function civicrm_resources_load($extention_full_name, $files = []) {
...
+/**
+ * Includes CSS/JS resources onto a page
+ *
+ * @param string $extension_path
+ *   Machine path of civicrm extension.
+ * @param array $resources_uris
+ *   Resource URI relative to extension paths to be included
+ */
+function _civicrm_resources_add_resource($extension_path, $resource_uri) {
+  $resource_type = pathinfo($resource_uri)['extension'];
```
davialexandre: Trying to access the extension like this will result in a notice if the $resource_uri doesn't contain an extension. To avoid that, you can ask pathinfo to only return the extension. This way, it will return a string instead of an array: `pathinfo($resource_uri, PATHINFO_EXTENSION)` instead

civicrm_resources/civicrm_resources.module
```diff
@@ -213,3 +51,25 @@ function civicrm_resources_load($extention_full_name, $files = []) {
...
+  $relative_to_drupal_extension_path = str_replace(DRUPAL_ROOT, '', $extension_path);
+  $relative_to_drupal_resource_path = $relative_to_drupal_extension_path . '/' . $resource_uri;
+
+  if ($resource_type === 'js') {
+    drupal_add_js($relative_to_drupal_resource_path,  array('scope' => 'footer'));
```
davialexandre: Please use short array syntax here

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -6692,14 +6700,17 @@ function _civihr_employee_portal_create_default_reports_configuration() {
  */
 function civihr_employee_portal_get_markup_for_extension($resource = '', $scriptExtensionKey, $markupModuleName) {
   $script = CRM_Core_Config::singleton()->debug ? "{$resource}.js" : "{$resource}.min.js";
+  $script_path_related_to_extension = 'js/dist/' . $script;
```
davialexandre: I'm not sure if this is correct. The line above this one indicates that this function should be able to also load non-distributive files, which I think are outside the `js/dist` folder, right? I think we shouldn't make any assumptions about the paths here

civicrm_resources/civicrm_resources.module
```diff
@@ -6,205 +6,43 @@
...
+ *   An array of resources URIs relative to extension paths to be included
  */
-function civicrm_resources_load($extention_full_name, $files = []) {
-  if (cache_get($extention_full_name, 'cache_civicrm_resources')) {
-    $ext_path_cache = cache_get($extention_full_name, 'cache_civicrm_resources');
+function civicrm_resources_load($extension_full_name, $resources_uris = []) {
+  if (cache_get($extension_full_name, 'cache_civicrm_resources')) {
+    $ext_path_cache = cache_get($extension_full_name, 'cache_civicrm_resources');
     $ext_path = $ext_path_cache->data;
   }
```
davialexandre: I was checking how civi resolves the path and it looks like it already does some caching internally, so I think that having this additional cache here it's actually hurting the performance. Could you please remove it?

civicrm_resources/civicrm_resources.module
```diff
@@ -213,3 +51,25 @@ function civicrm_resources_load($extention_full_name, $files = []) {
...
+ *
+ * @param string $extension_path
+ *   Machine path of civicrm extension.
+ * @param array $resources_uris
+ *   Resource URI relative to extension paths to be included
+ */
+function _civicrm_resources_add_resource($extension_path, $resource_uri) {
+  $resource_type = pathinfo($resource_uri)['extension'];
+  $relative_to_drupal_extension_path = str_replace(DRUPAL_ROOT, '', $extension_path);
+  $relative_to_drupal_resource_path = $relative_to_drupal_extension_path . '/' . $resource_uri;
```
davialexandre: To me, these two variable names sound a little bit awkward. How about `extension_path_relative_to_drupal` and `resource_path_relative_to_drupal`? Sound a bit more natural to my ears

civicrm_resources/civicrm_resources.module
igorpavlov: Are you sure @davialexandre ? Because it actually is an absolute path to the extension folder.

civicrm_resources/civicrm_resources.module
igorpavlov: done.

civicrm_resources/civicrm_resources.module
igorpavlov: Done.

civicrm_resources/civicrm_resources.module
igorpavlov: I checked:

|  | Current PR code | What is suggested |
| ------------- | ------------- | ------------- |
| After cache flush | 0.0013080 | 0.0001314 |
| Second load | 0.0006390 | 0.0001256 |

civicrm_resources/civicrm_resources.module
igorpavlov: Done.

civihr_employee_portal/civihr_employee_portal.module
davialexandre: I remember that, in the past, we used to include the src files instead of the dist one when debugging was enable. I know that Rene did some work to use sourcemaps instead, but I don't know if this works for all the extensions of not. If it's for all of them, then I don't think we the if condition checking if debugging is enabled or not.

civicrm_resources/civicrm_resources.module
```diff
@@ -6,210 +6,60 @@
...
-          if ($options['callback']) {
-            $options['callback']($uri);
-          }
-        }
-      }
-    }
-    closedir($handle);
-  }
-  return $files;
+function civicrm_resources_flush_caches() {
```
davialexandre: since we're not using the cache anymore, we can remove this

civihr_employee_portal/civihr_employee_portal.module
igorpavlov: I was thinking about this as well, I don't think we use src files anymore. Should we proceed and get rid of this condition?

davialexandre: Yes, let's do that

igorpavlov:
I simply squashed 3 PR Feedback commits into one for a cleaner history.

PR #544, opened 2018-08-30
PCHR-4152: Webform Logging Switch

mickadoo:
## Overview

We want to switch logging on when a user submits certain webforms, and off again after everything is saved. We will do this using webform hooks from the SSP. The actual logging will be handled in CiviHR in a separate PR.

## Before

There was no variable to track whether CiviHR logging was enabled or not.

## After

Logging for certain webform is enabled before webform data is saved and disabled after saving is complete.

## Technical Details

We use two hooks that `webform_civicrm` also uses. One is `webform_submission_presave` which is where CiviCRM saves most of the data. However it also saves some entity data in `webform_submission_insert` so the order of the hook firing should be as follows:

1. In SSP implementation of `webform_submission_presave` logging is enabled
1. In `webform_civicrm` implementation of `webform_submission_presave` some data is saved
1. In  `webform_civicrm` implementation of `webform_submission_insert` some more data is saved
1. In SSP implementation of `webform_submission_insert` logging is disabled

We control the order using the `module_implements_order` hook. This was already implemented in SSP but to reduce LOC in the module file and encapsulate the logic dedicated classes were created for this re-ordering of implementation.

The switch uses a global variable which will be accessed in CiviHR to do the actual logging. The reason we separate this is because we needed to use the Drupal hooks, but the logging is for CiviCRM changes which belongs more in a CiviCRM extension than a Drupal module.

The variable allows for separate streams, which allows you to start and stop logging for more than one event, for example during a single request with changes A, B, C and D we can log all changes from A -> C in one stream, but log changes B -> D in a separate stream.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/src/Hook/ModuleImplementsAlter/ImplementationAlterer.php
+++ b/civihr_employee_portal/src/Hook/ModuleImplementsAlter/ImplementationAlterer.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Drupal\civihr_employee_portal\Hook\ModuleImplementsAlter;
+
+interface ImplementationAlterer {
+
+  /**
+   * Check whether this class should act on the implementation array
+   *
+   * @param string $hookName
+   *
+   * @return bool
+   */
+  public function shouldAlter($hookName);
+
+  /**
+   * Make changes to the implementations
+   *
+   * @param array $implementations
+   *
+   * @return void
+   */
+  public function alter(array &$implementations);
+
+}
--- a/civihr_employee_portal/src/Hook/ModuleImplementsAlter/ImplementationAlterer.php
+++ b/civihr_employee_portal/src/Hook/ModuleImplementsAlter/ImplementationAlterer.php
@@ -2,7 +2,7 @@
 
 namespace Drupal\civihr_employee_portal\Hook\ModuleImplementsAlter;
 
-interface ImplementationAlterer {
+abstract class ImplementationAlterer {
 
   /**
    * Check whether this class should act on the implementation array
@@ -11,7 +11,7 @@ interface ImplementationAlterer {
    *
    * @return bool
    */
-  public function shouldAlter($hookName);
+  abstract public function shouldAlter($hookName);
 
   /**
    * Make changes to the implementations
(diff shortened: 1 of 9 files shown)


civihr_employee_portal/src/Hook/ModuleImplementsAlter/ImplementationAlterer.php
```diff
@@ -20,6 +20,22 @@ public function shouldAlter($hookName);
...
    * @return void
    */
-  public function alter(array &$implementations);
+  abstract public function alter(array &$implementations);
 
+  /**
+   * @param array $implementations
+   * @param string $moduleName
+   */
+  protected function moveToEnd(array &$implementations, $moduleName) {
```
tunbola: A method description would be good here.

civihr_employee_portal/src/Hook/ModuleImplementsAlter/ImplementationAlterer.php
```diff
@@ -20,6 +20,22 @@ public function shouldAlter($hookName);
...
+    if (!isset($implementations[$moduleName])) {
+      $serr = sprintf('Cannot re-order as module "%s" is not set', $moduleName);
```
tunbola: Can you rename this variable to something more readable. Maybe `error` or something better

civihr_employee_portal/tests/Helpers/LoggingStreamSwitcherTest.php
```diff
@@ -0,0 +1,28 @@
...
+
+  public function testDisablingWillSetGlobalVariableToFalse() {
+    LoggingStreamSwitcher::disableLogging('foo');
+    global $civihrChangeLogStreams;
+
+    $this->assertFalse($civihrChangeLogStreams['foo']);
+  }
+
+  public function testSwitchingWillChangeGlobalVariable() {
+    LoggingStreamSwitcher::enableLogging('foo');
```
tunbola: Since we are testing the switching, would be good to have an assert here: `$this->assertTrue(LoggingStreamSwitcher::isEnabled('foo'))` to check if it is true in addition to checking if it is false after the switch.

civihr_employee_portal/src/Hook/ModuleImplementsAlter/WebformSubmissionInsertReorderer.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Drupal\civihr_employee_portal\Hook\ModuleImplementsAlter;
+
+class WebformSubmissionInsertReorderer extends ImplementationAlterer {
```
tunbola: Looks like tests are missing for this class.

mickadoo:
Closing for now as based on comments from Jamie we will be going for a simpler approach without comparing data from previous submissions

PR #545, opened 2018-09-04
PCHR-4155: Notification Mails for SSP Edit and Onboarding Form

mickadoo:
## Overview

The purpose of this PR is to send notification mails to a specific user when someone updates their information either via the webforms on the SSP dashboard or via the onboarding form.

## Before

When a user updated their info via the SSP or onboarding form no notifications were sent.

## After

When a user updates their info via the SSP or onboarding form a notification is sent containing the submitted values to a specific email.

#### Edit Personal Details

image

#### Edit Contact Details

image

#### Edit My Address

image

#### Edit Payroll Information

image

#### Add Emergency Contact

image

#### Add Dependant

image

#### Onboarding Form (no skipped steps)

image

#### Onboarding Form (with skipped steps)

image

## Technical Details

This PR builds on #528 which added the notification settings.

The flow for the notifications is:

- Inside `hook_webform_submission_insert` call `WebformSubmissionnotifier::sendNotification`
- `sendNotification` checks notification settings to decide whether the mail should be sent
- `WebformSubmissionnotifier` fetches the target email and calls `drupal_mail`
- `civihr_employee_portal_mail` creates a mail class to represent the mail to be sent
- The data from the submission is fetched and formatted using the mail class
- `MailContentReplacer` generates the mail body using smarty and the params from the mail class
- The `dashboard_edit_details.tpl` file is used by smarty to create the mail body
- The mail is sent

## Comments

I refactored `civihr_employee_portal_mail_alter` to use the `MailContentReplacer`. In this function the pre-built mail is already ready to be sent, and we want to replace the body with new content from Smarty. However the logic for generating the mail content is the same and so I used this class to prevent duplication.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/src/Mail/OnboardingFormEditNotificationMail.php
+++ b/civihr_employee_portal/src/Mail/OnboardingFormEditNotificationMail.php
@@ -1,0 +1,157 @@
+<?php
+
+namespace Drupal\civihr_employee_portal\Mail;
+
+use Drupal\civihr_employee_portal\Helpers\WebformHelper;
+
+class OnboardingFormEditNotificationMail extends WebformSubmissionNotificationMail {
+
+  /**
+   * @param array $message
+   *
+   * @return array
+   */
+  public function getVariables($message) {
+    $variables =  parent::getVariables($message);
+    // Replace the title
+    $variables['webformTitle'] = 'the onboarding form';
+
+    return $variables;
+  }
+
+  /**
+   * @inheritdoc
+   */
+  protected function formatSubmittedValues($node, $submission) {
+    $submittedValues = WebformHelper::getSubmittedValues($node, $submission);
+
+    $this->formatPersonalDetails($submittedValues);
+    $this->formatPayroll($submittedValues);
+    $this->formatEmergencyContacts($submittedValues);
+    $this->formatDependants($submittedValues);
+
+    // We don't want this info in the email
+    unset($submittedValues['Profile Picture']);
+
+    return $submittedValues;
+  }
+
+  /**
+   * @param array $submittedValues
+   */
+  private function formatPersonalDetails(&$submittedValues) {
+    $personalDetails = &$submittedValues['Personal Details'];
+    unset($personalDetails[0]['Existing Contact']);
+
+    // The options for this are not stored correctly and need manual replacement
+    $ssnApplicationKey = 'I am currently applying for a NI/ SSN';
+    $value = $personalDetails[0][$ssnApplicationKey];
+    $replacement = $value ? 'Yes' : 'No';
+    $personalDetails[0][$ssnApplicationKey] = $replacement;
+  }
+
+  /**
+   * @param $submittedValues
+   */
+  protected function formatPayroll(&$submittedValues) {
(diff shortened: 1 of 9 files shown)


civihr_employee_portal/src/Mail/OnboardingFormEditNotificationMail.php
```diff
@@ -0,0 +1,157 @@
...
+    $payroll = &$submittedValues['Payroll'];
+    $skipStepOption = 'Add Now (Payroll)';
+    if ($this->skippedStep($payroll, $skipStepOption)) {
+      $this->formatSkippedStep($payroll);
+    }
+    unset($payroll[0][$skipStepOption]);
+  }
+
+  /**
+   * @param $submittedValues
```
ajesamson: param type is missing here as well.

civihr_employee_portal/src/Mail/OnboardingFormEditNotificationMail.php
```diff
@@ -0,0 +1,157 @@
...
+    if ($this->skippedStep($dependants, $skipStepOption)) {
+      $this->formatSkippedStep($dependants);
+      return;
+    }
+    // Unset unused data
+    unset($dependants[0]);
+    // Remove unused dependant entries
+    $dependantsKeys = ['Fifth', 'Fourth', 'Third', 'Second', 'First'];
+    foreach ($dependantsKeys as $index => $dependantKey) {
+      if (!isset($dependantKey[$index + 1])) {
```
ajesamson: The logic here is a bit confusing to me. From the foreach, `$dependantKey` is a string but is being accessed as an array. Or do you mean to write `$dependants`

civihr_employee_portal/src/Mail/OnboardingFormEditNotificationMail.php
```diff
@@ -0,0 +1,157 @@
...
+    $dependants = &$submittedValues['Dependants'];
+    $skipStepOption = 'Add Now (Dependants)';
+    if ($this->skippedStep($dependants, $skipStepOption)) {
+      $this->formatSkippedStep($dependants);
+      return;
+    }
+    // Unset unused data
+    unset($dependants[0]);
+    // Remove unused dependant entries
+    $dependantsKeys = ['Fifth', 'Fourth', 'Third', 'Second', 'First'];
```
ajesamson: Do you mind renaming this to `$dependantKeys` since you are using `$dependantKey` in the foreach. 🤔 
No errors just common convention.

civihr_employee_portal/src/Mail/OnboardingFormEditNotificationMail.php
mickadoo: Good catch! It should be `$dependantKeys`.

 I know the logic is a bit complicated here, I tried it in two different ways. Basically I want to go through the list of dependants and unset all those that have not been explicitly added to the form by the user. The user must select "Add another" from the bottom of each dependent input, so I though going through them in reverse order and checking the previous one would work, if you have another suggestion let me know :-)  I'll add a comment to explain it further anyway

PR #546, opened 2018-09-05
PCHR-4175: Emergency Contact Deletion Notification

mickadoo:
## Overview

The purpose of this PR is to add an email notification when a user deletes an emergency contact from the SSP hr-details page.

## Before

No notification was sent

## After

When you delete an emergency contact (or dependant) from the SSP hr-details page an email is sent to a specific user.

image

## Technical Details

Some refactoring was done to move the useful CiviCRM Contact related method to a dedicated service. The following methods were used so they could be shared:

- `getContactWorkEmail`
- `getLinkToContactProfile`

We use a new endpoint `emergency_contacts/{id}/delete` to handle the deletion request. We only call this from the SSP, which is implemented in https://github.com/compucorp/civihr-employee-portal-theme/pull/385. A new service was created to handle the emergency contact deletion and sending of notification. The reason for this is because we want to hook into the emergency contact deletion from the SSP, but previously when a direct CiviCRM API call was used this was difficult. Using the new endpoint we can handle sending of the notification as well as deletion of the emergency contact.

---

- [x] Tests Pass


civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -1723,6 +1731,17 @@ function civihr_employee_portal_menu() {
   return $items;
 }
 
+/**
+ * Callback function for 'emergency_contacts/{id}/delete
+ *
+ * @param string $id
```
davialexandre: Shouldn't this be an `int` instead?

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -17,9 +17,11 @@ use Drupal\civihr_employee_portal\Helpers\LinkProvider;
 use Drupal\civihr_employee_portal\Helpers\TaxonomyHelper;
 use Drupal\civihr_employee_portal\Mail\MailContentReplacer;
 use Drupal\civihr_employee_portal\Mail\WebformSubmissionNotificationMail;
-use Drupal\civihr_employee_portal\Mail\WebformSubmissionNotifier;
+use Drupal\civihr_employee_portal\Service\WebformSubmissionNotifier;
 use Drupal\civihr_employee_portal\Mail\OnboardingFormEditNotificationMail;
 use Drupal\civihr_employee_portal\View\CustomFieldsMappingFixer;
+use Drupal\civihr_employee_portal\Service\EmergencyContactDeletionService;
+use Drupal\civihr_employee_portal\Mail\EmergencyContactDeletionNotificationMail;
```
davialexandre: Please keep these in alphabetical order

civihr_employee_portal/src/Service/EmergencyContactDeletionService.php
```diff
@@ -0,0 +1,68 @@
...
+      $err = sprintf('Emergency contact with ID "%d" was not found', $id);
+      throw new \Exception($err);
+    }
+
+    if (!$this->canDelete($emergencyContact)) {
+      $err = 'You do not have permission to delete that emergency contact';
+      throw new \Exception($err);
+    }
+
+    $service->delete($id);
```
davialexandre: Is there any reason for using the `EmergencyContactService` directly instead of calling the `Contact.deleteEmergencyContact` API? I think that using the API would be better here for two reasons:
- We avoid exposing implementation details of the `hremergency` extension to the Drupal module (I know we already do this in other places, but we should avoid it when possible)
- This class is duplicating all the logic we already have [here](https://github.com/compucorp/civihr/blob/staging/org.civicrm.hremergency/api/v3/Contact/Deleteemergencycontact.php#L25-L43)

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -1601,6 +1603,12 @@ function civihr_employee_portal_menu() {
     'type' => MENU_CALLBACK,
   ];
 
+  $items['emergency_contacts/%/delete'] = [
+    'page callback' => 'civihr_employee_portal_emergency_contact_delete',
+    'page arguments' => [1],
+    'access callback' => TRUE,
```
davialexandre: What does this mean in terms of permissions? Who will have access to this page?

civihr_employee_portal/src/Mail/EmergencyContactDeletionNotificationMail.php
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Gets the emergency contact from the message and determines if it is a
+   * 'dependant' or 'emergency contact'
+   *
+   * @param array $message
+   *
+   * @return string
+   */
+  private function getEmergencyContactType($message) {
```
davialexandre: Looks like you only really need the emergency contact here. How about passing only that instead of the entire message?

civihr_employee_portal/src/Service/EmergencyContactDeletionService.php
```diff
@@ -0,0 +1,68 @@
...
+    $this->sendNotification($emergencyContact);
+  }
+
+  /**
+   * Sends the notification mail about emergency contact deletion
+   *
+   * @param array $emergencyContact
+   */
+  protected function sendNotification($emergencyContact) {
+    if (!WebformSubmissionSettingsService::shouldSendMail()) {
```
davialexandre: If I got this right, this is not really a webform submission and yet we're using the `WebformSubmissionSettingsService` here. Maybe we need a better name to this class?

civihr_employee_portal/civihr_employee_portal.module
mickadoo: It should, but it's actually a string since it seems all these params are passed as a string. I'm pretty sure there are plenty of places like this where I _want_ it to be an int, like if I'm fetching some value using a BAO, but in fact it's a string so I think it's better to say what it is instead of what it should be so nobody does a strict comparison which would fail

civihr_employee_portal/src/Service/EmergencyContactDeletionService.php
mickadoo: Good point, I forgot it was me that added that endpoint! I'd still like to use the `EmergencyContactService` to fetch the emergency contact details though, because it helps a lot in getting back a entity with logical property names (using the custom field name) instead of `name_12` etc.

civihr_employee_portal/civihr_employee_portal.module
mickadoo: Everyone, I thought about adding a permission callback, but it would just be duplicating the permission check in the service. I can do that if you want, but it would mean we check "is this the emergency contact owner" both before accessing the page and during page execution

civihr_employee_portal/src/Service/EmergencyContactDeletionService.php
mickadoo: Yep, it started just as webform submissions, but it does apply to emergency contact deletion too so I'll rename it

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -1723,6 +1731,32 @@ function civihr_employee_portal_menu() {
...
+ * @param string $id
+ *   The emergency contact ID
+ */
+function civihr_employee_portal_emergency_contact_delete($id) {
+  $service = new EmergencyContactDeletionNotifier();
+  $result = ['status' => 200];
+
+  try {
+    civicrm_api3('Contact', 'deleteemergencycontact', ['id' => $id]);
+    $service->notifyDeletion($id);
```
davialexandre: The `notifiyDeletion()` method loads the emergency contact with the given ID, but at this point, it was already deleted. Won't this be a problem?

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -1723,6 +1731,32 @@ function civihr_employee_portal_menu() {
...
+  } catch (\Exception $exception) {
+    $result = [
+      'status' => 500,
+      'error' => $exception->getMessage(),
+    ];
+  }
+
+  drupal_add_http_header('Content-Type', 'application/json');
+  http_response_code($result['status']);
```
davialexandre: Why was it necessary to change the response to be like this now?

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -1723,6 +1731,32 @@ function civihr_employee_portal_menu() {
...
+  echo json_encode($result);
+  exit();
```
davialexandre: I think we don't need to do anything about the response here, but if that is really necessary, the right way to do that in drupal is by setting the **delivery callback** in hook_menu to `drupal_json_output`: https://drupal.stackexchange.com/questions/149157/how-to-return-jsonp-from-a-menu-callback

civihr_employee_portal/civihr_employee_portal.module
davialexandre: As you said, there are plenty of places like this. Unfortunately, is something caused by how badly the tools and libraries we use deal with data.

Given this is something out of our control, I think it's better for us to use the docblock to inform what we expect this to be, rather than what the library/framework/whatever will pass. I think it's better having a failing strict comparison, than someone thinking they could pass something like `foobar` to the function.

civihr_employee_portal/civihr_employee_portal.module
mickadoo: I was using this, but it did some strange things like encoding quotes differently so I thought I'd do it myself. If you think we don't need to have any special response (see comment above) then I'll remove all output.

civihr_employee_portal/civihr_employee_portal.module
mickadoo: I just thought it would be nicer to have a properly functioning endpoint that returns a correct status code and JSON (since this is just used in AJAX requests) but if you think it's not necessary I can remove any output

civihr_employee_portal/civihr_employee_portal.module
mickadoo: Yep, my fault - I should have checked this but it was silently failing and showing an empty contact name in the email. I've fixed this now by fetching the contact before deletion and then passing it to the notifier

civihr_employee_portal/civihr_employee_portal.module
davialexandre: please remove it

civihr_employee_portal/civihr_employee_portal.module
```diff
@@ -1,25 +1,28 @@
...
 use Drupal\civihr_employee_portal\Mail\OnboardingFormEditNotificationMail;
+use Drupal\civihr_employee_portal\Mail\PasswordResetSystemMail;
+use Drupal\civihr_employee_portal\Mail\WebformSubmissionNotificationMail;
+use Drupal\civihr_employee_portal\Page\HRDetailsPage;
+use Drupal\civihr_employee_portal\Security\PublicFirewall;
+use Drupal\civihr_employee_portal\Service\EmergencyContactDeletionNotifier;
+use Drupal\civihr_employee_portal\Service\WebformSubmissionNotifier;
 use Drupal\civihr_employee_portal\View\CustomFieldsMappingFixer;
+use Drupal\civihr_employee_portal\Webform\WebformTransferService;
+use CRM_Hremergency_Service_EmergencyContactService as EmergencyContactService;
```
davialexandre: Please keep this one in alphabetical order as well.

PR #561, opened 2018-10-08
PCHR-4281: Image Display Fixes

mickadoo:
## Overview

Photos uploaded from the onboarding form are not displayed in either the contact summary page or the SSP. This PR fixes both those issues.

## Before

- Image URLs with `photo=0` in it would not display on the contact profile page
- Image URLs uploaded from the onboarding form would not display on the `/hr-details` personal details section

#### Profile Page

image

#### HR Details Page

image

## After

- Contact images are displayed regardless of where they're uploaded from
- Existing images with `photo=0` in them are fixed to remove this
- Upgrade functions in files in the `updates` directory are automatically included

#### Profile Page

image

#### HR Details Page

image

#### On Contact Profile Page

When fetching the image URL, `\CRM_Utils_File::getImageURL` the mime type incorrectly be `image/jpg?photo=0` for a jpg image with `photo=0` in the query string. This has a knock-on effect meaning the image will be treated as a file and not displayed.

#### HR Details Page

The SSP view field for the image URL uses an `image_type` of thumbnail. This was added [here](https://github.com/civicrm/civicrm-drupal/blob/6c718b104faf64a72900d5c99ec66cb9e612faf0/modules/views/civicrm/civicrm_handler_field_contact_image.inc#L127). This image type does not support images uploaded via the webform and always expects images to be in the format `photo=blahblah`. To fix this the type was removed meaning it will be handled as a normal image.

## Comments

I used the new upgrader format and auto-included all upgrader files in the `updates` directory.

---

- [x] Tests Pass

### DIFF
--- a/civihr_employee_portal/tests/Helpers/UrlHelperTest.php
+++ b/civihr_employee_portal/tests/Helpers/UrlHelperTest.php
@@ -63,6 +63,11 @@ class UrlHelperTest extends \PHPUnit_Framework_TestCase {
         'photo',
         'http://civihr.local/sites/default/files/webform/500.jpg',
       ],
+      [
+        'http://example.com?a=1#foo',
+        'a',
+        'http://example.com#foo',
+      ],
     ];
   }
 
--- a/civihr_employee_portal/tests/Helpers/UrlHelperTest.php
+++ b/civihr_employee_portal/tests/Helpers/UrlHelperTest.php
@@ -15,6 +15,58 @@ class UrlHelperTest extends \PHPUnit_Framework_TestCase {
   }
 
   /**
+   * @dataProvider urlWithQueryProvider
+   *
+   * @param string $url
+   * @param string $key
+   * @param string $expected
+   */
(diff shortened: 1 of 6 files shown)


civihr_employee_portal/tests/Helpers/UrlHelperTest.php
```diff
@@ -14,6 +14,63 @@ public function testDedupeWillReturnExpectedResults($url, $expected) {
...
+        'b',
+        'http://example.com?a=1',
+      ],
+      [
+        'http://civihr.local/sites/default/files/webform/500.jpg?photo=0',
+        'photo',
+        'http://civihr.local/sites/default/files/webform/500.jpg',
+      ],
+      [
+        'http://example.com?a=1#foo',
```
davialexandre: I think this is not even a valid URL, as, according to my understanding, the fragment must be part of the path. I guess it's fine to have a test to show how we expect the helper to work in a situation like that anyway

PR #566, opened 2018-12-12
PCHR-4378: Remove Vacancies from features slideshow

igorpavlov:
## Overview

This PR removes Vacancies from features slideshow for both new and existing sites (via an upgrader).

### Removing the slide for new sites 

We simply delete the whole Vacancies node from the `onboarding_slideshow_images.export` file.

### Removing the slide for existing sites

We need an upgrader here:

```php
function civihr_employee_portal_update_7043() {
  $vacanciesNodeTitle = 'Vacancies';
  # Fetching a Vacancies node ID by a title
  $vacanciesNodeId = db_query("SELECT nid FROM node WHERE title = '{$vacanciesNodeTitle}'")
    ->fetchField();
  
  # Then, if node exists, deleting it
  $vacanciesNodeId && node_delete($vacanciesNodeId);
```

## Comments

⚠️In case Vacancies node does **not** exist, the upgrader will not do anything. If the node was deleted manually - this is not a problem. However, if it was renamed from "Vacancies" to something else, it seems to be impossible to identify it and, thus, it would not be removed.

ℹ️`drush_civihr_employee_portal_refresh_node_export_files()` did not delete the node from the database. So we needed to come up with an alternative that you see in the upgrader.

---

✅Manual Tests - passed
⏹PHP Unit Tests - N/A


civihr_employee_portal/updates/7043.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * Remove Vacancies feature from the slideshow after new contact onboarding
+ */
+function civihr_employee_portal_update_7043() {
+  $vacanciesNodeTitle = 'Vacancies';
+  $vacanciesNodeId = db_query("SELECT nid FROM node WHERE title = '{$vacanciesNodeTitle}'")
+    ->fetchField();
+
+  $vacanciesNodeId && node_delete($vacanciesNodeId);
```
tunbola: While the shorthand looks good here, It's not easy to catch what is happening at first glance, for more clarity I would suggest we go the conditional route.

```php
if ($vacanciesNodeId) {
node_delete($vacanciesNodeId)
}

PR #571, opened 2019-01-04
PCHR-4424: Fixed File Get Contents Warning on Onboarding Form

ajesamson:
## Overview
When filling the onboarding forms, a warning is usually displayed on the `Home Address` page with 
`Warning: file_get_contents(...)`. This affects three files - require.min.js, reqangular.min.js, and tasks-notification-badge.min.js. The warning can also be noticed on `My Details` page and `Add New Staff Member` page. This warning was fixed in this PR

## Before
File get contents warning came up while rendering some pages as part of the resource files could not be loaded.
onboarding-form

## After
File get contents warning no longer show on affected pages.
onboarding-form-after

## Technical Details
When loading the affected resource files, a leading slash (`/`) was present in the file path. The `/` makes `file_get_contents` start at the top of filesystem, thereby raising a warning. A check was added to [civicrm_resources](https://github.com/compucorp/civihr-employee-portal/blob/master/civicrm_resources/civicrm_resources.module#L45) module  to test for leading slash and update resource path if present.

---

- [x] Manual Tests Pass

### DIFF
--- a/civicrm_resources/civicrm_resources.module
+++ b/civicrm_resources/civicrm_resources.module
@@ -46,6 +46,9 @@ function _civicrm_resources_add_resource($extension_path, $resource_uri) {
   $resource_type = pathinfo($resource_uri, PATHINFO_EXTENSION);
   $extension_path_relative_to_drupal = str_replace(DRUPAL_ROOT, '', $extension_path);
   $resource_path_relative_to_drupal = $extension_path_relative_to_drupal . '/' . $resource_uri;
+  if (substr($resource_path_relative_to_drupal, 0, 1) === '/') {
+    $resource_path_relative_to_drupal = substr($resource_path_relative_to_drupal, 1);
+  }
 
   if ($resource_type === 'js') {
     drupal_add_js($resource_path_relative_to_drupal,  ['scope' => 'footer']);


tunbola:
A few questions. Since the function we made changes to i.e `_civicrm_resources_add_resource` is a general function used in other places to load css and Js resources, won't this change affect those places. This scenario seems to be particular to onboarding form otherwise, we would notice errors on other pages notifying of failure in loading asserts. Is it worth checking if the why it only occurs for these pages only and not all pages? Or am I missing something?

ajesamson:
@tunbola 
> won't this change affect those places

It will not affect them. They are all using file_get_contents which expects a URL or resource file path. The issue with the affected files was that; while composing their full path, a leading slash was added to the provided [extension path](https://github.com/compucorp/civihr-employee-portal/blob/master/civicrm_resources/civicrm_resources.module#L47). 
Full path: `/vagrant/test1/sites/all/modules/civicrm/tools/extensions/civihr/org.civicrm.reqangular`
Drupal Root:
`/vagrant/test1`

>Is it worth checking if the why it only occurs for these pages only and not all pages

 The problem was also reported on [Add New Staff Member](https://compucorp.atlassian.net/browse/PCHR-4424?focusedCommentId=78512&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-78512) and My Details page. But the implementation catered for all as they all refer to same source.

